PR #2, opened 2018-03-12
Fix K3 send_morse and change generic to 0x20 padding per manual. Test…

mdblack98:
…ed on TS-2000

### DIFF
--- a/kenwood/kenwood.c
+++ b/kenwood/kenwood.c
@@ -2784,7 +2784,7 @@ int kenwood_send_morse(RIG *rig, vfo_t vfo, const char *msg)
         * If answer is something else, return with error to prevent infinite loops
         */
        if (!strncmp(m2,"KY0", 3)) break;
-       if (!strncmp(m2,"KY1", 3)) usleep(500000); else return -RIG_EINVAL;
+       if (!strncmp(m2,"KY1", 3)) usleep(50000); else return -RIG_EINVAL;
     }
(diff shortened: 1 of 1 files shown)


N0NB:
Hi Mike.

It looks like you included a formatting change with this patch.  Perhaps it would be better just to issue a pull request with the actual changes and then we'll do the formatting another time as its own patch.  What do you think?

mdblack98:
OK...didn't remember formatting but no problem...I'll re-do.

I get the feeling not many are using the send_morse as this has been broken for a long time apparently.


Mike 

    On Monday, March 12, 2018, 2:07:24 PM CDT, Nate Bargmann <notifications@github.com> wrote:  
 
 
Hi Mike.

It looks like you included a formatting change with this patch. Perhaps it would be better just to issue a pull request with the actual changes and then we'll do the formatting another time as its own patch. What do you think?

N0NB:
* On 2018 12 Mar 14:09 -0500, mdblack98 wrote:
> OK...didn't remember formatting but no problem...I'll re-do.

Well, Git is replacing the entire file for some reason which is hiding
your actual fix portion of the patch.

> I get the feeling not many are using the send_morse as this has been
> broken for a long time apparently.

Agreed.  Thanks for your attention.

73, Nate

-- 

Ham radio, Linux, bikes, and more: http://www.n0nb.us

PR #3, opened 2018-03-12
Fix K3 send_morse and change generic to 0x20 padding per manual. Test…

mdblack98:
…ed on TS-2000

### DIFF
--- a/kenwood/kenwood.c
+++ b/kenwood/kenwood.c
@@ -2786,18 +2786,28 @@ int kenwood_send_morse(RIG *rig, vfo_t vfo, const char *msg)
        if (!strncmp(m2,"KY0", 3)) break;
        if (!strncmp(m2,"KY1", 3)) usleep(500000); else return -RIG_EINVAL;
     }
-    /*
-     * Make the total message segments 28 characters
-     * in length because Kenwood demands it.
-     * Spaces fill in the message end.
-     */
     buff_len = msg_len > 24 ? 24 : msg_len;
 
     strncpy(m2, p, 24);
     m2[24] = '\0';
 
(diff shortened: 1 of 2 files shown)


N0NB:
I did have to patch the kenwood.c file for MinGW as it didn't like 'i' being declared as part of the for loop initializer on line 2807.  Oh well, it still works the same.  :-)

mdblack98:
I take it that didn't work on the earlier mingw...I'm using 5.3.0 now from Qt11 so that's why it worked here.  Working on fixing a guy's bug with WSJT-X where moving his window from a 4K monitor with 250% scaling crashes when moved to his other monitor.  Qt11 fixed the window title scaling problem he originally had and introduced the crash to have to debug Qt11 now.


----------------------------------- 
Michael D. Black 

    On Friday, March 16, 2018, 2:12:43 PM CDT, Nate Bargmann <notifications@github.com> wrote:  
 
 
I did have to patch the kenwood.c file for MinGW as it didn't like 'i' being declared as part of the for loop initializer on line 2807. Oh well, it still works the same. :-)

N0NB:
* On 2018 16 Mar 14:31 -0500, mdblack98 wrote:
> I take it that didn't work on the earlier mingw...

The Jessie VM I use has MinGW 3.2, so, yeah, it's rather old.

> I'm using 5.3.0 now from Qt11 so that's why it worked here. 

After the 3.2 release I may switch to using either a newer Debian
version or the latest Slackware in a VM for the builds.

I suspect that it really doesn't matter for the Windows stuff and may
even be an improvement.

73, Nate

-- 

Web: http://www.n0nb.us  GPG key: 0xD55A8819  GitHub: N0NB

PR #4, opened 2018-03-15
IC-7300: Enable get_split_vfo and remove rptr_shift/rptr_offs, which are not supported. Allow setting RIT/XIT to zero. Fix typo in bandplan (used only by IC-7300 so far).

mikaelnousiainen:
Fixing more IC-7300 issues:
* IC-7300 supports getting split VFO, but does not supported repeater shift/offset.
* Previously it was impossible to set RIT/XIT to zero. Also refactored BCD encoding/decoding.
* Fix typos in bandplan for 60m and 6m.


N0NB:
Thanks, Mikael.

PR #5, opened 2018-03-15
TM-D710: Rewrite backend

mikaelnousiainen:
Rewrite the whole Kenwood TM-D710 backend, since most of it was not working anyway.
TM-D710 protocol is significantly different from TM-D700 or the TH handheld ones.

Most features work, but the control protocol is quite limited: 
* It is not possible to change the selected frequency band of a VFO
* Selecting MEM VFO uses the previously selected band (VFOA/VFOB), since there are no separate MEM VFO definitions for A/B bands

Added FMN mode to string constants, so that it can be used with the set_mode command.

### DIFF
--- a/kenwood/tmd710.c
+++ b/kenwood/tmd710.c
@@ -48,7 +48,6 @@
 #include "tones.h"
 #include "num_stdio.h"
 #include "misc.h"
-#include "../include/hamlib/rig.h"
 
 static int tmd710_get_freq(RIG *rig, vfo_t vfo, freq_t *freq);
 static int tmd710_set_freq(RIG *rig, vfo_t vfo, freq_t freq);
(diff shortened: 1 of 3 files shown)


kenwood/tmd710.c
```diff
@@ -35,6 +48,7 @@
 #include "tones.h"
 #include "num_stdio.h"
 #include "misc.h"
+#include "../include/hamlib/rig.h"
```
N0NB: Please replace this with `#include <hamlib/rig.h>` as this is an installed header and should be included in this manner per GNU documentation.

mikaelnousiainen: Sure. That was inserted by the IDE and I didn't catch it :)

mikaelnousiainen: Actually, the file was already included, so just removing the line.

N0NB: I see that now and even there it is included incorrectly.  Sigh.  I'll get that corrected "one of these days".  One thing to be aware of is that relative include paths like that will probably break VPATH builds and would be caught by running `make distcheck` which is the main way that I check for regressions.

N0NB:
Thanks, Mikael.

PR #6, opened 2018-03-19
Update AOR 3030 to actually work now.  Tested with WSJT-X

mdblack98:


N0NB:

73, Nate

-- 

PR #8, opened 2018-03-27
Allow rigctld to reopen after error

mdblack98:
To test
Run rigctld to connect to socket client like FLRig
Run rigctl -m 2 to connect to rigctld
Kill FLRig
Execute command from rigctl to read Freq...quits on error as expected.
Start rigctl again and it can't connect
This patch fixes this behavior


N0NB:
Thanks, Mike.

PR #9, opened 2018-04-12
Improve network connect error message

mdblack98:


N0NB:
So time is added with -vvvvvv?  If so, we should also include mention in the documentation.

N0NB:
OTOH, I think it would be useful to add time to any of the debug levels in some way from the command utilities.

mdblack98:
Just got his out of my spam folder.Yeah, right now the 6th v is needed.But do you think a command line switch would be better to allow time on any debug level?
Mike

 

    On Thursday, April 12, 2018, 6:45:21 AM CDT, Nate Bargmann <notifications@github.com> wrote:  
 
 
So time is added with -vvvvvv? If so, we should also include mention in the documentation.

mdblack98:
Went and used -Z for this since -T was already taken in rigctld.
de Mike W9MDB



 

    On Friday, April 13, 2018, 6:20:51 PM CDT, Nate Bargmann <notifications@github.com> wrote:  
 
 * On 2018 13 Apr 15:51 -0500, mdblack98 wrote:
> 
> Just got his out of my spam folder.Yeah, right now the 6th v is
> needed.But do you think a command line switch would be better to allow
> time on any debug level?

I think that would be quite useful. As '-t' is already in use, I'm
thinking that '-T', '--time-stamp', would be a good choice. It should
only be used if '-v'+ is present, otherwise ignored.

Thoughts?

73, Nate

-- 

possible worlds. The pessimist fears this is true."

Web: http://www.n0nb.us GPG key: D55A8819 GitHub: N0NB

PR #10, opened 2018-04-14
Improve network message and add time stampes to debug messages

mdblack98:


N0NB:
I did a bit more fixup to the files and rebased the whole thing!  Hopefully, that won't be a problem for you.

PR #13, opened 2018-04-16
Adding correct get/set functions for LNA- Mixer- IF gain

dl1ksv:
The Funcube Pro+ differs from the Funcube in it's functions.

- It has a LNA gain setting which only can be enabled  or disabled
- It has a Mixer gain setting which only can be enabled  or disabled
- It has a IF gain setting which ranges form 0 .. 59
- It has no get RSS call
So different functions have to be implemented  for the Funcube and the Funcube Pro+
This patch implements the correct functions for the Pro+
Get /Set frequency can treated the same way


N0NB:
Applied and pushed.  Thanks, Volker.

PR #16, opened 2018-04-22
Set FILPATHLEN to 512 instead to support longer device file names

mikaelnousiainen:
Set FILPATHLEN to 512 instead of 100 to support longer device file names.

In some cases, static serial port device file names (for USB devices) generated by udev may be longer than 100 chars.


N0NB:
Oh, yay!  More goodness brought to the world from the idiots at Red Hat!

N0NB:
Applied and pushed.  Thanks, Mikael.

PR #17, opened 2018-04-25
Fix TenTec OmniVII get_mode for VFOB

mdblack98:


N0NB:

PR #18, opened 2018-04-25
Fix -Z option on rigctld

mdblack98:


N0NB:

PR #23, opened 2018-05-09
Allow rigctld to continue operating after tcp or serial rig disappears

mdblack98:


N0NB:
* On 2018 08 May 23:16 -0500, mdblack98 wrote:
> -- Commit Summary --
> 
>   * Allow rigctld to continue operating after tcp or serial rig
>   disappears

Would it be a good idea to apply the same fix to rotctld?  I like to see
each deamon have the same basic behavior if possible.

73, Nate

-- 

PR #24, opened 2018-05-09
Fix read_string terminator length parameters

mdblack98:
Also promoting rigctld to STABLE.  Heavily tested with WSJT-X split mode ops.

### DIFF
--- a/dummy/netrigctl.c
+++ b/dummy/netrigctl.c
@@ -52,9 +52,11 @@ static int netrigctl_transaction(RIG *rig, char *cmd, int len, char *buf)
 {
   int ret;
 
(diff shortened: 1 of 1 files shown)


N0NB:
I'm wondering if the "sizeof" operators should be replaced with a hard coded value?  Judging by what I find I think this change should be safe.

N0NB:
* On 2018 09 May 08:17 -0500, mdblack98 wrote:
> Also promoting rigctld to STABLE.  Heavily tested with WSJT-X split mode ops.
> You can view, comment on, or merge this pull request online at:

73, Nate

-- 

PR #26, opened 2018-05-09
new kenwood th-d72 driver

bagel99:
The current master version of the TH-D72 driver just doesn't work.  This one is almost a complete rewrite and is mostly tested.

### DIFF
--- a/kenwood/thd72.c
+++ b/kenwood/thd72.c
@@ -1,8 +1,7 @@
 /*
- *  Hamlib Kenwood backend - TH-D72 description
- *                            cloned after TH-D7
+ *  Hamlib Kenwood TH-D72 backend
  *  Copyright (c) 2000-2011 by Stephane Fillod
- *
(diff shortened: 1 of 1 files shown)


N0NB:
I am seeing the following warning from MinGW when compiling for Windows:

  CC       thd72.lo
thd72.c: In function ‘thd72_set_freq’:
thd72.c:268:5: warning: format ‘%ld’ expects argument of type ‘long int’, but argument 3 has type ‘long long int’ [-Wformat=]
     sprintf(fbuf, "%010ld", (int64_t)freq);
     ^

Whenever you get back to that file it might be worth checking that.

N0NB:
* On 2018 09 May 16:25 -0500, bagel99 wrote:
> The current master version of the TH-D72 driver just doesn't work.
> This one is almost a complete rewrite and is mostly tested.

Welcome to Hamlib!

73, Nate

-- 

PR #27, opened 2018-05-10
Fix a format 

bagel99:
Fix a format so that it (hopefully) works with Windows.  I can't test because I have no Windows machines.


N0NB:
* On 2018 10 May 12:38 -0500, bagel99 wrote:
> Fix a format so that it (hopefully) works with Windows.  I can't test
> because I have no Windows machines.

I don't think it made a difference at run time, it was just a compiler
warning that is now nicely quelled.  Thanks for the fix!

73, Nate

-- 

PR #37, opened 2018-06-27
Fixes for the piHPSDR backend.

dl1ycf:
Two issues:
a) fix setting RFPOWER level: the value in the PC command must be max. 100

b) Allow hijacking two RTTY modes for PKTUSB and PKTLSB, piHPSDR uses these for
    DIGU and DIGL ("digital" versions of USB and LSB).


N0NB:
Applied and pushed.  Thanks!

73, Nate

-- 

PR #39, opened 2018-07-06
Force back to split after set_vfo A if needed\nFixes split getting tu…

mdblack98:
…rned off


N0NB:
Applied, rebased, and pushed.  Thanks, Mike.

73, Nate

-- 

PR #40, opened 2018-07-06
Added Meade LX200 Protocol to Rotors

andz82:
* Added new rotor module (18) for Meade telescope rotors
* Implemented all needed rotor functions
* Added README.md

* Tested with DS-2000 Rotor with Autostar 494 and 506 serial cable
* Tested on Linux (Debian) with rotctl and gpredict


N0NB:
Applied, rebased, and pushed.  Thanks!

73, Nate

-- 

PR #44, opened 2018-08-10
Fix RFPOWER level for FT-450D

bagel99:
Yaesu chose to change the way the PC command works. On the FT-450 the range is 000-255.
On the FT-450D the range is 005-100 (the actual power in watts).


N0NB:

PR #47, opened 2018-09-08
Add support for CM108AH chipsets

phl0:
CM108 uses PID of 0x13c which needs to be added. Tests with rigctl showed it is working fine.


N0NB:
Applied and pushed.  Thanks!

PR #49, opened 2018-09-13
Update power reading values for FT900

mdblack98:


N0NB:
Applied and pushed, also rebased onto another patch.

Thanks, Mike.

PR #50, opened 2018-09-16
Fix install instructions on README.

2E0PGS:
I had to run the bootstrap script for the configure script to exist then the rest of the process could be followed.

### DIFF
--- a/README
+++ b/README
@@ -78,6 +78,7 @@ Recompiling
 Hamlib is entirely developed using GNU tools, under various Linux systems.
 The library may be recompiled by the familiar "three step":
 
+        ./bootstrap
         ./configure
         make
         sudo make install


N0NB:
Actually, this README is generic and included with the source tarballs where the bootstrap has already been done.  The 'bootstrap' step is appropriate for the README.developer file which is also referenced from README.betatester only when working from a Git cloned local working tree.

2E0PGS:
My apologies. Perhaps somewhere it should point to README.developer on the main README. Otherwise a WIKI page for "Building from source" which mentions README.developer? Just to make it clear as day.

My apologies again that it wasn't clear enough for me to know the info for Git related environment was in that file.

N0NB:
* On 2018 16 Sep 12:36 -0500, Peter Stevenson wrote:
> My apologies. Perhaps somewhere it should point to README.developer on
> the main README. Otherwise a WIKI page for "Building from source"
> which mentions README.developer? Just to make it clear as day.
> 
> My apologies again that it wasn't clear enough for me to know the info
> for Git related environment was in that file.

Don't sweat it, Peter.  :-)

I guess the assumption has been that those doing a Git clone would see
the other two files and take a peek at them.  The README is tailored for
the source tarballs so mentioning things only applicable to Git clones
may create more confusion.

73, Nate

-- 

PR #51, opened 2018-09-18
kenwood/thd72.c bug fix

bagel99:
Fix range argument in several calls to thd72_get_menu_item.


N0NB:
Sorry for the delay.  This came in when my attention was directed elsewhere and I forgot about it.

It has now been applied and pushed.

Thanks!

PR #55, opened 2018-10-03
Fix multi-mode-mapping in FLRig

mdblack98:


N0NB:

PR #56, opened 2018-10-28
ft817: adding delay before next retry after returning from tx to rx

phaethon:
My previous commit added checking of actual status change after coming from tx to rx and retrying in case the status had not changed. Apparently, it is much more useful if there is a delay between retries as my rig (MCHF) takes some time (<.1s) to actually change the state. For original FT817 this should work the same as before as delay is only after first attempt.


N0NB:
Applied and pushed.  Thanks, Eriks.

PR #60, opened 2018-11-07
WIP: Add lo_freq setting to allow use with transverters

C-Elegans:
This PR allows software using rigctld with radio that has a transverter connected to display the true tx frequency of the radio instead of the IF frequency. 

To use, invoke rigctld with -C lo_freq=xxxx, where xxxx is the lo_frequency of the transverter in Hz.


N0NB:
That is an outstanding idea!  Thanks!

PR #61, opened 2018-11-11
Change barrett P command to T command to avoid EPROM burnout

mdblack98:


N0NB:

PR #62, opened 2018-11-20
Change variable length argument handling in affected functions

airween:
dummy/netrigctl.c uses sprintf() without check of argument length. This patch replaces the fixed length character buffers by dynamically allocated.


N0NB:
Applied and pushed.  Thanks, Ervin.

PR #68, opened 2018-12-19
Complete IC-R6 functions

malcolmh:
All CI-V commands listed in the current manual have been implemented, with the exception of DTCS squelch setting due to RIG_FUNC being full.


N0NB:
Applied and pushed.  Thanks, Malcom.

73, Nate

-- 

PR #71, opened 2018-12-21
Add IC-R30

malcolmh:
Second attempt!


N0NB:
Looks good!

73, Nate

-- 

PR #74, opened 2018-12-24
Icr6 feature complete - move to Beta

malcolmh:
Utilized the new 64-bit function space to create two new functions needed for this device.


N0NB:

73, Nate

-- 

PR #77, opened 2018-12-31
Automate Icom CI-V duplex setting

malcolmh:
A new icom_rig_open function is available to test which duplex state the CI-V port is in & uses that state for all further transactions.


g4wjs:
Hi Malcolm,

this confuses me. I see that detecting full duplex changes the controller id in commands sent from 0xe0 to 0x80 but I don't see any mention of that in modern Icom CI-V docs. Is this some undocumented feature?

I ask because if a user turns off USB Echo Back on a rig that allows it, say the IC-7300, this patch will detect it as full duplex and send commands with the 0x80 controller id, does this actually work?

73
Bil
G4WJS.

N0NB:
On 31 Dec 2018, at 11:30, Bill Somerville <notifications@github.com> wrote:
> 
> this confuses me. I see that detecting full duplex changes the controller id in commands sent from 0xe0 to 0x80 but I don't see any mention of that in modern Icom CI-V docs. Is this some undocumented feature?
> 
> 

My mods do not change the controller the ID. Are you referring to the line: "ctrl_id = priv->serial_full_duplex == 0 ? CTRLID : 0x80;” in frame.c? If so, all I changed was the serial_full_duplex source from priv_caps to priv_data. As to the intent of this statement, you will have to ask the original author.

g4wjs:
Hi Malcolm,

icom_rig_open does more than change the source of serial_full_duplex, it sets it automatically. My concern is that if, for example, an IC-7300 owner turns off USB Echo Back this change (assuming icom_rig_open were enabled for that rig, and I assume that is the intention from the comments about USB connected rigs) then serial_full_duplex will be enabled and the controller address sent in messages will change to 0x80. My question is, will that work?

As far as I can see this serial_full_duplex option was added for the IC-92(D|A|E) hand-held by Stéphane in commit 8bf9bbb back in 2011, I can't find any CAT documentation for that rig but I guess the 0x80 controller address is specific to that model. Do you have information that it is a general CI-V feature?

73
Bill
G4WJS.

malcolmh:
On 31 Dec 2018, at 12:02, Bill Somerville <notifications@github.com> wrote:
> 
> icom_rig_open does more than change the source of serial_full_duplex, it sets it automatically. My concern is that if, for example, an IC-7300 owner turns off USB Echo Back this change (assuming icom_rig_open were enabled for that rig, and I assume that is the intention from the comments about USB connected rigs) then serial_full_duplex will be enabled and the controller address sent in messages will change to 0x80. My question is, will that work?

If the full duplex setting is manually changed on the rig after hamlib start-up, then CI-V transactions will fail. That was true before my mods & is still true.

> 
> As far as I can see this serial_full_duplex option was added for the IC-92(D|A|E) hand-held by Stéphane in commit 8bf9bbb back in 2011, I can't find any CAT documentation for that rig but I guess the 0x80 controller address is specific to that model. Do you have information that it is a general CI-V feature?
> 
> 

The current IC-92AD manual has no CI-V info. I see from ic92d.c that the address is set to 0x01.

malcolmh:
PS: Looking at the debug output on rigctl, when the ID is changed to 0x80, icom_transaction receives and processes the incoming frames OK.

g4wjs:
Hi Malcolm,

ok, that is somewhat reassuring. So to be clear, on a USB connected Icom with USB Echo Back disabled and REMOTE CI-V un-linked, the rig still responds to *all* commands even though they originate from CI-V address 0x80 rather than the usual 0xe0.

I am still concerned, do you know that this is true for *all* USB connected Icoms that can disable CI-V Echo Back? Would it not be better to have a new status flag that is set by the test in icom_rig_open by discovery, then in icom/frame.c only do the address change from the existing capabilities flag, that the IC-E92 sets, but do the echo handling depending on either flag being set? This seems safer rather than hijacking a different but similar facility for a new purpose.

Am I missing something here? I should point out that I don't have access to either an IC-E92 or a modern Icom with USB connection to test with.

73
Bill
G4WJS.

malcolmh:
The default value for the rig_open member of rig_caps is NULL, which nearly all current Icom rigs are set to. In these cases, icom_rig_open is not called and the value of serial_full_duplex in the icom_priv_caps struct is used and not altered. Only if rig_open=icom_rig_open does the test and possible override of serial_full_duplex occur. I did it this way to enure full backward compatibility.

g4wjs:
Hi Nate,

not by me. It is expecting all current and future Icom rigs with a USB connection to work with an undocumented controller CI-V address of 0x80. I don't see why that should be necessary. I would also like to know what CI-V address the rig responds to when it has been commanded by something on address 0x80 i.e. does it reply to 0x80 or to the conventional documented controller address of 0xe0?

Note the response that the new icom_rig_open is not used yet doesn't solve the issue for me as I believe the intention is that all Icoms with the USB Echo Back SET MENU option should use the new icom_rig_open.

Until someone from Icom, or some published documentation is uncovered that, explains what CI-V address 0x80 is really for, I think the conservative option of adding a new flag for discovery of USB Echo Back being disabled should be used rather than hijacking some feature that appears to be specific to the IC-E92 range of handhelds.

73
Bill
G4WJS.

malcolmh:
The 0x80 thing is not part of my changes.  That has been in the code since 2011. Is Stéphane still in circulation? He was the author, so maybe he can shed some light on the subject.

g4wjs:
Hi Malcolm,

well it is part of your change. You have written code to set a flag that changes the CI-V controller address. The fact that Stéphane contributed the code to change the controller address is not relevant, what is relevant is that I don't think it is necessary or wise for this usage. It is trivial to add a new flag with a more appropriate name, as this is nothing to do with half/full-duplex, something like priv->USB_echo_back_disabled and set that flag with your code in icom_rig_open rather than messing with the priv_caps->serial_full_duplex flag. Then all that is needed is to check that new flag in icom/frame.c @ line 130:

	if (!priv_caps->serial_full_duplex || !priv->USB_echo_back_off) {

73
Bill
G4WJS.

g4wjs:
Hi Malcolm,

that should of course be icom/frame.c @ line 130:

if (!priv_caps->serial_full_duplex && !priv->USB_echo_back_off) {

73
Bill
G4WJS.

malcolmh:
Bill,
I don't understand your objection. The serial_full_duplex flag is there now, but as a  member of a const struct. All I did was make a mutable copy of the flag so it can be flipped at start-up to reflect the actual duplex state of the CI-V link.

g4wjs:
Malcolm,

my objection is simple, you have proposed a change that relies on an undocumented behaviour of some rigs responding to commands from a CI-V address of 0x80. That is unnecessary, I have demonstrated how to fix it. Why should Hamlib rely on an undocumented behaviour that Icom can change at any time even if it actually works on every rig where icom_rig_open might be useful at the moment.

I don't understand on your insistence for using an existing flag to do something different from its original intent while it still is in use for that original intent, that is just hacking and even if it works for the rig you have tested it with that doesn't make it right.

Am I misunderstanding that your change is about automatically detecting if USB Echo Back has been disabled and USB Unlink from REMOTE has been enabled. That is not anything to do with duplex, all CI-V interfaces are full-duplex, the options are about whether an underlying CI-V CSMA/CD bus is emulated over a USB serial link.

Anyway, you clearly do see the new use for the flag as different from its original intent even though it has an undesirable side-effect, so I will shut up now.

Bill.

malcolmh:
Bill, your contention is that I am "using an existing flag to do something different from its original intent ". What do you understand that original intent to be? Here is its declaration in icom.h:

int serial_full_duplex; /*!< Whether RXD&TXD are not tied together */

On the 0x80 issue, I tried an experiment - I removed that statement that changes the ID. It works without it!
If I include that change in my PR, will that satisfy you?

g4wjs:
Malcolm,

Sure you can do that if you have an IC-E92 handheld to test with and whatever interface cable Stéphane used (some sort of programming cable I assume) when he found it necessary to add that option. The serial full duplex capabilities flag was added for that rig, it was only ever switched on for that rig.

That option you have now hijacked for another purpose was added to support some feature of the IC-E92 handheld, if you change the controller address to 0xe0 then I can only assume you will break that.

I just don't get it, we are having a huge conversation thread about something so trivial. Why is my suggestion to add a new private flag, instead of hijacking one already used for something else, not the best solution here?

Bill.

malcolmh:
Let's start again!

g4wjs:
Hi Malcolm,

that's much better. I like the new capabilities flag as well so that icom_rig_open can be used with any Icom backend. It should be rolled out for all the rigs with the USB Echo Back SET MENU option at least.

Minor nit-pick, line https://github.com/Hamlib/Hamlib/pull/77/commits/7d517b267c6454de7233d37a74fd65870e83cbf7#diff-939bdd6a231ff1e3c8ed55e0f980028bR471 may be superfluous or maybe moved to a new function icom_rig_close(). No big deal either way or as you have it as it is not clear whether there should be a close to match any open function in Hamlib.

73
Bill
G4WJS.

PR #82, opened 2019-01-02
Automatically detect USB echo state on start-up

malcolmh:
New icom_rig_open function that if the flag serial_USB_echo_check is set in icom_priv_caps, will test the echo state of the CI-V link & set a flag for frame.h to use accordingly.


N0NB:
I guess this has settled out a bit.  I would like to see this cleaned up to just the relevant changes without extra commits if possible.

malcolmh:
OK, I'll rewind to just the relevant changes & re-submit.

PR #83, opened 2019-01-09
Automatically detect CI-V echo state

malcolmh:
Some Icom rigs use USB for the CI-V connection. This can be set to echo or not. This mod permits a check of the echo state on start-up & flags the frame handling accordingly.

A new member of the icom_priv_caps is defined to flag that this check should be done. A new icom_rig_open function performs this check & sets a flag in priv_data which frame.c then uses to process the input accordingly.


N0NB:
Applied and pushe.  Thanks, Malcolm.

73, Nate

-- 

PR #93, opened 2019-02-22
Added support for microHAM MicroKeyer-III

dl1ycf:
The title says it all. There is a new microHam device MicroKeyer-III which uses the same protocol
but has a new identification string. The change is to add this string to the list of "known devices",
and increase NUMUHTYPES by 1.


N0NB:
Applied and pushed.  Thanks!

73, Nate

-- 

PR #102, opened 2019-05-13
Change TS790 handhsake to none and promote to stable.  Works with WSJT-X

mdblack98:


g4wjs:
Hi Mike,

why the change of handshaking default? This rig requires an IF-232C or IF10D interface, both require hardware handshaking. If you are basing this change to a different non-Kenwood interface, I wonder if it is wise.

73
Bill
G4WJS.

mdblack98:
Hmmm...let me double check my tester....I thought he had just a USB/serial cable.  But he many have missed connecting the correct pins.It's definitely working without flow control on both FLRig and hamlib.
The IF232C and IF10D don't do any protocol changes...it's just a 5V/12V conversion so should be the same as a current USB cable.
I don't see any mention of hardware flow in the manual (doesn't mention baud rate either though).
Mike 

    On Monday, May 13, 2019, 7:30:16 AM CDT, Bill Somerville <notifications@github.com> wrote:  
 
 
Hi Mike,

why the change of handshaking default? This rig requires an IF-232C or IF10D interface, both require hardware handshaking. If you are basing this change to a different non-Kenwood interface, I wonder if it is wise.

73
Bill
G4WJS.

g4wjs:
Hi Mike,

some users loop RTS/CTS at the rig, I suspect some third-party interfaces do the same. That would make it look like hardware handshaking was not required.

73
Bill
G4WJS.

mdblack98:
Yeah...but that should still work with hardware handshaking if looped correctly.  I don't think this guy did that but I'll check.
The TS-790 cat manual does show the CTS/RTS pins so something is amiss here.  CTS would have to be set high if the rig was requiring hardware flow control and the computer was not providing it.
Nate...please close this pull request for now


 

    On Monday, May 13, 2019, 7:54:36 AM CDT, Bill Somerville <notifications@github.com> wrote:  
 
 
Hi Mike,

some users loop RTS/CTS at the rig, I suspect some third-party interfaces do the same. That would make it look like hardware handshaking was not required.

73
Bill
G4WJS.

N0NB:
* On 2019 13 May 08:04 -0500, Michael Black wrote:
> Nate...please close this pull request for now

Okay!

73, Nate

-- 

N0NB:
Closing per request of Mike.

PR #110, opened 2019-05-31
Fix: Get Frequency. Status Beta

mrmoonlv:
Fixed: 
- Get Frequency 

Changes
- Status to Beta
- Credits

73
DE
YL3ALK


N0NB:
Applied and pushed.  Thanks, Kārlis.

73, Nate

-- 

PR #112, opened 2019-06-05
Add new capabilities, improve Icom and Elecraft backends

mikaelnousiainen:
This is a rather large change to
a) fix many issues in Icom rig backends
b) add new capabilities to enable features present in many (Icom and also Elecraft) rigs
c) clean up code a bit

General changes:
* Extend `dump_caps` output for `rigctld`: add complete details of ext levels and frequency ranges
* Add new level: `RFPOWER_METER`, `COMP_METER`, `VD_METER`, `ID_METER`, `NOTCHF_RAW` (= `NOTCHF` without Hz scale, it could have a better name), `MONITOR_GAIN`, `NB` (noise blanker level)
* Add new functions: `DUAL_WATCH` and `DIVERSITY` (receive)

Icom fixes:
* Add support for `RFPOWER_METER`, `COMP_METER`, `VD_METER`, `ID_METER` meters as many recent Icom models support them. There is a rig-specific calibration table (similar to the S-meter one= for each model.
* Use `NOTCHF_RAW` instead of `NOTCHF` to make manual notch frequency work properly (according to API spec). Icom doesn't provide any information on the scale of notch frequency.
* Fix `VOXDELAY` level for many Icom rigs. The command is rig-specific.
* Add `MONITOR_GAIN` level for many Icom rigs.
* Add `NB` level for Icom rigs that support it
* Add `DUAL_WATCH` support for higher-end Icom rigs that support it
* Add rig-specific AGC levels, they vary from model to model
* Make set/get commands for RIT/XIT shared functions and enable them on recent Icom models
* Add basic set of filters (with 3 choices per mode, so the filter choices work properly) for many Icom models
* Fix `TUNER` func and `VFO_OP_TUNE` and enable them for recent Icom models that support them
* Fix `RIG_MODE_PKTAM` and add missing PKT modes for many Icom models
* Add custom levels for drive gain and DIGI-SEL for higher-end Icom rigs that support them
* Remove `RIG_FUNC_FAGC` from models that support proper AGC levels
* Move rig-specific code from `icom_set_parm()`/`icom_get_parm()` to backends of the specific models
* Remove `civ_version` definition, as there is no such concept in CI-V command sets: all rig-specific commands were moved to their respective backends, so the shared Icom code is a bit cleaner now.
* Introduce `icom_set_raw()`/`icom_get_raw()` helper functions for executing rig-specific commands.
* Introduce `icom_set_custom_parm()`/`icom_get_custom_parm()` helper functions for executing rig-specific "parameters" (`0x1A 0x05` set of commands that vary from model to model)
* Introduce `icom_set_custom_parm_time()`/`icom_get_custom_parm_time()` helper functions for setting/getting time with a rig-specific command
* Add `set_powerstat`/`get_powerstat` for many rigs that support it, but were missing the definition
* Add set/get split for many rigs that support it, but were missing the definition

Elecraft fixes:
* Add support for many K3 commands (most of these are supported also by K3S, KX2 and KX3, as the command sets are identical for the _most_ part), including the new `DUAL_WATCH` and `DIVERSITY` functions
* Add some custom ext levels for Elecraft-specific settings
* Add rig models for K3S, KX2 and KX3 -- although they are all very similar, they do have some differences. There may be some differences that I have not noticed (I could only test on K3 at the moment), but now it's easier to implement rig-specific fixes.

Kenwood fixes:
* Add RIT/XIT functions for TS-480 and TS-890

All new features have been tested on IC-7000, IC-756ProIII, IC-7300, IC7600 and the Elecraft ones on K3. Obviously there is a possibility of a typo or a similar bug on rig models that I haven't directly tested, so additional testing on Elecraft KX2/KX3/K3S or on some of the newer Icom rigs (e.g. IC-7100, IC-7610, IC-7700, IC-7800, IC-7851) is welcome.


kenwood/k3.c
```diff
@@ -228,6 +308,449 @@ const struct rig_caps k3_caps = {
...
+    .set_powerstat =	kenwood_set_powerstat,
+    .get_powerstat =	kenwood_get_powerstat,
+    .set_ant =		kenwood_set_ant_no_ack,
+    .get_ant =		kenwood_get_ant,
+    .send_morse =		kenwood_send_morse,
+
+};
+
+const struct rig_caps kx3_caps = {
+    .rig_model =		RIG_MODEL_K3,
```
N0NB: Did you mean RIG_MODEL_KX3 here?

kenwood/k3.c
```diff
@@ -228,6 +308,449 @@ const struct rig_caps k3_caps = {
...
+    .set_powerstat =	kenwood_set_powerstat,
+    .get_powerstat =	kenwood_get_powerstat,
+    .set_ant =		kenwood_set_ant_no_ack,
+    .get_ant =		kenwood_get_ant,
+    .send_morse =		kenwood_send_morse,
+
+};
+
+const struct rig_caps kx2_caps = {
+    .rig_model =		RIG_MODEL_K3,
```
N0NB: Did you mean RIG_MODEL_KX2 here?

.editorconfig
```diff
@@ -0,0 +1,8 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 4
```
N0NB: This looks like an editor specific file that I'd prefer we don't add to the repository.

kenwood/k3.c
mikaelnousiainen: :+1: Indeed!

kenwood/k3.c
mikaelnousiainen: Yep :)

.editorconfig
mikaelnousiainen: This is actually not an editor-specific file, but one that is designed so that it can be read and supported by almost any editor. Many major IDEs and some of the more lightweight editors already support this directly or have a plugin for it. See: https://editorconfig.org/ -- anyway, your call.

N0NB: Ahh, very good!  I was obviously unaware of that.  Sometime some comments added to it to explain what it is for and the Web URL will be a good idea for others that are curious.  I've no problem keeping it in the patch set as it appears useful.

N0NB:
Applied and pushed.  Thanks, Mikael.

73, Nate

-- 

PR #113, opened 2019-06-09
Backup serial options on open, restore on close

stromnet:
On Mac OS the Prolific PL2032 kernel driver locks after hamlib (and plenty of other tools) closes the serial port. The next call to open() will hang forever, and a hard reboot is required to recover. And the shutdown even fails, since the kernel driver is totally hung, and a hard poweroff is required.

Restoring the termios settings before closing seems to remove the issue. I originally observed that `cu` was able to run repeatedly, this was just a hunch.

The implementation uses a single-linked list based on the `fd` to find the original options, not sure if there might be other things to take into consideration as this is my first look at the hamlib source. Feel free to point me in the right direction if needed :)

Note that I've only tested this on MacOS, but I don't see how it would cause any problems to restore the port to the state it was before started using it.

---

This was originally observed with WSJT-X, clicking the Test CAT button never gave desired results and then WSJT-X failed to exit without force-killing it (and even then the process was stuck).

In my case the PL2303 device in question is the one built into an "Icom Data Mode Cable" (UDO2IC), which normally only sets RTS, but I've modified it to use the RX/TX lines to communicate with the CI-V port on my IC-703, allowing me to use a single USB tty for both PTY and CI-V.

### DIFF
--- a/src/serial.c
+++ b/src/serial.c
@@ -88,6 +88,19 @@
 static int uh_ptt_fd   = -1;
 static int uh_radio_fd = -1;
 
+typedef struct term_options_backup {
+    int fd;
+#if defined(HAVE_TERMIOS_H)
+    struct termios options;
+#elif defined(HAVE_TERMIO_H)
+    struct termio options;
+#elif defined(HAVE_SGTTY_H)
+    struct sgttyb sg;
+#endif
+    struct term_options_backup *next;
+} term_options_backup_t;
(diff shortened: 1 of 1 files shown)


N0NB:
Hi, Johan, welcome to Hamlib.

Thanks for this patch.  Hopefully, it will get much testing.

73, Nate

-- 

stromnet:
Thanks, that was quick! :)

73 de SA6JHN

PR #116, opened 2019-07-10
Fix FT-891 split operation

jothan:
This fixes split operation for the FT-891.

The current state of affairs was using commands that did not exist on this model.

I have successfully tested WSJT-X in split mode with this PR.

WSJT-X would previously timeout when an invalid command was sent with a "timeout while getting VFO frequency error".

73,
Jonathan VA2XZA


N0NB:
Applied and pushed.  Thanks, Jonathan.

73, Nate

-- 

Web: https://www.n0nb.us

PR #117, opened 2019-07-10
easycomm: Fix futile check of speed range

acinonyx:
The conditional expression in which the speed parameter is validated
is such that it can never become true. This looks like an error
inadvertently introduced.

Signed-off-by: Vasilis Tsiligiannis <acinonyx@openwrt.gr>


N0NB:
Applied and pushed.  Thanks and welcome to Hamlib, Vasilis.

73, Nate

-- 

Web: https://www.n0nb.us

PR #120, opened 2019-07-16
easycomm: Return the correct value of pointer val

zisi:
Add value of ackbuf into value of val instead of
the address, because isn't the same after the call
of function easycomm_rot_get_conf().

Signed-off-by: Agis Zisimatos <agzisim@gmail.com>


N0NB:
Applied and pushed.  Thanks, Agis.

73, Nate

-- 

Web: https://www.n0nb.us

PR #121, opened 2019-07-24
Remove old comment in k3.c

mdblack98:


N0NB:
I might have known when I wrote that, it would come back to haunt me...

N0NB:
Finally remembered to merge this...

PR #127, opened 2019-09-10
Fix issues with EasyComm II in rotctl

mfncooper:
This pull request addresses the issues brought up in e-mail regarding (a) commands that have no response timing out when waiting for one, and (b) an EasyComm I command sequence being sent to EasyComm II controllers. The fixes for (a) and (b) are in separate commits for convenience in tracking.


N0NB:
Pulled, merged, and pushed.  Thanks for contributing to Hamlib, Martin.

73, Nate

-- 

Web: https://www.n0nb.us

PR #135, opened 2019-10-11
Enable verbose mode in bindings/pytest.py

gvanem:
Support 'pytest.py -v to show more details.

I've not managed to build a Python 3.x module, so I've not patched py3test.py.
But imagine it would be exactly the same.

### DIFF
--- a/bindings/pytest.py
+++ b/bindings/pytest.py
@@ -10,13 +10,13 @@ import sys
 
 import Hamlib
 
-def StartUp():
+def StartUp(verbose):
     """Simple script to test the Hamlib.py module with Python2."""
 
     print "%s: Python %s; %s\n" \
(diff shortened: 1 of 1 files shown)


N0NB:
Applied and pushed.  Thanks for your contribution, Gisle.

As for compiling against Python 3, see
https://github.com/N0NB/Hamlib/blob/master/bindings/README.python
which covers the steps.  In the future Python 3 will be the default, it
appears the next stable release of Debian will enforce this, so the
extra step will need to be taken for a Python 2 module.  Over time that
need will likely decrease.

73, Nate

-- 

Web: https://www.n0nb.us

gvanem:
I also tried `luatest.lua`, but that failed on some cryptic issue. A PR for another time.

PR #152, opened 2019-12-05
Fix & re-map Icom receiver DCS function calls

malcolmh:
As receivers cannot have functions that are transmitter-only, it was illogical to have a get/set_dcs_code function & no get/set_dcs_sql function. This change duplicates the former to the latter and re-maps all receivers that have DCS functionality from the "code" functions to the "sql" functions.


N0NB:

73, Nate

-- 

Web: https://www.n0nb.us

PR #174, opened 2020-01-13
CPPCheck changes

mdblack98:


dl1ycf:
To be consistent, uh_open_wkey() should also be added to src/microham.c.

dl1ycf:
And likewise, both uh_open_wkey() and uh_close_wkey() should be "activated" in
src/microham.c, if this is not yet the case.

PR #236, opened 2020-05-04
Update sample Windows build scripts

N0NB:
Mostly replace backtick calls with more portable $() syntax.

Revise each script to minimize differences between them to only those
that are necessary.

### DIFF
--- a/scripts/build-w32.sh
+++ b/scripts/build-w32.sh
@@ -1,14 +1,14 @@
 #!/bin/bash
 
-# Builds Hamlib 3.x W32 binary distribution.
+# Builds Hamlib 4.x W32 binary distribution.
 
-# A script to build a set of W32 binary DLLs from a Hamlib tarball.
-# This script assumes that the Hamlib tarball has been extracted to the
-# directory specified in $BUILD_DIR and that libusb-1.x.y has also
-# been extracted to $BUILD_DIR.  The MS VC++ Toolkit must also be installed
-# and working with Wine.
-#
-# See README.build-win32 for complete details.
+# A script to build a set of W32 binary DLLs from a Hamlib tarball.  This
+# script assumes that the Hamlib tarball has been extracted to the directory
+# specified in $BUILD_DIR and that libusb-1.x.y has also been extracted to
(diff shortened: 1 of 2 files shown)


N0NB:
I'd like to drop the use of Wine and trying to build .lib files for the various MS tools.  Wine is simply not up to the task as it is not fully compatible with the various .NET components required.  Trying to keep up with the changes is a maintenance headache.

Better, IMO, is to simply provide the hamlib-4.def file along with some basic instructions on how Windows developers can create a local hamlib-4.lib file with their installed tooling.  Doing so relieves us of the burden of trying to make this work on a foreign system and provides them with the correct file for their development installation

I propose we simply drop the need for Wine and the calls to it in the scripts.

mdblack98:
Itried the tools installation too and it doesn't work under Wine anymore.
But...was able to use my Ubuntu VM to reference the underlying Win10 installation and link that way.
Have to test if we can just copy the link.exe over and use that rather than having to modify the script for every version of tools available.  Then we just tell the user to find the link.exe and copy it to some fixed place we decide on.

N0NB:
I think we can safely operate under the assumption that developers know their tools.  By providing the libhamlib-4.def file we're giving them what is needed to generate the libhamlib-4.lib file that will match their toolset installation, so long as we document it.

I will start a discussion on the list.

mdblack98:
I'll see about coming up with a solution to make it easier...or at least more obvious....

Mike
 

    On Monday, May 4, 2020, 11:27:56 AM CDT, Nate Bargmann <notifications@github.com> wrote:  
 
 


I think we can safely operate under the assumption that developers know their tools. By providing the libhamlib-4.def file we're giving them what is needed to generate the libhamlib-4.lib file that will match their toolset installation, so long as we document it.

I will start a discussion on the list.

N0NB:
* On 2020 04 May 11:32 -0500, Michael Black wrote:
> I'll see about coming up with a solution to make it easier...or at least more obvious....

No, let's just drop that part of the generation.  It's a maintenance
headache and there is no reason that Windows developers cannot do this
as needed or via a script.

73, Nate

-- 

Web: https://www.n0nb.us

PR #296, opened 2020-06-11
Increased timeout, necessary because of changes in pihpsdr software

dl1ycf:
Since recently, the processing of all CAT commands within pihpsdr is put into the "GTK idle queue". Therefore it can take a *long* time until a response is sent. This is in particular the case for "set frequency" if it changes the band, or for RX/TX transitions. Although these commands produce now answer, the answer to the subsequent "ID;" command is delayed, possibly by few hundred milli-seconds. The same applies if the CAT command arrives while the GUI is buy with mouse click events.


dl1ycf:
Withdrawn because of an open PR on the same file.

PR #374, opened 2020-09-10
update version string for icm803

bkubesh:
Update version string to show rig version instead of backend version.


mdblack98:
Not the correct thing to do.  The ICM803 uses the icmarine routines exclusively so is totally dependent on the icmarine version.
icmarine is more likely to change.  That's why it uses BACKEND ".0" where the .0 can be updated if needed.
If you remove BACKEND then any icmarine changes won't show up in the ICM803 version at all which would be incorrect.
This is what we do with all rigs.  Only rigs that are standalone get their own version.

Issue #423, opened 2020-10-19
FTDX101 control issues


mikaelnousiainen:
I've done some thorough testing on an FTDX101 (only accessible to me remotely via Hamlib) and found the following issues with the latest master on Oct 19th 2020.

Rig model in Hamlib is:
```
  1040  Yaesu                  FT-DX101D               20201016.5      Stable      RIG_MODEL_FTDX101D
```

- [x] FM mode is missing (PKTFM is there)
- [x] FM / AM filters cannot be used (only 2 filters?) -- changing filter width has no effect
- [x] Check if FTDX101 supports freely adjustable filter width (filter width 0 in Hamlib)
- [x] MIC GAIN range is incorrect (only 0-40 when it should be 0-100%)
- [x] COMP cannot be turned off (but no error)
- [x] MONITOR cannot be turned on (but no error)
- [x] LOCK cannot be turned off
- [x] Meters added: S-meter, comp meter, RF power meter, current, voltage
- [x] Fixed '+\get_ctcss_tone' fails with response: Communication timed out (error code -5)
- [x] Fixed incorrect scale for RF POWER
- [x] SPLIT verified to work correctly
- [x] RIT/XIT verified to work correctly, removed automatic enabling of RIT/XIT when adjusting offset frequency

I'll be fixing (hopefully most of) these in the coming days and will be updating this issue, but I thought I'd leave this open if I don't get all of them fixed.

mikaelnousiainen:
At least some of these (the filters?) may apply to other Yaesu rigs, like FTDX5000. I'll most likely be able to test FTDX5000 after this rig.

mikaelnousiainen:
@mdblack98 Just a note that I'm currently working on fixing a bunch of issues in Yaesu newcat control commands, will push out changes next week once I have properly tested them.

mikaelnousiainen:
@mdblack98 Would you happen to have any ideas on what the scales for Yaesu meters (S-meter, comp, power output, vdd, idd) are? It's really weird that many rigs have the capability to read these meters, but there seems to be no documentation on what the values actually mean in the official Yaesu CAT manuals...

mdblack98:
Lots of rigs need custom tables -- you can see the FTDX101D swr table for example in newcat.cSWR is usually non-linear so a simple scale doesn't work.
Other values that are linear should work with a simpler scaling factor.
[...]

PR #426, opened 2020-10-26
A large set of fixes for Yaesu FTDX rigs

mikaelnousiainen:
I've done some thorough testing on an FTDX101, found and fixed the following issues with the latest master on Oct 19th 2020.

- Add FM mode (PKTFM is there)
- Fix FM / AM filters -- there is only a single filter in FTDX101
- Add freely adjustable filter width (filter width 0 in Hamlib) for FTDX101
- MIC GAIN range is incorrect (only 0-40 when it should be 0-100%)
- COMP cannot be turned off (but no error)
- MONITOR cannot be turned on (but no error)
- LOCK cannot be turned off
- Fix CTCSS tone get/set for FTDX rigs with new command format
- Fixed incorrect scale for RF POWER
- SPLIT verified to work correctly
- RIT/XIT verified to work correctly, removed automatic enabling of RIT/XIT when adjusting offset frequency
- Meters added: S-meter, comp meter, RF power meter, current, voltage
  - Added sane defaults for S-meter and SWR meter, which seem to work fine for FTDX101. The rest of the meters are defined with dummy scales (0.0-1.0) waiting for more accurate information.
- Changed FTDX rig names to match how Yaesu calls the rigs (e.g. FTDX 3000).
- Changed FT-1200 rig name to FTDX 1200.
- Add support for setting roofing filter explicitly via an ext level in all FTDX rigs
- Allow `?` response from rig to be interpreted as "rejected by rig" error, because many commands return this response if the parameter is not valid (e.g. when browsing to a memory channel that doesn't exist).
- Add MEM VFO for all FTDX rigs. Verified that memory channel browsing works correctly for FTDX101.

Most of these issues have been fixed for all of the following Yaesu rigs: FTDX 1200, FTDX 3000, FTDX 5000, FT-891 and FT-991. Testing has been performed using an FTDX101D, but I will have a possibility to test with FTDX 5000 in a couple of days.


mikaelnousiainen:
This fixes issue #423

PR #433, opened 2020-11-07
Fix max macro. Added missing > operator.

darvark:
Not sure why my gcc wasn't able to find this error. I couldn't find any tests for this lib, to check if this macro is used anywhere.


mdblack98:
You apparently aren't using the HEAD as this patch applies to the old directory structure which changed a  long time ago.
So closing this and making a patch for it.
Good thing this max function wasn't being used at all....it's just there for completeness.

PR #507, opened 2021-01-12
Re-enable Yaesu question mark handling and fix newcat_get_cmd retries in case of question mark response (2nd attempt)

mikaelnousiainen:
@mdblack98 Another attempt at fixing #505 -- please check if the retries work now. I can also add conditions for the question_mark flag for particular rig, so please don't close this PR yet as invalid. I just wish to verify that the logic works correctly for all rigs now. Regular set/get VFO/mode/IF etc commands should not be affected by the flag in ANY case.

### DIFF
--- a/rigs/yaesu/newcat.c
+++ b/rigs/yaesu/newcat.c
@@ -4090,14 +4090,32 @@ int newcat_get_level(RIG *rig, vfo_t vfo, setting_t level, value_t *val)
         snprintf(priv->cmd_str, sizeof(priv->cmd_str), "KS%c", cat_term);
         break;
 
-    case RIG_LEVEL_MICGAIN:
+    case RIG_LEVEL_MICGAIN: {
+        pbwidth_t width;
+        rmode_t mode = 0;
+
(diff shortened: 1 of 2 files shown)


mdblack98:
Please make this rig and mode specific.;
I'm not going to accept affecting every other Yaesu rig disabling retries by the lack of capability of a couple of them.  You also don't want to disable retires on the 5000 or 101 when not in RTTY mode.

mikaelnousiainen:
Sure, but please keep the PRs open, I can add commits to them. :)

mikaelnousiainen:
@mdblack98 Please have a look at the code now. I had previously verified the same set of invalid commands for FT-991 and FTDX 5000. The same behaviour may apply for the new FTDX10 too BTW!

Caching the mode would be useful for this purpose, but I'm not aware of such functionality in Hamlib...

mikaelnousiainen:
The fixes regarding the value of `rc` variable in `newcat_get_cmd` now make the retries work (in case the flag isn't set) -- I verified it too!

Issue #612, opened 2021-03-16
WSJT-X commands failing randomly on latest master


mikaelnousiainen:
I pulled the latest changes (commit 5dfb3b85671e8542a0016aae4d63f5ef76bab023) and now WSJT-X rigctld commands (control over network via Hamlib NET rigctl) fail very randomly. The commands and responses seem to vary, I've been getting "invalid parameter", "I/O error" and "command rejected by the rig". It all seems like it's got something to do with timings -- maybe the latest changes with caching affect this, so could #610 be related?

I can usually make the commands fail when clicking the test CAT/PTT buttons in WSJT-X settings. During actual use of WSJT-X, it may take a while until it fails. None of this has happened for me before. I've been running the same WSJT-X v2.2.2 for a long time and it's been stable.

Here's an example of IC-7600 failing to set PTT off in WSJT-X settings when testing PTT (apologies for the extra mess, it's the application logging):

```
64,"process":"/usr/local/bin/rigctld","line":"rigctl(d): T 'currVFO' '0' '' ''\n"}
64,"process":"/usr/local/bin/rigctld","line":"rigctl_parse.c(642):rigctl_parse entered\n"}
64,"process":"/usr/local/bin/rigctld","line":"rigctl_parse: vfo_opt=0\n"}
64,"process":"/usr/local/bin/rigctld","line":"rigctl_parse: called, interactive=1\n"}
64,"process":"/usr/local/bin/rigctld","line":"rigctl_parse.c(2279):rigctl_set_ptt entered\n"}
64,"process":"/usr/local/bin/rigctld","line":"rigctl_parse: cmd==0x0a\n"}
64,"process":"/usr/local/bin/rigctld","line":"rig.c(2602):rig_set_ptt entered\n"}
64,"process":"/usr/local/bin/rigctld","line":"icom_set_ptt called\n"}
64,"process":"/usr/local/bin/rigctld","line":"frame.c(325):icom_transaction entered\n"}
64,"process":"/usr/local/bin/rigctld","line":"icom_transaction: cmd=0x1c, subcmd=0x00, payload_len=1\n"}
64,"process":"/usr/local/bin/rigctld","line":"frame.c(119):icom_one_transaction entered\n"}
64,"process":"/usr/local/bin/rigctld","line":"frame.c(56):make_cmd_frame entered\n"}
64,"process":"/usr/local/bin/rigctld","line":"frame.c(90):make_cmd_frame return(8)\n"}
[...]

PR #614, opened 2021-03-17
Fix unnecessary rig re-open cycles by refactoring rigctl_parse return values

mikaelnousiainen:
The current command loop in both `rigctl` and the `rigctld` server perform rig close/open cycles in case of hardware or I/O errors. However, the logic to detect these errors is not very accurate, as `rigctl_parse` returns value `2` (indicating re-open is required) for all `RIG_E*` errors from `rig_` commands. There are many errors, e.g. "invalid parameter", which do not indicate an issue in the rig communication and re-opening the rig is totally unnecessary in this case.

Re-opening the rig causes real trouble when there are multiple clients connected to `rigctld` server and one of the clients issues a command resulting in "invalid parameter" error, for example. This will lead to the rig being unavailable for some time all other clients. See #612 for a more detailed explanation.

This PR includes:

1) New, more deterministic return values in `rigctl_prase` function. The function returns the negative `RIG_E*` errors as they are, uses `RIG_OK` for indicating a successful command and two positive integers `RIGCTL_PARSE_END` (1) and `RIGCTL_PARSE_ERROR` (2) to indicate end-of-stream and other internal parsing errors -- both of which will make the command loop exit.

2) New macro `RIG_IS_SOFT_ERRCODE` for determining if a particular error code is a "soft error", indicating it is not a hardware or I/O fault and cannot be potentially fixed by re-opening the rig. The soft errcodes are: RIG_EINVAL, RIG_ENIMPL, RIG_ERJCTED, RIG_ETRUNC, RIG_ENAVAIL, RIG_ENTARGET, RIG_EVFO and RIG_EDOM.

3) Some fixes to Icom backend to remove ACK/NAK detection in commands that get/read data from the rig, as the ACK/NAK bytes are present in responses to commands that set/change the rig state only.

### DIFF
--- a/include/hamlib/rig.h
+++ b/include/hamlib/rig.h
@@ -141,6 +141,15 @@ enum rig_errcode_e {
 };
 
 /**
+ * \brief Determines if the given error code indicates a "soft" error
+ * Soft errors are caused by invalid parameters and software/hardware features
+ * and cannot be fixed by retries or by re-initializing hardware.
+ */
+#define RIG_IS_SOFT_ERRCODE(errcode) (errcode == RIG_EINVAL || errcode == RIG_ENIMPL || errcode == RIG_ERJCTED \
+    || errcode == RIG_ETRUNC || errcode == RIG_ENAVAIL || errcode == RIG_ENTARGET \
+    || errcode == RIG_EVFO || errcode == RIG_EDOM)
+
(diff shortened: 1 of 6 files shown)


mikaelnousiainen:
That was fast! Thanks :)

PR #658, opened 2021-04-12
Add Radant AZV-1 Rotator controller protocol.

b-o-r-m-a-l-e-y:
Hello guys, I have added in rotctl(d) basic Radant AZV-1 support. Tested it, and it looks pretty good. You can read about this rotator and protocols [here](https://xn--80adyobadlt.xn--p1ai/download_center/). As all info in Russian, I quickly translated it into .txt file you can see in my commit.


mdblack98:
The mfg_name would seem to be Radant and the model AZ-1/AZV-1, correct?

b-o-r-m-a-l-e-y:
> The mfg_name would seem to be Radant and the model AZ-1/AZV-1, correct?

Yes, that`s correct.

PR #691, opened 2021-05-05
Fix rig_get_ant NULL check

mikaelnousiainen:
The `rig_get_ant` function doesn't work currently, as there is a typo in the code. This fixes it.

However, I'm wondering if there should be null checks for `ant_tx` and `ant_rx` too? Now the code simply sets value to those pointers without any checks.


mikaelnousiainen:
@mdblack98  Thanks for applying the fix.

One more thing: The line `*ant_tx = *ant_rx = RIG_ANT_UNKNOWN;` needs to be moved after the NULL checks on `rig_get_ant`.

PR #769, opened 2021-08-17
Move priv_data structs of ft100, ft600, ft817, ft857, ft897 to the C file

wutje:
Note that is already so for all other Yaesu device.
Leave private data in newcat; it is used by multiple rigs.

### DIFF
--- a/rigs/yaesu/ft100.c
+++ b/rigs/yaesu/ft100.c
@@ -41,6 +41,13 @@
 #include "misc.h"
 #include "bandplan.h"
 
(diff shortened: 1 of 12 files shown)


mdblack98 (approved):
Looks OK.

wutje:
I see this PR contains more commits then I intended. Sorry for that, but thanks for merging!
I tested the VFO setting on a FT818, but the KA7OEI EEPROM documents only the 817 so I guess it works for a 817 as well.

PR #777, opened 2021-08-28
Corrected issue where transmit VFO could not be set in split mode.

KaComet:
Corrected an issue with setting an FT-736R's TX VFO while in split mode.


KaComet:
NVM. commit f552e3f0e8900e9285b23770a590e07cbc16b90a corrected this issue.

PR #780, opened 2021-08-30
Yaesu cmnd indirection removal

wutje:
Most Yaesu rigs have a private command table, which they pointlessly copy to their private structure where it is never written.
So drop the copy step and use the table direct; much easier to read and understand.
Also drop single use variable rig_s in most the of the same function already touched.


mdblack98:
Deferring until 4.4 in just a couple of days

wutje:
Okay, no change intended but I understand.

PR #790, opened 2021-09-07
Fix compile-time errors in Android NDK environment

AuroraRAS:
Fixed some errors. It seems that the NDK files are outdated and are not synchronized with the code of the master branch.


N0NB:
Cherry-picked all commits into Hamlib-4.3 branch.

PR #793, opened 2021-09-09
Improve doc

wutje:
Fix some Doxygen issues.
Make functions static to stop Doxygen from generating documentation for them.
Add documentation on get_cache. I hope I understood this correct.

### DIFF
--- a/scripts/build-w32.sh
+++ b/scripts/build-w32.sh
@@ -129,9 +129,9 @@ directory.
 
 In short, the command syntax is of the form:
 
-  rigctl -m 120 -r COM1 -vvvvv
+  rigctl -m 1020 -r COM1 -vvvvv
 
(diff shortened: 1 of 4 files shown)


mdblack98 (changes requested):
RETURNFUNC(RIG_OK) is not really the correct documentation from a caller's perspective.
The call simply returns RIG_OK as currently documented or an  error.  The macro call is transparent so I wouldn't' document it that way.

wutje:
That is exactly what I did?
The macro was added by accident I guess during a mass replacement. So I "reverted" that change.

wutje:
Ah I see I totally forgot to mention it breaks Doxygen documentation. Just take a look at the current online documentation for rig.c and its functions. That is what the change is fixing.

PR #804, opened 2021-09-20
Fix rig_cookie: use lock for reads and write.

### LINKED ISSUE #802: What is the intended use case of rig_cookie?
I think I understand what rig_cookie is trying to achieve: implement some sort of locking to allow multi-threaded applications.
But I think there are some serious issues;

1. Locking is only done when having pthread, so if that is not enabled there is no multi thread guarantee.
2. Even if pthread is available the lock should also be used when reading cookie to protect against writes, not just when setting.
3. The lock should also protect timestamp.
4. Fetching timestamp should be done AFTER getting the lock. otherwise significant time could have past waiting for the lock.

To me it really seems like a problem that should be solved at the application layer. Why is hamlib trying to provide this? Are there any known users?

wutje:
Ensure we never print more then HAMLIB_COOKIE_SIZE otherwise we read out-of-bounds.
Drop stray printf.
Add tests for invalid input and overly large input.
Fix test2 to release the cookie.

This fixes #802

### DIFF
--- a/src/rig.c
+++ b/src/rig.c
@@ -6791,16 +6791,16 @@ int HAMLIB_API rig_cookie(RIG *rig, enum cookie_e cookie_cmd, char *cookie,
     // only 1 client can have the cookie so these can be static
     // this should also prevent problems with DLLs & shared libraies
     // the debug_msg is another non-thread-safe which this will help fix
-    // 27 char cookie will last until the year 10000
     static char
-    cookie_save[HAMLIB_COOKIE_SIZE];  // only one client can have the 26-char cookie
+    cookie_save[HAMLIB_COOKIE_SIZE];  // only one client can have the cookie
     static double time_last_used;
-    double time_curr;
     struct timespec tp;
+    int ret;
 #ifdef HAVE_PTHREAD
     static pthread_mutex_t cookie_lock = PTHREAD_MUTEX_INITIALIZER;
 #endif
 
+    /* This is not needed for RIG_COOKIE_RELEASE but keep it simple. */
     if (cookie_len < HAMLIB_COOKIE_SIZE)
     {
         rig_debug(RIG_DEBUG_ERR, "%s(%d): cookie_len < %d\n",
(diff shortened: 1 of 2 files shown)


mdblack98:
I don't think we want to put a mutex in there as the only solution.  The idea of a cookie is if it gets refused then you can do other things before asking again or just skip the request....e.g. inside a wait loop or such rather than getting stuck on the mutex request.
This cookie idea would be implemented in hamlib 5.0 so that no request could come in without holding the cookie.
Another possibility is to use the mutex if a cookie has not (ever) been requested by the client -- that would provide backwards compatibility and could be implemented in hamlib 4.4.
This approach might also allow multiple clients to talk to one serial port if we can make all operations atomic.  So another client opening the serial port would get a fake OK.  Sounds possible.

wutje:
The mutex is for doing an cookie operation nothing else.
So thread A gets the mutex and then starts a get operation. Thread B will have to wait a little before thread A is done with grabbing the cookie. But after that short while it will run and gets the result the cookie is taken already. And it can do what ever it wants.
For the cookie to work properly the access on cookie and the timestamp needs to be atomic. That is fixed by the lock.

wutje:
Also note that for the "cookie get" command the mutex usage did not change. I did not intend to change any logic flow with regard to the mutex.

mdblack98:
We're going to need a bunch more locks...inside every rig_ function in rig.c unique to each function.But this is a good start to help keep things thread safe.
There's one area I know is not thread safe and that's the debug msgs.  It's a global buffer right now so could be clobbered by threading.  Not likely to happen as retreiving debug should be very fast compared to any rig actions but with the new caching that will be happening due to the multicast it will be more likely.
Mike W9MDB

 

    On Tuesday, September 21, 2021, 06:06:53 AM CDT, Wouter ***@***.***> wrote:  
 
 


Also note that for the "cookie get" command the mutex usage did not change. I did not intend to change any logic flow with regard to the mutex.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

PR #910, opened 2021-12-29
Fix spurious writes when no RT-21 second port

arodland:
The if condition for whether there actually is a second port is only
covering the sprintf for the elevation command, and not the write
attempt. So we call rotorez_send_priv_cmd2 whether rotport2 is valid or
not. When it's not, we try to write the command to fd 0 (stdin). When
running on the console, this results in trash on the console. When
runnning under something else (e.g. systemd) this results in errors or
other wierd behavior (e.g. EBADF because stdin isn't connected, causing
the command to fail with RIG_EIO).

### DIFF
--- a/rotators/rotorez/rotorez.c
+++ b/rotators/rotorez/rotorez.c
@@ -493,15 +493,14 @@ static int rt21_rot_set_position(ROT *rot, azimuth_t azimuth,
     {
         sprintf(cmdstr, "AP1%05.1f\r;",
                 elevation);    /* Total field width of 5 chars */
+
+        err = rotorez_send_priv_cmd2(rot, cmdstr);
+
+        if (err != RIG_OK)
+        {
+            return err;
+        }
     }
-
-    err = rotorez_send_priv_cmd2(rot, cmdstr);
-
-    if (err != RIG_OK)
-    {
-        return err;
-    }
-
     return RIG_OK;
 }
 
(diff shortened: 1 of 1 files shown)


mdblack98:
Thanks Andrew.

Mike W9MDB




On Wednesday, December 29, 2021, 12:59:42 PM CST, Andrew Rodland ***@***.***> wrote: 





The if condition for whether there actually is a second port is only
covering the sprintf for the elevation command, and not the write
attempt. So we call rotorez_send_priv_cmd2 whether rotport2 is valid or
not. When it's not, we try to write the command to fd 0 (stdin). When
running on the console, this results in trash on the console. When
runnning under something else (e.g. systemd) this results in errors or
other wierd behavior (e.g. EBADF because stdin isn't connected, causing
the command to fail with RIG_EIO).

  https://github.com/Hamlib/Hamlib/pull/910

Commit Summary

     
    * 8bf26f1 Fix spurious writes when no RT-21 second port
    
File Changes  (1 file) 
     
    *  M rotators/rotorez/rotorez.c (12) 
    

Patch Links:

     
    * https://github.com/Hamlib/Hamlib/pull/910.patch
    
    * https://github.com/Hamlib/Hamlib/pull/910.diff
    

—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

PR #913, opened 2022-01-02
Enable "stop" command for RT-21 rotator.

arodland:
I got a request from G0GJV to see about why the RT-21 driver doesn't support the "stop" command. It looks like it was commented out when the RT-21 caps were added back in 2014, and hasn't been touched since. I uncommented it and it works 100% fine with my unit (and the manual agrees that sending a bare `;` is the stop command). Might as well enable it?


mdblack98 (changes requested):
Please update the .version too

PR #976, opened 2022-02-28
do not search for tokens if name does not look like a number

pidloop:


mdblack98:
That's the hard way to determine a number....here's an easier way
int tmpnum;
char tmpbuf[2];
if (sscanf(name, "%d%1s", &tmpnum,tmpbuf)==1) name_isa_number = 1
else name_isa_number = 0

Or make an alldigits function and use that
int alldigits(char *s)
{
    int i;
    for(i=0;s[i]!=0;++i)
    {
      if (!isdigit(s[i])) return 0;
    }
    return 1;
}

name_isa_number = alldigits(cmd);

pidloop:
That's ridiculous, this usage of strtol is the main reason it returns a pointer.

But you're the supreme boss.. I'll close this pull.

pidloop:
One important aspect of good programming is maintainability. One way to
achieve that is using established patterns when possible. The usage of
strtol to check for all digits is such a pattern, so much so it’s even
described in its man page, look for yourself.

Inventing solutions when there are established patterns just adds noise.

PR #1022, opened 2022-05-07
bindings: Fix tcl build with slibtool

orbea:
Gentoo issue: https://bugs.gentoo.org/798273

When building hamlib with `--with-tcl-binding` and slibtool (https://dev.midipix.org/cross/slibtool) it will fail with many undefined references for tcl.

This is because the `TCL_LIBS` is never set correctly and it uses `-no-undefined` during the build. GNU libtool does not reveal this problem because it will silently ignore `-no-undefined`. This patch will use `PKG_CHECK_MODULES` to find the linker flags and sets them where needed.

Full build log: [hamlib.slibtool.log](https://github.com/Hamlib/Hamlib/files/8644100/hamlib.slibtool.log)

### DIFF
--- a/bindings/Makefile.am
+++ b/bindings/Makefile.am
@@ -149,8 +149,9 @@ PKG_VER = $(ABI_VERSION).$(ABI_REVISION)
 DLL = hamlibtcl-$(PKG_VER)$(TCL_SHLIB_SUFFIX)
 
 nodist_hamlibtcl_la_SOURCES = hamlibtcl_wrap.c
+hamlibtcl_la_CFLAGS = $(TCL_CFLAGS)
 hamlibtcl_la_LDFLAGS = -no-undefined -module -release $(PKG_VER) -avoid-version
-hamlibtcl_la_LIBADD = $(top_builddir)/src/libhamlib.la $(TCL_LIB_SPEC)
+hamlibtcl_la_LIBADD = $(top_builddir)/src/libhamlib.la $(TCL_LIB_SPEC) $(TCL_LIBS)
 
 hamlibtcl_ladir = $(tcldir)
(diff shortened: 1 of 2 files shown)


mdblack98:
Please test this patch with libtool.
It fails here now so I reverted this merge.  Without this patch --with-tcl-binding compiles just fine.
configure: error: Unable to find Tcl pkgconfig
Mike W9MDB

 

    On Friday, May 6, 2022, 07:18:12 PM CDT, orbea ***@***.***> wrote:  
 
 
Gentoo issue: https://bugs.gentoo.org/798273

When building hamlib with --with-tcl-binding and slibtool (https://dev.midipix.org/cross/slibtool) it will fail with many undefined references for tcl.

This is because the TCL_LIBS is never set correctly and it uses -no-undefined during the build. GNU libtool does not reveal this problem because it will silently ignore -no-undefined.

Full build log: hamlib.slibtool.log

  https://github.com/Hamlib/Hamlib/pull/1022

Commit Summary
   
   - 9dfbd23 bindings: Fix tcl build with slibtool

File Changes 
 
(2 files)
   
   -  M bindings/Makefile.am (3) 
   -  M configure.ac (7) 

Patch Links:
   
   - https://github.com/Hamlib/Hamlib/pull/1022.patch
   - https://github.com/Hamlib/Hamlib/pull/1022.diff

orbea:
Hi,

Sorry about the issue and I did test this with GNU libtool, but I didn't realize that tcl doesn't always ship with a pkgconfig file. However gentoo lacks `tclConfig.sh` which is what the current check is looking for.

A solution could be to check for `tcl.pc` as done in this PR and fallback to finding `tclConfig.sh` instead of an error if its missing.

PR #1023, opened 2022-05-07
bindings: Fix tcl build with slibtool

orbea:
This is the second attempt at PR https://github.com/Hamlib/Hamlib/pull/1022, please confirm it works for you.

Gentoo issue: https://bugs.gentoo.org/798273

When building hamlib with `--with-tcl-binding` and slibtool (https://dev.midipix.org/cross/slibtool) it will fail with many undefined references for tcl.

This is because the `TCL_LIBS` is never set correctly and it uses `-no-undefined` during the build. GNU libtool does not reveal this problem because it will silently ignore `-no-undefined`. This patch will use `PKG_CHECK_MODULES` to find the linker flags and sets them where needed.

The difference with the original PR is that instead of an error when `tcl.pc` is not found it will then check for `tclConfig.sh` instead. I also added an error for when the `SC_LOAD_TCLCONFIG` macro fails to load the config file in case both the pkgconfig file and `tclConfig.sh` are missing while the `tcl.h` header is present.

Full build log: [hamlib.slibtool.log](https://github.com/Hamlib/Hamlib/files/8644100/hamlib.slibtool.log)

### DIFF
--- a/bindings/Makefile.am
+++ b/bindings/Makefile.am
@@ -149,8 +149,9 @@ PKG_VER = $(ABI_VERSION).$(ABI_REVISION)
 DLL = hamlibtcl-$(PKG_VER)$(TCL_SHLIB_SUFFIX)
 
 nodist_hamlibtcl_la_SOURCES = hamlibtcl_wrap.c
+hamlibtcl_la_CFLAGS = $(TCL_CFLAGS)
 hamlibtcl_la_LDFLAGS = -no-undefined -module -release $(PKG_VER) -avoid-version
-hamlibtcl_la_LIBADD = $(top_builddir)/src/libhamlib.la $(TCL_LIB_SPEC)
+hamlibtcl_la_LIBADD = $(top_builddir)/src/libhamlib.la $(TCL_LIB_SPEC) $(TCL_LIBS)
 
 hamlibtcl_ladir = $(tcldir)
 hamlibtcl_la_DATA = pkgIndex.tcl
--- a/configure.ac
+++ b/configure.ac
@@ -589,11 +589,16 @@ AC_ARG_WITH([tcl-binding],
     [build_tcl=no])
 AC_MSG_RESULT([$build_tcl])
 
-dnl SC_PATH_TCLCONFIG and SC_LOAD_TCLCONFIG from macros/tcl.m4
+dnl tcl.pc or SC_PATH_TCLCONFIG and SC_LOAD_TCLCONFIG from macros/tcl.m4
 AS_IF([test x"${build_tcl}" = "xyes"],[
-    dnl Search for and load tclConfig.sh.
-    SC_PATH_TCLCONFIG
-    SC_LOAD_TCLCONFIG
+    dnl Search for and load tcl.pc or tclConfig.sh.
+    PKG_CHECK_MODULES([TCL], [tcl],
+        [],
+        [
+            AC_MSG_WARN([Unable to find Tcl pkgconfig])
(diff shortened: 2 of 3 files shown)


mdblack98:
Nope...building TCL only requires tcl-dev and not the tcl package itself.
checking Whether to build Tcl bindings... yes
checking for Tcl configuration... found /usr/lib/tclConfig.sh
checking for existence of /usr/lib/tclConfig.sh... loading
checking tcl.h usability... yes
checking tcl.h presence... yes
checking for tcl.h... yes
checking for tcl... no
configure: error: Unable to find Tcl pkgconfig


Try keeping just tcl-dev installed and see if you can duplicate the error
Could be the tcl.m4 macro needs to be updated.
Mike W9MDB
 

    On Saturday, May 7, 2022, 12:18:58 AM CDT, orbea ***@***.***> wrote:  
 
 
This is the second attempt at PR #1022, please confirm it works for you.

Gentoo issue: https://bugs.gentoo.org/798273

When building hamlib with --with-tcl-binding and slibtool (https://dev.midipix.org/cross/slibtool) it will fail with many undefined references for tcl.

This is because the TCL_LIBS is never set correctly and it uses -no-undefined during the build. GNU libtool does not reveal this problem because it will silently ignore -no-undefined. This patch will use PKG_CHECK_MODULES to find the linker flags and sets them where needed.

The difference with the original PR is that instead of an error when tcl.pc is not found it will then check for tclConfig.sh instead. I also added an error for when the SC_LOAD_TCLCONFIG macro fails to load the config file in case both the pkgconfig file and tclConfig.sh are missing while the tcl.h header is present.

Full build log: hamlib.slibtool.log

  https://github.com/Hamlib/Hamlib/pull/1023

Commit Summary
   
   - 4fa9a21 bindings: Fix tcl build with slibtool

File Changes 
 
(3 files)
   
   -  M bindings/Makefile.am (3) 
   -  M configure.ac (16) 
   -  M macros/tcl.m4 (1) 

Patch Links:
   
   - https://github.com/Hamlib/Hamlib/pull/1023.patch
   - https://github.com/Hamlib/Hamlib/pull/1023.diff

orbea:
> configure: error: Unable to find Tcl pkgconfig

Are you sure you are testing the current patch? This should not be an error and should be a warning now.

https://github.com/Hamlib/Hamlib/pull/1023/files#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810R598

It should only fail if the `tcl.pc` pkgconfig file is missing and then it also fails to load `tclConfig.sh`.

https://github.com/Hamlib/Hamlib/pull/1023/files#diff-f1bbd3d58f30c67495f335628fcafecba2799dfcfa310c3c7c026718a05ce891R133

mdblack98:
I  must've used the wrong patch.  It does work now.
merged

Mike W9MDB







On Saturday, May 7, 2022, 10:12:06 AM CDT, orbea ***@***.***> wrote: 







>  
> configure: error: Unable to find Tcl pkgconfig
> 

Are you sure you are testing the current patch? This should not be an error and should be a warning now.

https://github.com/Hamlib/Hamlib/pull/1023/files#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810R598

It should only fail if the tcl.pc pkgconfig file is missing and then it also fails to load tclConfig.sh.

https://github.com/Hamlib/Hamlib/pull/1023/files#diff-f1bbd3d58f30c67495f335628fcafecba2799dfcfa310c3c7c026718a05ce891R133

PR #1045, opened 2022-05-31
Dummy rotator max_az = 450

PianetaRadio:
I propose to extent dummy rotator max_az to 450 degrees (was 180) to better emulate rotators with positive only azimuth position.


mdblack98:
Seems harmless enough

Issue #1082, opened 2022-07-12
Setting of frequency with Kenwood TM-D710G slightly broken


mikaelnousiainen:
After changes in #1071 TM-D710G sets the frequency for the opposite VFO always. I have not figured out a way to fix this.

As I have the mentioned rig, I will try to get to the bottom of this and fix it -- while trying to preserve the "targetability" of VFO A/B frequencies for satellite work.

mikaelnousiainen:
Fixed in #1095

PR #1095, opened 2022-07-28
Fix current and split VFO frequency handling for Kenwood TM-D710(G)

mikaelnousiainen:
This fixes handling of current and split VFO frequencies on Kenwood TM-D710(G).

The code uses CTRL VFO in the rig for denoting current VFO and PTT VFO for split VFO.

`\set_split_freq` command now always sets the frequency of the TX VFO, which can be changed using `\set_split_vfo` command. This works now the way most other rigs handle split too.

To enable split, one can simply do `\set_split_vfo 1 VFOB` (when current VFO is VFOA). I've tested opposite use of VFOs too (current = VFOB, TX = VFOA).

### DIFF
--- a/rigs/kenwood/tmd710.c
+++ b/rigs/kenwood/tmd710.c
@@ -652,13 +652,18 @@ typedef struct
 
 static int tmd710_open(RIG *rig)
 {
+    split_t split;
+    vfo_t vfo;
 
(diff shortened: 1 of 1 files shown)


mikaelnousiainen:
This fixes #1082

mikaelnousiainen:
@RhinoLance Please test if this code works for you in split / gpredict use

PR #1129, opened 2022-10-21
fix build instructions in INSTALL

ckuhtz:
./bootstrap and autoconf are missing. the is broken as-is and now fixed.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -15,6 +15,8 @@ main directory and do the following:
 
   1. Configure the source code by typing:
 
+        $ ./bootstrap
+        $ autoconf
         $ ./configure
 
      If configure does not exist you can create it with ./bootstrap


ckuhtz (commented):
fixed build instructions

N0NB:
Actually, the original INSTALL file is correct.  The instructions for INSTALL are intended for the release tarball and not a Git clone.  README.developer should have the correct information as developers are expected to be working from a cloned Git working tree.

ckuhtz:
However you want to handle this is fine by me, but this doesn't make any sense.  Whether tarball or  git clone, the dependency created is to build the configure file through the bootstrap script.  The "three step" salute doesn't work without it.  You can either make it easy for people to find the right build instructions in README and INSTALL or waste their time looking for answers.  This stuff should build simply and easily as it does when it's made clear that bootstrap script is required.

N0NB:
Mike already addressed this.  To be clear, a release tarball has no requirement to run the bootstrap script.  I am talking about the ones that have been tagged and signed by my GPG key as actual releases, not the .ZIP  files offered by GitHub which, unfortunately, we cannot disable.

See the text for the 4.4 release:  https://github.com/Hamlib/Hamlib/releases/tag/4.4

ckuhtz:
I git cloned the repo. IMHO, the recompiling instructions in README need to be amended or removed as they are incomplete. They appear to be complete on their face, however.

mdblack98:
I did add a bit to help clarify things.  Did you see the change in INSTALL?

 

    On Friday, October 21, 2022 at 03:57:19 AM CDT, ckuhtz ***@***.***> wrote:  
 
 


I git cloned the repo. IMHO, the recompiling instructions in README need to be amended or removed as they are incomplete. They appear to be complete on their face, however.

PR #1148, opened 2022-11-14
Yaesu FT-991 NR level

PianetaRadio:
Fix NR level range for Yaesu FT-991


mdblack98:
The ft991.c patch is incorrect -- the range is supposed to be 0 to 1 meaning 0=min and 1=max.
The underlying rig behavior is abstracted.    The step size would be 1.0/15.0
The change to newcat.c looks correct.  There we have to know what the rig can do.
Please change the LEVEL_NR line in ft991.c to reflect the proper 0-1 range and step size.

PR #1166, opened 2022-11-30
Yaesu NB2

PianetaRadio:
From FT-950 CAT manual, there are two NB options:

NB P1 P2 ;
P1 0: Fixed
P2 0: Noise Blanker “OFF”
1: Noise Blanker “ON”
2: Noise Blanker (Wide) “ON”

Maybe also for other Yaesu rigs.


PianetaRadio:
I forgot to increment newcat and backend version

PR #1242, opened 2023-03-04
Fix unterminated comment

dforsi:
Fixes a cppcheck error:
rigs/mds/serialnum.c:106:1: error: Unmatched '}'. Configuration: ''. [syntaxError]


dforsi:
There is a mismatch in `serialnum.c` that I don't know how to fix:

at line 34
```    if (s[5] == 'B') { operation = "Base"; }```
but at line 55
```5 = X/N Base/Remote```

PR #1347, opened 2023-07-28
Fix segfault on rigctl exit

GeoBaltz:
If rigctl does a recovery close/open cycle, a freed buffer gets reused. This can cause segfault or worse.


GeoBaltz:
Fix use after free

PR #1413, opened 2023-10-31
Refactor multicast publisher/server code

mikaelnousiainen:
@mdblack98 NOTE: This is a draft PR until I get to test the code properly on Windows too. It seems also the UDP server has some bug on Linux, but didn't spot it yet :)

This PR addresses the following:

* Create a separate "receiver" server routine for multicast command handling (move it out of the publisher, they're not related) -> See the TODOs in the `multicast_receiver()` function for the actual implementation
* Move the rig cache change detection to the rig poll routine (see event.c now!) and enable it by default -> publishing of rig state changes via UDP seems to work
* Remove conflicting mutlicast publisher initialization from rigctld daemon
* Add rig set-conf address/port parameters for both multicast data publisher and the command server

### DIFF
--- a/include/hamlib/rig.h
+++ b/include/hamlib/rig.h
@@ -2758,7 +2758,13 @@ struct rig_state {
     volatile int morse_data_handler_thread_run;
     void *morse_data_handler_priv_data;
     FIFO_RIG *fifo_morse;
-    int port_multicast;  /*!< May be different so this is initially a copy of rigctl'd port selection */
+    char *multicast_data_addr;  /*!< Multicast data UDP address for publishing rig data and state */
+    int multicast_data_port;  /*!< Multicast data UDP port for publishing rig data and state */
+    char *multicast_cmd_addr;  /*!< Multicast command server UDP address for sending commands to rig */
+    int multicast_cmd_port;  /*!< Multicast command server UDP port for sending commands to rig */
+
+    volatile int multicast_receiver_run;
+    void *multicast_receiver_priv_data;
 };
 
 /**
--- a/src/conf.c
+++ b/src/conf.c
@@ -208,7 +208,7 @@ static const struct confparams frontend_cfg_params[] =
     {
         TOK_MULTICAST_CMD_PORT, "multicast_cmd_port", "Multicast command server UDP port",
         "Multicast data UDP port for sending commands to rig",
-        "4531", RIG_CONF_NUMERIC, { .n = { 0, 1000000, 1 } }
+        "4532", RIG_CONF_NUMERIC, { .n = { 0, 1000000, 1 } }
     },
 
     { RIG_CONF_END, NULL, }
--- a/src/conf.c
+++ b/src/conf.c
@@ -208,7 +208,7 @@ static const struct confparams frontend_cfg_params[] =
     {
         TOK_MULTICAST_CMD_PORT, "multicast_cmd_port", "Multicast command server UDP port",
         "Multicast data UDP port for sending commands to rig",
-        "4532", RIG_CONF_NUMERIC, { .n = { 0, 1000000, 1 } }
+        "4531", RIG_CONF_NUMERIC, { .n = { 0, 1000000, 1 } }
     },
 
     { RIG_CONF_END, NULL, }
--- a/src/conf.c
+++ b/src/conf.c
@@ -92,8 +92,8 @@ static const struct confparams frontend_cfg_params[] =
     },
     {
         TOK_POLL_INTERVAL, "poll_interval", "Rig state poll interval in milliseconds",
-        "Polling interval in milliseconds for transceive emulation, value of 0 disables polling",
-        "0", RIG_CONF_NUMERIC, { .n = { 0, 1000000, 1 } }
+        "Polling interval in milliseconds for transceive emulation, defaults to 1000, value of 0 disables polling",
+        "1000", RIG_CONF_NUMERIC, { .n = { 0, 1000000, 1 } }
     },
     {
         TOK_PTT_TYPE, "ptt_type", "PTT type",
@@ -189,6 +189,26 @@ static const struct confparams frontend_cfg_params[] =
         TOK_OFFSET_VFOB, "offset_vfob", "Offset value in Hz",
         "Add Hz to VFOB/Sub frequency set",
         "0", RIG_CONF_NUMERIC, { .n = {0, 1e12, 1}}
+    },
+    {
+        TOK_MULTICAST_DATA_ADDR, "multicast_data_addr", "Multicast data UDP address",
+        "Multicast data UDP address for publishing rig data and state, value of 0.0.0.0 disables multicast data publishing",
+        "224.0.0.1", RIG_CONF_STRING,
+    },
+    {
+        TOK_MULTICAST_DATA_PORT, "multicast_data_port", "Multicast data UDP port",
+        "Multicast data UDP port for publishing rig data and state",
+        "4532", RIG_CONF_NUMERIC, { .n = { 0, 1000000, 1 } }
+    },
+    {
+        TOK_MULTICAST_CMD_ADDR, "multicast_cmd_addr", "Multicast command server UDP address",
+        "Multicast command UDP address for sending commands to rig, value of 0.0.0.0 disables multicast command server",
+        "224.0.0.2", RIG_CONF_STRING,
+    },
+    {
+        TOK_MULTICAST_CMD_PORT, "multicast_cmd_port", "Multicast command server UDP port",
+        "Multicast data UDP port for sending commands to rig",
+        "4532", RIG_CONF_NUMERIC, { .n = { 0, 1000000, 1 } }
     },
 
     { RIG_CONF_END, NULL, }
@@ -608,7 +628,11 @@ static int frontend_set_conf(RIG *rig, token_t token, const char *val)
         break;
 
     case TOK_POLL_INTERVAL:
-        rs->poll_interval = atof(val);
+        if (1 != sscanf(val, "%ld", &val_i))
+        {
+            return -RIG_EINVAL;
+        }
+        rs->poll_interval = val_i;
         // Make sure cache times out before next poll cycle
         rig_set_cache_timeout_ms(rig, HAMLIB_CACHE_ALL, atol(val));
         break;
@@ -735,7 +759,31 @@ static int frontend_set_conf(RIG *rig, token_t token, const char *val)
         rig_debug(RIG_DEBUG_VERBOSE, "%s: offset_vfob=%ld\n", __func__, val_i);
         break;
 
+    case TOK_MULTICAST_DATA_ADDR:
+        rs->multicast_data_addr = strdup(val);
+        break;
 
+    case TOK_MULTICAST_DATA_PORT:
+        if (1 != sscanf(val, "%ld", &val_i))
+        {
+            return -RIG_EINVAL;
+        }
+
+        rs->multicast_data_port = val_i;
+        break;
+
+    case TOK_MULTICAST_CMD_ADDR:
+        rs->multicast_cmd_addr = strdup(val);
+        break;
+
+    case TOK_MULTICAST_CMD_PORT:
+        if (1 != sscanf(val, "%ld", &val_i))
+        {
+            return -RIG_EINVAL;
+        }
+
+        rs->multicast_cmd_port = val_i;
+        break;
 
     default:
         return -RIG_EINVAL;
--- a/src/event.c
+++ b/src/event.c
@@ -87,6 +87,10 @@ void *rig_poll_routine(void *arg)
     // Rig cache time should be equal to rig poll interval (should be set automatically by rigctld at least)
     rig_set_cache_timeout_ms(rig, HAMLIB_CACHE_ALL, rs->poll_interval);
 
+    // Attempt to detect changes with the interval below (in milliseconds)
+    int change_detection_interval = 50;
+    int interval_count = 0;
+
     update_occurred = 0;
 
     while (rs->poll_routine_thread_run)
@@ -333,9 +337,18 @@ void *rig_poll_routine(void *arg)
         {
             network_publish_rig_poll_data(rig);
             update_occurred = 0;
+            interval_count = 0;
         }
 
-        hl_usleep(rs->poll_interval * 1000);
+        hl_usleep(change_detection_interval * 1000);
+        interval_count++;
+
+        // Publish updates every poll_interval if no changes have been detected
+        if (interval_count >= (rs->poll_interval / change_detection_interval))
+        {
+            interval_count = 0;
+            network_publish_rig_poll_data(rig);
+        }
     }
 
     rig_debug(RIG_DEBUG_VERBOSE, "%s(%d): Stopping rig poll routine thread\n",
--- a/src/event.c
+++ b/src/event.c
@@ -75,12 +75,11 @@ void *rig_poll_routine(void *arg)
     int update_occurred;
 
     vfo_t vfo = RIG_VFO_NONE, vfo_prev = RIG_VFO_NONE;
-    freq_t freq_main = 0, freq_sub = 0, freq_main_prev = 0, freq_sub_prev = 0;
-    rmode_t mode_main = RIG_MODE_NONE, mode_sub = RIG_MODE_NONE,
-            mode_main_prev = RIG_MODE_NONE, mode_sub_prev = RIG_MODE_NONE;
-    pbwidth_t width_main = 0, width_sub = 0, width_main_prev = 0,
-              width_sub_prev = 0;
-    split_t split, split_prev = -1;
+    freq_t freq_main_a = 0, freq_main_b = 0, freq_main_c = 0, freq_sub_a = 0, freq_sub_b = 0, freq_sub_c = 0;
+    rmode_t mode_main_a = 0, mode_main_b = 0, mode_main_c = 0, mode_sub_a = 0, mode_sub_b = 0, mode_sub_c = 0;
+    pbwidth_t width_main_a = 0, width_main_b = 0, width_main_c = 0, width_sub_a = 0, width_sub_b = 0, width_sub_c = 0;
+    ptt_t ptt = RIG_PTT_OFF;
+    split_t split = RIG_SPLIT_OFF;
 
     rig_debug(RIG_DEBUG_VERBOSE, "%s(%d): Starting rig poll routine thread\n",
               __FILE__, __LINE__);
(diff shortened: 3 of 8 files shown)


mdblack98 (commented):
According to socket(7) man page SO_REUSEADDR is an int -- where do you see it as a char?

mikaelnousiainen:
@mdblack98 MinGW complains about it in Windows... weird.

mdblack98:
32-bit or 64-bit mingw?

 

    On Wednesday, November 1, 2023 at 05:30:28 PM CDT, Mikael Nousiainen ***@***.***> wrote:  
 
 


@mdblack98 MinGW complains about it in Windows... weird.

mikaelnousiainen:
@mdblack98 I'm using the 64-bit compiler. I commented it out, it didn't work even on Linux.

Anyway, the Linux UDP multicast works in both directions, but on Windows the reception fails with this:

```
multicast_receiver: error binding UDP socket to 224.0.0.2:4532: No error
```

Very confusing? Can you test this branch on Windows?

mdblack98:
I had another Windows user seeing that error.
Running message_aggregator in WSJTX  fixed the error.
You have your own fork I can test?

 

    On Wednesday, November 1, 2023 at 05:39:59 PM CDT, Mikael Nousiainen ***@***.***> wrote:  
 
 


@mdblack98 I'm using the 64-bit compiler. I commented it out, it didn't work even on Linux.

Anyway, the Linux UDP multicast works in both directions, but on Windows the reception fails with this:
multicast_receiver: error binding UDP socket to 224.0.0.2:4532: No error

Very confusing? Can you test this branch on Windows?

mdblack98:
And can you try and compile this standalone client/server set?
Run server and client and client will talk back to server.
https://www.dropbox.com/scl/fi/blm71c90dq90c3ovenihg/multicast.tar?rlkey=55mklp80gwkyw2co91ab278rd&dl=0

mikaelnousiainen:
@mdblack98 I wonder what the mesage aggregator does with network calls in Windows? Never used it.

Also, I'm not doing anything with WSJT-X -- just testing plain UDP multicast and sockets.

Multicast UDP sending works fine on Windows too.

My fork/branch is here: https://github.com/mikaelnousiainen/Hamlib/tree/hamlib-multicast-1

mikaelnousiainen:
@mdblack98 Tested your code (Windows 10 + MinGW64), neither of the applications receive anything from the other... I think bind() is still needed, but not sure why it would fail. Confusing. Anyway, need to go offline for now, it's getting late :) Will check back tomorrow.

mikaelnousiainen:
@mdblack98 I think I found a way to make this all work on both Linux and Windows. The workaround is to:

1) Use `INADDR_ANY` when binding (in Windows only) and just subscribing to the multicast group. However, this still allows other traffic like unicast to be received, but rigctl can just ignore any message it doesn't recognize.

2) Use a different port for the multicast commands (multicast receiver) to not consume the multicast state updates / spectrum data sent to port 4532.

It seems that on Windows one cannot bind sockets directly to multicast group addresses. This should be supported -- although discouraged -- based on the official MS documentation here: https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-bind#remarks

> For multicast operations, the preferred method is to call the bind function to associate a socket with a local IP address and then join the multicast group. Although this order of operations is not mandatory, it is strongly recommended. So a multicast application would first select an IPv4 or IPv6 address on the local computer, the wildcard IPv4 address (INADDR_ANY), or the wildcard IPv6 address (in6addr_any). The the multicast application would then call the bind function with this address in the in the sa_data member of the name parameter to associate the local IP address with the socket. If a wildcard address was specified, then Windows will select the local IP address to use. 

However, in practice the bind() calls seem to fail when trying to bind to a multicast address, see:

* https://stackoverflow.com/questions/6140734/cannot-bind-to-multicast-address-windows
* https://cygwin.cygwin.narkive.com/gcVdp3ce/bind-to-multicast-address-fails

So on Windows we'd use the workaround to bind to `INADDR_ANY`.

mikaelnousiainen:
@mdblack98 Please test my latest changes here: https://github.com/mikaelnousiainen/Hamlib/tree/hamlib-multicast-1 (branch `hamlib-multicast-1`)

It will publish state/data updates to `224.0.0.1:4532` and receive+log packet data (with verbose log level) from `224.0.0.2:4531`.

mdblack98:
I don't like the separate port/number -- we don't need multicast on the client side to send anything back -- that can be regular UDP just back to the server (i.e. rigctl, rigctld, hamlib).
I don't see any reason this all can't  live  on one address/port (224.0.0.1/4532 default).  Multiple instances can send multiple packets (separable by Id).

I've tested this standalone and it works here....my HamlibGUI now alllows you to select which Id you want  to watch (future control).

This is receive only so should work with you version...I'll test your branch shortly.

Mike 






On Thursday, November 2, 2023 at 11:01:57 AM CDT, Mikael Nousiainen ***@***.***> wrote: 







@mdblack98 Please test my latest changes here: https://github.com/mikaelnousiainen/Hamlib/tree/hamlib-multicast-1 (branch hamlib-multicast-1)

It will publish state/data updates to 224.0.0.1:4532 and receive+log packet data (with verbose log level) from 224.0.0.2:4531.

mikaelnousiainen:
@mdblack98 Ok, regarding port numbers, maybe we could state what you/we are after here:

1) Is the goal to allow clients to send UDP commands to a constant, known multicast address (e.g. 224.0.0.2:4531 here) so that there is no need to configure a unicast address of rigctl for rig control? This can be only done with multicast. Would be useful?

2) If we use multicast for receiving and there are multiple rigs open, all of them publish state updates to 224.0.0.1:4532, so if the multicast receiver listens to the same multicast group/port, it will receive the multicast publisher state updates, which is not desirable. This is what I tried to avoid with the separate port number. Also, I think even the same instance of Hamlib might actually end up in a "loopback" when publishing multicast packets if it's listening to the same group. Or do you know a way to avoid this?

3) If the command receiver listens to a unicast address only, there should be no issue with "loopback", but clients need to be configured to know the Hamlib unicast IP/port just like with rigctld TCP connections... So just wondering if that is something you would like to avoid as well?

mdblack98:
Clients should only be able to send when they have received a multicast status packet (otherwise they don't know the rig is alive) 
Thinking about this some more if we have multiple rigctl, rigctld, or WSJT-X instances on different rigs or slices (e.g. Flex) we should have each on a separate address at least (not necessarily a separate port).  I'd like to stick with 4532 as much as possible here.
So each server should determine if somebody else is already broadcasting on an address and or auto-select the next address.  So we start at  224.0.0.1 and just 1-up the address automatigically perhaps.
Or can we bind to the multicast to only allow one broadcaster? 

    On Thursday, November 2, 2023 at 12:10:16 PM CDT, Mikael Nousiainen ***@***.***> wrote:  
 
 


@mdblack98 Ok, regarding port numbers, maybe we could state what you/we are after here:
   
   -    
Is the goal to allow clients to send UDP commands to a constant, known multicast address (e.g. 224.0.0.2:4531 here) so that there is no need to configure a unicast address of rigctl for rig control? This can be only done with multicast. Would be useful?

   -    
If we use multicast for receiving and there are multiple rigs open, all of them publish state updates to 224.0.0.1:4532, so if the multicast receiver listens to the same multicast group/port, it will receive the multicast publisher state updates, which is not desirable. This is what I tried to avoid with the separate port number. Also, I think even the same instance of Hamlib might actually end up in a "loopback" when publishing multicast packets if it's listening to the same group. Or do you know a way to avoid this?

   -    
If the command receiver listens to a unicast address only, there should be no issue with "loopback", but clients need to be configured to know the Hamlib unicast IP/port just like with rigctld TCP connections... So just wondering if that is something you would like to avoid as well?

mikaelnousiainen:
@mdblack98 Ok. But then I'm wondering how the clients know which multicast IP to listen to -- especially if the Hamlib instances for different rigs get the IP automatically assigned (e.g. +1 for each rig like you mentioned)?

I'm thinking if we could have as little configuration as possible:

If all Hamlib rigs would publish multicast packets (by default) to the same group 224.0.0.1:4532 so any client could just listen to this address/port to get an idea of ALL the rigs in the local network (also outside the local computer).

Imagine having both local and remote Hamlib instances running in the network. A logbook app could be running on a different PC, right?

Each of the packets would have the unique Hamlib/rig ID, so the clients could then display a list of detected rigs just by listing the unique IDs.

What do you think?

Then, again to have no configuration, all clients would have a single multicast address (e.g. 224.0.0.2:4531) to send commands to -- and the Hamlib instance could be running on a different PC again. The command would include the Hamlib/rig ID from the previously received packet to target a single Hamlib/rig instance.

One IP/port pair for messages published by Hamlib and another IP/port pair for commands.

The only reason I'm suggesting to use different pairs is to avoid mixing state packets and command packets. Of course they content of the packets could indicate whether the data is a command or state/data from rig.

Thoughts?

Can you list some practical use cases you have in mind for the UDP commands, so I get a clearer idea of what you're after?

mdblack98:
#1 Log4OM -- rigctld would hook up to a multicast server and allow Log4OM to use query VFOA/B using get_rig_info as it does now along with vfo_list.  Also able to set freq/mode.  With WSJT-X running no additional configuration should be needed with default 224.0.0.1/4532#2 W1AW -- I can imagine W1AW monitoring all rigs in the building.  Imagine a screen with all the rigs showing PTT status and such.#3 Flex -- people with up to 8 slices on Flex not needing to set up CAT control to WSJT-X.#4 FLRig -- I plan on adding same multicast to FLRig too.  In that case WSJT_X would connect  to FLRig's multicast ability#5 N1MM (and other non-hamlib programs) could use rigctlcom to connect to multicast and not need any virtual serial ports.#6 HRD -- should be able to finally get HRD working with rigctlcom using raw write/reads so HRD would then work with FLRig or WSJT-X transparently.
I'm a bit reluctant to use different ports as there are only 2^16 ports and I don't want to start using too many and risk colliding with other things.
My concern was the UDP return packet that only the appropriate  server retrieved it since it wouldn't be multicast....but maybe  we need that too instead of using different IP/Ports.I was thinking about just using the rigctld commands but could wrap that in a JSON packet with the Id: 

    On Thursday, November 2, 2023 at 01:04:53 PM CDT, Mikael Nousiainen ***@***.***> wrote:  
 
 


@mdblack98 Ok. But then I'm wondering how the clients know which multicast IP to listen to -- especially if the Hamlib instances for different rigs get the IP automatically assigned (e.g. +1 for each rig like you mentioned)?

I'm thinking if we could have as little configuration as possible:

If all Hamlib rigs would publish multicast packets (by default) to the same group 224.0.0.1:4532 so any client could just listen to this address/port to get an idea of ALL the rigs in the local network (also outside the local computer).

Imagine having both local and remote Hamlib instances running in the network. A logbook app could be running on a different PC, right?

Each of the packets would have the unique Hamlib/rig ID, so the clients could then display a list of detected rigs just by listing the unique IDs.

What do you think?

Then, again to have no configuration, all clients would have a single multicast address (e.g. 224.0.0.2:4531) to send commands to -- and the Hamlib instance could be running on a different PC again. The command would include the Hamlib/rig ID from the previously received packet to target a single Hamlib/rig instance.

One IP/port pair for messages published by Hamlib and another IP/port pair for commands.

The only reason I'm suggesting to use different pairs is to avoid mixing state packets and command packets. Of course they content of the packets could indicate whether the data is a command or state/data from rig.

Thoughts?

Can you list some practical use cases you have in mind for the UDP commands, so I get a clearer idea of what you're after?

mdblack98:
Couple of things....
#1 On windows just running "rigctl" and quitting never quitsRig command: q 2:rig.c(1549):rig_close entered  3:rig.c(7985):morse_data_handler_stop entered  3:rig.c(8028):morse_data_handler_stop returning(0)  3:rig.c(7944):async_data_handler_stop entered  3:rig.c(7977):async_data_handler_stop returning(0)  3:event.c(413):rig_poll_routine_stop entered../../src/src/event.c(341): Stopping rig poll routine thread  3:event.c(446):rig_poll_routine_stop returning(0)  3:network.c(1359):network_multicast_receiver_stop entered
#2 I still like the idea of 1 packet per second plus faster packets when things change.  Otherwise clients have to know when to request.  Imagine running a dozen clients and you have to restart them.  The would all have to request a status packet or could just pick up the next packet....the 1-per-second is more like a heartbeat too so you know the rig is alive.
#3 During shutdown should send a new packet showing status of "Rig not available" or "Rig powered down" depending on the true shutdown status.  But once could CTRL-C rigctl/rigctld or kill it and no packet would be sent...ergo the heartbeat packet is important.
As  long as we use 224.0.0.2 from the client why do we need another port?  Everybody should be filtering on the packets they want.I haven't given this much thought but trying to minimize # of ports...not sure I see the advantage of separate IP address. 

    On Thursday, November 2, 2023 at 11:01:57 AM CDT, Mikael Nousiainen ***@***.***> wrote:  
 
 


@mdblack98 Please test my latest changes here: https://github.com/mikaelnousiainen/Hamlib/tree/hamlib-multicast-1 (branch hamlib-multicast-1)

It will publish state/data updates to 224.0.0.1:4532 and receive+log packet data (with verbose log level) from 224.0.0.2:4531.

mikaelnousiainen:
@mdblack98 Responding to your comments from the last 2 messages:

> I'm a bit reluctant to use different ports as there are only 2^16 ports and I don't want to start using too many and risk colliding with other things.

Yeah, totally get that. See my comments below about this.

> My concern was the UDP return packet that only the appropriate  server retrieved it since it wouldn't be multicast....but maybe  we need that too instead of using different IP/Ports.I was thinking about just using the rigctld commands but could wrap that in a JSON packet with the Id:

Yeah, I think the "no configuration" approach (by using multicast UDP for commands) would be a great addition. That way things would "just work" from the end-user point of view. It does require some filtering at Hamlib to check which command is actually meant for which instance, but I think it is a reasonable trade-off for simplicity! Hey, this could potentially make the life of Hamlib maintainers easier! :D

> On windows just running "rigctl" and quitting never quitsRig command: q 

> I still like the idea of 1 packet per second plus faster packets when things change.  Otherwise clients have to know when to request.  Imagine running a dozen clients and you have to restart them.  The would all have to request a status packet or could just pick up the next packet....the 1-per-second is more like a heartbeat too so you know the rig is alive.

Yeah, good point, I will modify the code so that it checks cache more often, but sends packets at least at the poll_interval frequency (defaults to 1 second).

> During shutdown should send a new packet showing status of "Rig not available" or "Rig powered down" depending on the true shutdown status.  But once could CTRL-C rigctl/rigctld or kill it and no packet would be sent...ergo the heartbeat packet is important.

Again, good idea. Let's see how to implement this. Maybe we could have power status included in the state packets as well?

> As  long as we use 224.0.0.2 from the client why do we need another port?  Everybody should be filtering on the packets they want.I haven't given this much thought but trying to minimize # of ports...not sure I see the advantage of separate IP address.

There are two things here:

* The issue with needing more than 1 por is just Windows having to use `INADDR_ANY` -- it binds to all IPs so you will get traffic to port 4532 from ALL IPs! That said, if we used `recvmsg()` we could filter based on the destination IP, I guess? Will have to check!

* The idea behind using two multicast IP addresses is just to separate traffic between published rig state and the commands. 224.0.0.1 would contain traffic going from Hamlib to clients and 224.0.0.2 would contain traffic from clients to Hamlib. That way we would avoid mixed traffic -- I'm assuming many clients would NOT be happy to get commands from other clients (even if Hamlib handles/ignores state updates coming from other rigs/Hamlib instance). Of course parsing in Hamlib still needs to ignore any invalid command. 

* If we use a single multicast IP/port pair, all clients will get commands from other clients (when trying to receive Hamlib state updates) and Hamlib will get state update packets from other Hamlib instances (when trying to receive client commands).

What if I try to make this work with just one port (4532 -- I think this should be doable in Windows too), but still use the two IPs for different types of traffic?

mikaelnousiainen:
@mdblack98 Please test the latest changes. I've made the following changes:

1) Use the same port 4532 for both state snapshots and commands. This does not cause conflicts, as the multicast groups are still different (and Hamlib only subscribes to one of them). The IPs are still 224.0.0.1 for state snapshots and 224.0.0.2 for commands. Seems to work both on Windows and Linux. However, on Windows, the socket DOES pick up UDP packets sent to localhost (127.0.0.1:4532) and unicast any network interface IP address. I guess that should not be an issue if Hamlib state snapshots are ALWAYS published to 224.0.0.1:4532. Those won't end up there!

2) I've fixed shutdown/stopping of the multicast receiver. It was necessary to start using non-blocking sockets and `select()` for this to work on Windows too. The `shutdown()` command worked only on Linux -- I left it there still on purpose to tear down the socket completely. Doesn't cause any harm.

3) The poll routine now publishes state snapshots at a maximum of 50ms interval and also waits up to the defined `poll_interval`  (defaulting to 1 second) to publish state snapshots even if nothing changes.

Haven't tested this much on real rigs yet. And the "shutdown" state message is still missing.

mdblack98:
Good progress -- your patch does show some conflicts that need to be resolved.
Would like to see a shutdown message sent changing the "status".
I can think of several status possibilities.
Common:"OK" means all is well"Power off" means rig power has been turned off but Hamlib is till running and connected"Disconnected" means Hamlib has been shut down (politely)Not  so common:"Error" means an error has occurred"Warning" means a warning has occurred"Aborted" Hamlib was aborted with kill signal or CTRL-C
Any suggestions here?

Have you gotten this to work yet?  Still lots more work here...
https://www.dropbox.com/scl/fi/9c00adeqk1o549lfz3cnh/HamlibGUI20231102.zip?rlkey=oufjes2ip2jqmwasteyl5o82l&dl=0

Mike

 

    On Friday, November 3, 2023 at 10:24:40 AM CDT, Mikael Nousiainen ***@***.***> wrote:  
 
 


@mdblack98 Please test the latest changes. I've made the following changes:
   
   -    
Use the same port 4532 for both state snapshots and commands. This does not cause conflicts, as the multicast groups are still different (and Hamlib only subscribes to one of them). The IPs are still 224.0.0.1 for state snapshots and 224.0.0.2 for commands. Seems to work both on Windows and Linux. However, on Windows, the socket DOES pick up UDP packets sent to localhost (127.0.0.1:4532) and unicast any network interface IP address. I guess that should not be an issue if Hamlib state snapshots are ALWAYS published to 224.0.0.1:4532. Those won't end up there!

   -    
I've fixed shutdown/stopping of the multicast receiver. It was necessary to start using non-blocking sockets and select() for this to work on Windows too. The shutdown() command worked only on Linux -- I left it there still on purpose to tear down the socket completely. Doesn't cause any harm.

   -    
The poll routine now publishes state snapshots at a maximum of 50ms interval and also waits up to the defined poll_interval (defaulting to 1 second) to publish state snapshots even if nothing changes.


Haven't tested this much on real rigs yet. And the "shutdown" state message is still missing.

mikaelnousiainen:
I'll fix the conflicts and add the status messages (at least the ones that are easy) tomorrow. Any other follow-up work could be done in new PRs to avoid keeping this one open for too long.

mikaelnousiainen:
@mdblack98 Just noticed you merged this :) It's ok, but I was still working on the status messages. I will file another PR on those.

Issue #1412, opened 2023-10-31
700ms  delay needed for ptt on transition


mdblack98:
My 10GHz set needs 700 ms to go from receive to transmit due to the waveguide relay. So I think the audio must be delayed at least 700 ms with respect to the PTT-signal.
I could not set the TX delay to a value higher than 0.5 s.
And I am worried that the delay is too long, so that the transmission is too late

GeoBaltz:
This commit causes me to lose PTT control from wsjt-x (on my TS-890S). Verified with git bisect.

mdblack98:
Try the latest and if that doesn't work please provide  debug.
Please place this file as described belowhttps://www.dropbox.com/s/t52ngcalsgnpm8m/wsjtx_log_config.ini?dl=0
 C:\Users\[username]\AppData\Local\WSJT-X The WSJT-X_Rigcontrol.log file will be in the same location Here's a youtube video showing Windows users howto -- https://www.youtube.com/watch?v=q9tAvhNmSvc
For Linux put it in   ~/.config The WSJT-X_Rigcontrol.log file will be here: ~/.local/share/WSJT-X
For MacOS put it in /Users/[username]/Library/Preferences Restart WSJT-X and duplicate the problem.Shut down WSJT-X
Then send me the WSJT-X_RigControl.log fileMike W9MDB

 

    On Wednesday, November 1, 2023 at 03:42:53 PM CDT, GeoBaltz ***@***.***> wrote:  
 
 


This commit causes me to lose PTT control from wsjt-x (on my TS-890S). Verified with git bisect.

GeoBaltz:
Still broken with current git 
[WSJT-X_RigControl.log](https://github.com/Hamlib/Hamlib/files/13309778/WSJT-X_RigControl.log)

Still looks like this is caused by the additional item added to hamlib_port_t by commit 6cb17e49dc "Add post_ptt_delay for situations..."

mdblack98:
Please try the latest.....

Should be fixed  now.








On Thursday, November 9, 2023 at 09:25:11 AM CST, GeoBaltz ***@***.***> wrote: 







Still broken with current git
WSJT-X_RigControl.log

Still looks like this is caused by the additional item added to hamlib_port_t by commit 6cb17e4 "Add post_ptt_delay for situations..."

GeoBaltz:
That was with the latest, as of 2023-11-09 10:20EST

GeoBaltz:
[...]

PR #1444, opened 2023-12-04
Restore shared library ABI compatibility with previous hamlibs

GeoBaltz:
No need for post_ptt_delay to be in hamlib_port - move it to misc
  rig_state data.
Update references to post_ptt_delay
Fix a couple of typos
Add code to actually set post_ptt_delay

Fixes issue #1412

### DIFF
--- a/include/hamlib/rig.h
+++ b/include/hamlib/rig.h
@@ -2388,7 +2388,7 @@ typedef struct hamlib_port {
     int fd_sync_error_read;     /*!< file descriptor for reading synchronous data error codes */
 #endif
     short timeout_retry;    /*!< number of retries to make in case of read timeout errors, some serial interfaces may require this, 0 to disable */
-    int post_ptt_delay;         /*!< delay after PTT to allow for relays and such */
+// DO NOT ADD ANYTHING HERE UNTIL 5.0!!
 } hamlib_port_t;
 
  
@@ -2787,6 +2787,8 @@ struct rig_state {
     rig_comm_status_t comm_status; /*!< Detailed rig control status */
     char device_id[HAMLIB_RIGNAMSIZ];
     int dual_watch; /*!< Boolean DUAL_WATCH status */
+    int post_ptt_delay;         /*!< delay after PTT to allow for relays and such */
+// New rig_state items go before this line ============================================
 };
 
 /**
@@ -2799,7 +2801,7 @@ struct rig_state {
  * It is NOT fine to touch this struct AT ALL!!!
  */
 struct rig_state_deprecated {
-    /********* ENSURE YOU DO NOT EVERY MODIFY THIS STRUCTURE *********/
+    /********* ENSURE YOU DO NOT EVER MODIFY THIS STRUCTURE *********/
     /********* It will remain forever to provide DLL backwards compatiblity ******/
     /*
      * overridable fields
--- a/src/conf.c
+++ b/src/conf.c
@@ -268,7 +268,15 @@ static int frontend_set_conf(RIG *rig, token_t token, const char *val)
         }
 
         rs->rigport.post_write_delay = val_i;
-        rs->rigport_deprecated.timeout = val_i;
+        rs->rigport_deprecated.post_write_delay = val_i;
+        break;
+
+    case TOK_POST_PTT_DELAY:
+        if (1 != sscanf(val, "%ld", &val_i))
+        {
+	  return -RIG_EINVAL;
+        }
+        rs->post_ptt_delay = val_i;
         break;
 
     case TOK_TIMEOUT:
@@ -841,7 +849,7 @@ static int frontend_get_conf2(RIG *rig, token_t token, char *val, int val_len)
         break;
 
     case TOK_POST_PTT_DELAY:
-        SNPRINTF(val, val_len, "%d", rs->rigport.post_ptt_delay);
+        SNPRINTF(val, val_len, "%d", rs->post_ptt_delay);
         break;
 
     case TOK_TIMEOUT:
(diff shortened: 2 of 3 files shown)


mdblack98:
I need to understand why this was broken before applying this.

That change should NOT have broken backwards compatibility.

If it did that means all sorts of future problems are possible and the fix I put in to allow backwards compatibility isn't working and I may need to improve that.

Were you using the ptt_delay?  That may be the problem.  Setting the value in the old hamlib might have corrupted the ptt call.






On Monday, December 4, 2023 at 09:58:00 AM CST, GeoBaltz ***@***.***> wrote: 





No need for post_ptt_delay to be in hamlib_port - move it to misc
rig_state data.
Update references to post_ptt_delay
Fix a couple of typos
Add code to actually set post_ptt_delay

Fixes issue #1412

  https://github.com/Hamlib/Hamlib/pull/1444

Commit Summary

     
    * f4d40e0 Restore shared library ABI compatibility with previous hamlibs
    
File Changes  (3 files) 
     
    *  M include/hamlib/rig.h (6) 
    
    *  M src/conf.c (12) 
    
    *  M src/rig.c (2) 
    

Patch Links:

     
    * https://github.com/Hamlib/Hamlib/pull/1444.patch
    
    * https://github.com/Hamlib/Hamlib/pull/1444.diff

GeoBaltz:
The problem isn't with rig_cache (which you fixed), but with the addition of post_ptt_delay into the hamlib_port structure.  It changes the size of that structure, so everything that follows it inside the rig_state structure is moved.

I'm running on a distro that uses shared libraries for everything.  Since I'm working/testing on hamlib4.6-git, the libhamlib.so I built is loaded when rigctld(or cqrlog) is started. When wsjtx is started, it also is linked to the loaded hamlib.so.  But I'm using the system provided wsjtx, which was built against the released hamlib 4.5.5.  Since wsjtx looks into the rig structure, it has different offsets compiled into it, and gets wrong data.  It is expecting the structures to be the same (i.e., ABI compatibility), as has been the case throughout hamlib 4.0 to 4.5.5.

It only affects shared (dynamicly linked) library usage.

To explain from a different perspective: this is exactly what happens when you ignore the comments around line 2320 of rig.h, right above 'typedef struct hamlib_port', that say 'DO NOT CHANGE THIS STRUCTURE ALL UNTIL 5.0'.

The commit I sent only moves the data to where it won't affect the offset of other structures, and actually fixes the post_ptt_delay, as there was no code that ever stored into it.

mdblack98:
That's not true -- you'll find a deprecated structure which does not change.   That comment needs to be removed as it was applicable to the old structure...still applies to the deprecated structure.

But...there may be a problem with accessing new elements in the new library with old shared code that doesn't have them.
That's what I need to check.  I may need to disable new items when the old shared library is being used.

You didn't answer my question -- were you using post_ptt_delay?

I need to try and duplicate this problem so I can do a permanent fix.  I should be able to add to hamlib_port_t without causing this problem.






On Monday, December 4, 2023 at 12:12:37 PM CST, GeoBaltz ***@***.***> wrote: 







The problem isn't with rig_cache (which you fixed), but with the addition of post_ptt_delay into the hamlib_port structure. It changes the size of that structure, so everything that follows it inside the rig_state structure is moved.

I'm running on a distro that uses shared libraries for everything. Since I'm working/testing on hamlib4.6-git, the libhamlib.so I built is loaded when rigctld(or cqrlog) is started. When wsjtx is started, it also is linked to the loaded hamlib.so. But I'm using the system provided wsjtx, which was built against the released hamlib 4.5.5. Since wsjtx looks into the rig structure, it has different offsets compiled into it, and gets wrong data. It is expecting the structures to be the same (i.e., ABI compatibility), as has been the case throughout hamlib 4.0 to 4.5.5.

It only affects shared (dynamicly linked) library usage.

To explain from a different perspective: this is exactly what happens when you ignore the comments around line 2320 of rig.h, right above 'typedef struct hamlib_port', that say 'DO NOT CHANGE THIS STRUCTURE ALL UNTIL 5.0'.

The commit I sent only moves the data to where it won't affect the offset of other structures, and actually fixes the post_ptt_delay, as there was no code that ever stored into it.

GeoBaltz:
Nobody could use post_ptt_delay - the code to set it was missing.   I certainly wasn't using it.

FWIW - if I build/install from current git master, wsjtx fails to activate PTT.  With that commit applied(i.e., with the hamlib_port structure reverted to 4.5.5 size), it works.

mdblack98:
FYI....this was fixed in WSJT-X...just after the release of 2.6.1

If you apply the HamlibTransceiver.cpp from 2.6.2 or later it should work as-is.
Meanwhile I'll take your patch but I need to come up with a better solution which means deprecating these structure one more time changing them to a pointer,  and changing all references to function calls.

commit f972fc18e14cd4ce1b46ac66956108fc7db263a0
Author: Bill Somerville ***@***.***>
Date:   Tue Feb 2 22:45:45 2021 +0000


    Remove direct struct access from usage of the Hamlib API
    
    Preparation for safe dynamic linking to Hamlib where minor Hamlib
    upgrades can be deployed just by replacing the DLL/SO.










On Monday, December 4, 2023 at 02:09:45 PM CST, GeoBaltz ***@***.***> wrote: 







I'm not talking about the deprecated structure at the top of rig->state. I'm looking further down (around line 2712) where there are 3 hamlib_port_t declarations: rigport, pttport and dcdport. Your change makes rigport 4 or 8 (probably 8) bytes bigger (to align on 64bit boundary), so pttport get moved 8 bytes further away from the struct address. And wsjtx 2.6.1 looks at rig->state.pttport.type.ptt in several places, finding wrong data.

And one final argument that the size of the hamlib_port struct is the real culprit. If I change the declaration of post_ptt_delay from int to short, so it uses the leftover bytes from the preceding short and doesn't change the struct length from 4.5.5, wsjtx works.

I rest my case.

Issue #1445, opened 2023-12-04
rigstate hamlib_port_t deprecation


mdblack98:
Need to deprecate hamlib_port_t entries,  add pointer references to them, and change all internal references in all Hamlib code to the pointers.
This goal is to remove any dependency on rig_caps or rig_structure alignment by only accessing via function calls which will be internal to the shared libary instead of depending on rig.h
This will maintain backwards compatibility from 4.7 on and also allow older apps to update their access methods too for maximum compatibility if applicable.
WJST-X already did this change in Feb 2023.

GeoBaltz:
I'd like to get the ball rolling on this one, as it's going to take a long time....

What I'd like to do:

1. Define macros to access all the hamlib_port structures, replicating the current methods.  Make the macro calls generic enough so they cover both the current and final internal and external usages.
2. Port a few files to use the new macros with the current structures, to show proof of concept.
3. Start the long process of porting the entire code base.  Should be possible without breaking ABI compatibility with current S/W, and keeping the ongoing conversion transparent to anyone using current git or snapshots.
4. Make up external definitions for the macros to call functions within the library to retrieve pointers to the structures, as needed.  Write the functions, possibly with some sort of access control; should be flexible enough for other data access, if needed.
5. Release it.  Big warning messages - This is the last ABI compatible version, next version will require use of new methods access to port data (and anything else we throw in.)
6. Be ready to help 3rd parties port to new interface.
7. Cut over.  Redefine macros.  Rearrange structures. calloc() ports. See if anything else can be dropped. Clean up.

Timing:  I should have a PR of 1 & 2 ready for discussion in a day or so.  With the holidays and 4.6 coming up, I don't expect much for a while.  Then maybe 3-4 months for 3 & 4.

By a quick grep -R | wc -l, there are over 1500 references to the 3 port tables.  Gonna take a some time.

mdblack98:
Your idea has merit but we don't need to break the ABI.  

#1 The old structure stays around but gets deprecated.
#2 During open the new structure gets copied to the old one (just the sizeof for the deprecated entry -- that way old accesses work...new items aren't needed for old code.
#3 The macros to access make sense -- anything that updates it must update the deprecated one too if applicable.







On Monday, December 18, 2023 at 06:22:34 PM CST, GeoBaltz ***@***.***> wrote: 







I'd like to get the ball rolling on this one, as it's going to take a long time....

What I'd like to do:

     
[...]

PR #1455, opened 2023-12-20
Basic definitions and first steps for making port tables addressed by pointers

GeoBaltz:
Proposed implementation for pointerizing addresses of port structures.

Define macros that return pointers to the current tables.  After converting all references to use the macros instead of hard coding, the macro definitions can be changed as needed to address dynamically created and expandable structures.

Phase 1 of issue #1445

### DIFF
--- a/include/hamlib/rig.h
+++ b/include/hamlib/rig.h
@@ -2464,6 +2464,21 @@ typedef hamlib_port_t_deprecated port_t_deprecated;
 typedef hamlib_port_t port_t;
 #endif
 
+/* Macros to access port structures/pointers
+ * Make it easier to change location in preparation
+ *   for moving them out of rig->state.
+ * See https://github.com/Hamlib/Hamlib/issues/1445
+ */
+// Note: Experimental, and subject to change!!
+#define RIGPORT(r) (&r->state.rigport)
+#define PTTPORT(r) (&r->state.pttport)
+#define DCDPORT(r) (&r->state.dcdport)
+/* Then when the rigport address is stored as a pointer somewhere else(say,
+ *  in the rig structure itself), the definition could be changed to
+ *  #define RIGPORT(r) r->somewhereelse
+ *  and every reference is updated.
+ */
+
 #define HAMLIB_ELAPSED_GET 0
 #define HAMLIB_ELAPSED_SET 1
 #define HAMLIB_ELAPSED_INVALIDATE 2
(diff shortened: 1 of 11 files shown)


include/hamlib/rig.h
```diff
@@ -2464,6 +2464,21 @@ typedef hamlib_port_t_deprecated port_t_deprecated;
 typedef hamlib_port_t port_t;
 #endif
 
+/* Macros to access port structures/pointers
+ * Make it easier to change location in preparation
+ *   for moving them out of rig->state.
+ * See https://github.com/Hamlib/Hamlib/issues/1445
+ */
+// Note: Experimental, and subject to change!!
+#define RIGPORT(r) (&r->state.rigport)
```
mikaelnousiainen, line 2473: This is a good start for fully transitioning to pointers in 5.0.

I'd try to prefix all definitions, both to scope them by device type and to be sure no name will easily conflict with any other code. Since these are specific rigs (radios), I'd just add a RIG_ prefix to all of them.

Have you checked rotator and amp structs BTW?

mdblack98:
I think we want to preface the macros with HAMLIB.
HAMLIB_RIGPORT
That should prevent any conflicts.
And I'm not going to merge this until 4.7 -- 4.6 is due to be release in the next two days -- one more bug to fix.
Mike

 

    On Wednesday, December 20, 2023 at 12:30:12 AM CST, Mikael Nousiainen ***@***.***> wrote:  
 
 


@mikaelnousiainen commented on this pull request.

In include/hamlib/rig.h:
> @@ -2464,6 +2464,21 @@ typedef hamlib_port_t_deprecated port_t_deprecated;
 typedef hamlib_port_t port_t;
 #endif
 
+/* Macros to access port structures/pointers
+ * Make it easier to change location in preparation
+ *   for moving them out of rig->state.
+ * See https://github.com/Hamlib/Hamlib/issues/1445
+ */
+// Note: Experimental, and subject to change!!
+#define RIGPORT(r) (&r->state.rigport)

This is a good start for fully transitioning to pointers in 5.0.

I'd try to prefix all definitions, both to scope them by device type and to be sure no name will easily conflict with any other code. Since these are specific rigs (radios), I'd just add a RIG_ prefix to all of them.

Have you checked rotator and amp structs BTW?

GeoBaltz:
Since

1. Adding qualifiers to the names is redundant within the hamlib source tree.
2. The external definition will have to be different - allowing these outside would defeat the purpose of the whole exercise.
3. I'm lazy, and averse to typing any more than I have to.

I'd propose to just add the HAMLIB_ prefix to the external interfaces.  See the commit I added to the branch.

PR #1470, opened 2024-01-04
Fix copy/paste errors in rigs/icom/

GeoBaltz:
I have no concrete proof that these are broken, and no way to test them, but these suuuuure don't look right.

Feel free to laugh at me if I'm wrong.


mdblack98:
No laughing here...good catch...

PR #1544, opened 2024-04-28
Sync local build scripts with main repository

N0NB:
Updates for including libgcc* provided by MinGW in Debian 11 and Debian 12 (untested).

Include other updates added to the build virtual machine along the way.


N0NB:
Mike, look over these changes.  These are the current Windows build files from my daily snapshot VM.  See if I've missed anything.

Issue #1420, opened 2023-11-08
rig_cache structure change to pointer


mdblack98:
We need to deprecate the rig_cache struct reference in rig_state and make a  new pointer reference  like we did for rig_state.

GeoBaltz:
I have gotten back to working on this, and even have a dynamically allocated cache version that works with WSJT-X.  However, I made a few changes while deprecating the definitions in rig.h that may need to be discussed.

The definition in rig.h was changed from 'struct rig_cache' to 'struct rig_cache_deprecated', and the references in rig_state changed likewise.   However, the new definition went into src/cache.h, not include/hamlib/rig.h - out of scope for any application compiled with Hamlib.

I did this for several reasons -
- My general dislike for having active internal data shared between library and application - how the cached operations are really done is not germane to the app, only to Hamlib.
- Current programming practice - let the object manage its own data, without outside meddling.
- Having another albatross around the necks of current and future Hamlib developers.  See how much code was needed to handle previous deprecations, and even then compatibility is only in one direction.
- And the big technical reason - if Hamlib is ever going to handle rigs sending state changes automatically (Auto Info in Kenwood-speak), then the whole idea of cache needs to change.  STATE(rig) should always be up to date, commands sent to the rig are confirmed automatically, inquiries should have the requested data already available; cache may not even be involved.

Since I don't have a feel for what applications are doing with current cache data, I'm really looking for feedback on how to proceed.  Do I need to provide some sort of bridge to/from the old cache structure; are changes to the API necessary; or do I start over?  Should I start a discussion on the ML?

N0NB:
Yes, please start a ML discussion as these updates aren't forwarded to the list so far as I know and unless one is "subscribed" to an issue or actively looks for an issue, this update won't bee seen.

GeoBaltz:
On Thursday, May 8, 2025 7:33:20 AM Eastern Daylight Time Nate Bargmann wrote:
> N0NB left a comment (Hamlib/Hamlib#1420)
> 
> Yes, please start a ML discussion as these updates aren't forwarded to the
> list so far as I know and unless one is "subscribed" to an issue or
> actively looks for an issue, this update won't bee seen.
Will do.  As I got further into writing the comment on the issue, I started 
thinking it might deserve a wider audience.

N0NB:
Closing per request of George.

Issue #536, opened 2021-02-08
Make rig_cache dynamic allocation


mdblack98:
Along with other structures make rig_cache dynamically allocated.
This will help shared library problems.
Right now any modification to rig_cache can change the rig_caps data structure offsets....and we want to reduce this likelihood.

Issue #487, opened 2021-01-03
rig structure references change to ptr


mdblack98:
In 5.0 will be changing all internal structure references in rig_caps and such to use pointers.
Problem with shared libaries was created when hamlib_port_t had a variable added (i.e. flushx) which changed offsets of everything past hamlib_port_t in rig_caps.
rig_caps will be changes to make any such items pointers instead of static structures.

N0NB:
@GeoBaltz I think you're working toward completing this in 5.0.

GeoBaltz:
Yes, Nate, I am.  Even with all the prep work done over the last year and a half I'm still not sure what I'll run into at the end.  I do intend to have the application side ready for 4.7, so the user/app side source should work on either 4.7 or 5.0.

Aside from the obvious benefit of not having structures within structures, I'd like to make rig.h enough to call the API, with as little as possible of the internals of Hamlib in it.  This sort of future-proofs the dll problems of the past.

It might be good to add a reference to this issue to the 5.0 feature list.

N0NB:
Good idea.  I will do so.

PR #1573, opened 2024-07-02
More conversion to pointers

GeoBaltz:
Convert src/*.c to use pointers for state.

Fix up some recent additions.

Part of issues #487, #536, #1420


GeoBaltz:
Fixed?  I hope github is satisfied

PR #1590, opened 2024-07-29
FT-817 power/SWR levels

chrbayer84:


mdblack98:
When you can I'd like to know the serial numbers for the rigs you test

Issue #1634, opened 2024-12-01
kenwood_send_morse adds an unwanted space at the end of most transmissions


dl1jbe:
First reported by OK1RR on TLF mailing list for his Elecraft K4. Confirmed by myself on Kenwood TS-590.

The following commands in rigctl can show the problem:

```
b test<cr>
b text<cr>
```

is always sent as 

```
'test text'
```

Problem seems to be the left aligned positioning of the text in the 24 characters space followed by trailing spaces.

Doing some experiments showed that any number of trailing spaces gets compressed to exactly one (additional) space. As soon as the last character in morsebuffer is no space character there will be no additional space sent. Any number of spaces before the text gets ignored.

mdblack98:
And what do you expect it to do?

Does look like the Elecraft KY doesn't need the trailing spaces like the Kenwood's do (they need all 24 bytes) but the behavior on the Kenwood is not something that Hamlib can change.
Might be that the K4 would do "KY test" "KY text" as "testtext" if we get rid of the trailing spaces...that would make the behavior different than the Kenwood.

I don't think removing the space from between two invocations of KY would be the right thing to do so I need some justification to make any change.

Mike W9MDB








On Sunday, December 1, 2024 at 12:38:44 PM CST, Thomas Beierlein ***@***.***> wrote: 







First reported by OK1RR on TLF mailing list for his Elecraft K4. Confirmed by myself on Kenwood TS-590.

The following commands in rigctl can show the problem:
b test<cr>
b text<cr>

is always sent as
'test text'

Problem seems to be the left aligned positioning of the text in the 24 characters space followed by trailing spaces.

Doing some experiments showed that any number of trailing spaces gets compressed to exactly one (additional) space. As soon as the last character in morsebuffer is no space character there will be no additional space sent. Any number of spaces before the text gets ignored.

dl1jbe:
If I read the code correctly at the moment the K4 gets the same handling as the normal Kenwood rigs. Other variants (K3, ...) have a special handling. Maybe the K4 should go there too. That needs to be tested by someone owning one of these.

For the Kenwood rigs it seems we just need to right align the text in the 24 characters buffer - works here at least for the TS-590.

dl1jbe:
To add something for justification: Think about a CW keyboard sending the characters via hamlib. As we do not know the timing of the writer we would have to sent each character to hamlib as soon as it came in. But in the current state there would be a space between any character you type.

GeoBaltz:
[...]

PR #1636, opened 2024-12-02
Change Kenwood _send_morse() for TS-890/990

GeoBaltz:
Use variable length messages for CW keying - drops extra space in output, shortens message by a bunch.

Doesn't help TS-480/590, sorry.

Issue #1634


mdblack98:
Do not use "for (int i"
That is not portable.

PR #1639, opened 2024-12-21
Added multi-stage Docker build and hamlib-runtime image

8r4n:
I added a multi-stage build to simplify and speed up the build process. The docker-build README provides the user with instructions on how to execute the hamlib-runtime container after the build is complete.


8r4n:
Thanks for accepting my contribution.

PR #1640, opened 2024-12-22
Fix unbalanced ENTERFUNC/RETURNFUNC pairs in icom.c

GeoBaltz:
Also tweak comments/formats to eliminate false positives in code base
   (except for the Catch-22 ones in tests/func_chk.c itself)


mdblack98:
I take it you have some utility you wrote that checks this.
Would be nice to integrate it so it would automatically check as part of "make check" or the test programs.

PR #1642, opened 2024-12-25
Fix kendwood_priv_data definition

bradfordboyle:
When compiling Hamlib 4.6 on Alpine Linux, the build fails with the following error message:

    kenwood.h:186:5: error: unknown type name 'mode_t'; did you mean 'rmode_t'?

According to the man page for mode_t (on Debian), the standard C library provides this type and it is an integer type. The struct member last_mode_pc is compared with curr_mode which has type rmode_t so this commit updates the definition to use this type.

With this change, Hamlib 4.6 builds successfully on Alpine Linux 3.21.0.

### DIFF
--- a/rigs/kenwood/kenwood.h
+++ b/rigs/kenwood/kenwood.h
@@ -183,7 +183,7 @@ struct kenwood_priv_data
     int save_k2_ext_lvl; // so we can restore to original
     int save_k3_ext_lvl; // so we can restore to original -- for future use if needed
     int voice_bank; /* last voice bank send for use by stop_voice_mem */
-    mode_t last_mode_pc; // last mode memory for PC command
+    rmode_t last_mode_pc; // last mode memory for PC command
     int power_now,power_min,power_max;
 };
 


mdblack98:
Hmm...trying to figure out what that even builds here on 3 different machines.  There is no mode_t defined anywhere.

bradfordboyle:
@mdblack98 On my Debian machine, it looks like `mode_t` is defined in `sys/types.h` so my best guess is that it gets pulled in via that (directly or indirectly). It happens to be defined as uint64_t so that's probably why it builds.

mdblack98:
And that would explain why it originally got named rmode_t years ago.
Thanks







On Tuesday, December 24, 2024 at 10:57:03 PM CST, Bradford Boyle ***@***.***> wrote: 







@mdblack98 On my Debian machine, it looks like mode_t is defined in sys/types.h so my best guess is that it gets pulled in via that (directly or indirectly). It happens to be defined as uint64_t so that's probably why it builds.

PR #1649, opened 2025-01-05
Convert text encoding from ISO-8859-1 to UTF-8

dforsi:
This PR converts text in comments from ISO-8859-1 to UTF-8 for consistency with other files.
In the diff below, for some reason, github shows the same glyph before/after for ± µ ä but the actual difference is visible when using `git diff` in the terminal.

### DIFF
--- a/rigs/aor/README.ar5000
+++ b/rigs/aor/README.ar5000
@@ -24,4 +24,4 @@ Now, it:
 - filter-list mode/bw rearranged.
 - list of tuning steps corrected.
 
-Göran Sandin, SM6PPS
+GÃ¶ran Sandin, SM6PPS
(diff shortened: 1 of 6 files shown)


mdblack98:
And why do we want UTF-8?  If you look at the commit on github you can see it doesn't display correctly in UTF-8 where ISO-8859-1 does display correctly.

mdblack98:
I'll also add that ISO-8859-1 displays correctly on my Ubuntu system in vi and in Notepad in Windows.
So what does not work with ISO-8859-1?

dforsi:
We want UTF-8 for consistency, as I wrote.
Or we want ISO-8859-* for consistency and we convert 24 files, or we keep this minor inconsistency.
```
$ find -name '?*.[ch]' -exec file --brief {} + | perl -pe 's/, with.*//' | sort | uniq -c
     16 ASCII text
    670 C source, ASCII text
     11 C++ source, ASCII text
      5 C source, ISO-8859 text
     24 C source, Unicode text, UTF-8 text
      1 CSV ASCII text
      2 empty
```

Also on my system the same editor can show correctly the accents of ISO-8859-1 and of UTF-8 in different files because it can detect the encoding automatically, but the terminal can't do that.

PR #1676, opened 2025-03-05
More fixes for send_morse sync

GeoBaltz:
More of fix26 - add locks for rig_stop_morse, fix timeouts in kenwood_stop_morse, and clean up code from first batch.


GeoBaltz:
Any problems with this?

Issue #1500, opened 2024-01-31
Enable both static and dynamic library builds


mdblack98:
See if we can create both libhamlib.a and libhamlib.so during the same build.
Utilities can be built against the shared library I think so it's really just building the static library along with shared.

hamlibdk:
Intent of this request is to better support the integration of the FL-toolkit into the JTSDK.

Initial request asked for tools etc. that were say dynamically linked at the end of the process. On reflection the endpoint tools, if produced at all by this pathway, should be static linked.

hamlibdk:
WIth Ubuntu 24.04 LTS released this is now possibly of greater need. Many Linux distros base off the Ubuntu repos. 

Maintainers will not release .deb / snap packages that will not support both Static and Dynamic Endpoints.

N0NB:
I'm not sure what Mike's reason for disabling concurrent static and shared libraries with commit 43159e55a.  His commit message doesn't reference an issue and I didn't find anything relevant in that time frame when searching the closed issues.

My quick testing shows that commenting out the lines added in that commit restore the default behavior of building shared and static libraries concurrently.  The compiled utilities are dynamically linked to the shared library so I don't see an issue there.  I will continue to research and I may drop a revert commit in a few days for wider testing.

N0NB:
Fixed with PR #1694.

PR #1694, opened 2025-04-09
Restore libtool default of building shared and static libs

N0NB:
Commit 43159e5 prevented the libtool default of building both shared and static libraries during a compilation run.

This addresses GitHub issue 1500 that seeks to enable both shared and static libs.

At this point, I have not found the motivation for commit 43159e5 despite searching all forums and the mailing list.  While the reason might not now be ever known, restoring the default behavior alligns the project with other software package.


N0NB:
Reference issue #1500

PR #1696, opened 2025-04-16
Ignore files related to tests and delete generated files

dforsi:
This PR makes `git status` to ignore one file and one directory created when executing some tests one test program and makes `make clean` to delete them.


N0NB:
Thanks, Daniele.

PR #1708, opened 2025-04-21
Minor fixes/cleanups for 4.6.3

GeoBaltz:
Cleanups noticed while investigating reported rotctld problem - I think it was fixed in 4.6.x.


N0NB:
I guess I missed the dates in the NEWS file.  Good catch.

Thanks, George.

PR #1709, opened 2025-04-21
Remove dead code

dforsi:
This code:
```
            if (!optarg)
            {
                usage();    /* wrong arg count */
                exit(1);
            }
```
is never executed because two things can happen:
1) the argument that needs the option is the last one on the command line:
`getopt()/getopt_long()` will return `'?'` which is handled by the default case of the switch which prints an error message to stderr.

1) the argument that needs the option is followed by something:
`optarg` points to that, even if it starts with a `-`, so it's never `NULL`

For example running `rigctl -m -v` prints
> Unknown rig num 0, or initialization error.

because `atoi()` receives `-v` in `optarg` and converts it to zero.

I'm marking this PR as draft because I haven't yet decided if this code should be just removed or replaced with different checks, eg. checking if `optarg` starts with a dash which would also detect negative values, eg:
```
$ rigctl -m -1
Unknown rig num 4294967295, or initialization error.
Please check with --list option.
```
while this may be confusing for users, there is no data loss or equipment damage.

### DIFF
--- a/tests/ampctl.c
+++ b/tests/ampctl.c
@@ -160,32 +160,14 @@ int main(int argc, char *argv[])
             exit(0);
 
         case 'm':
-            if (!optarg)
-            {
-                usage();    /* wrong arg count */
-                exit(1);
-            }
-
             my_model = atoi(optarg);
             break;
 
         case 'r':
-            if (!optarg)
-            {
-                usage();    /* wrong arg count */
-                exit(1);
-            }
-
             amp_file = optarg;
             break;
 
         case 's':
-            if (!optarg)
-            {
-                usage();    /* wrong arg count */
-                exit(1);
-            }
-
             if (sscanf(optarg, "%d%1s", &serial_rate, dummy) != 1)
             {
                 fprintf(stderr, "Invalid baud rate of %s\n", optarg);
@@ -195,12 +177,6 @@ int main(int argc, char *argv[])
             break;
 
         case 'C':
-            if (!optarg)
-            {
-                usage();    /* wrong arg count */
-                exit(1);
-            }
-
             if (*conf_parms != '\0')
             {
                 strcat(conf_parms, ",");
@@ -217,12 +193,6 @@ int main(int argc, char *argv[])
             break;
 
         case 't':
-            if (!optarg)
-            {
-                usage();        /* wrong arg count */
-                exit(1);
-            }
-
             if (strlen(optarg) > 1)
             {
                 send_cmd_term = strtol(optarg, NULL, 0);
--- a/tests/ampctld.c
+++ b/tests/ampctld.c
@@ -203,32 +203,14 @@ int main(int argc, char *argv[])
             exit(0);
 
         case 'm':
-            if (!optarg)
-            {
-                usage();    /* wrong arg count */
-                exit(1);
-            }
-
             my_model = atoi(optarg);
             break;
 
         case 'r':
-            if (!optarg)
-            {
-                usage();    /* wrong arg count */
-                exit(1);
-            }
-
             amp_file = optarg;
             break;
 
         case 's':
-            if (!optarg)
-            {
-                usage();    /* wrong arg count */
-                exit(1);
-            }
-
             if (sscanf(optarg, "%d%1s", &serial_rate, dummy) != 1)
             {
                 fprintf(stderr, "Invalid baud rate of %s\n", optarg);
@@ -238,12 +220,6 @@ int main(int argc, char *argv[])
             break;
 
         case 'C':
-            if (!optarg)
-            {
-                usage();    /* wrong arg count */
-                exit(1);
-            }
-
             if (*conf_parms != '\0')
             {
                 strcat(conf_parms, ",");
@@ -260,22 +236,10 @@ int main(int argc, char *argv[])
             break;
 
         case 't':
-            if (!optarg)
-            {
-                usage();    /* wrong arg count */
-                exit(1);
-            }
-
             portno = optarg;
             break;
 
         case 'T':
-            if (!optarg)
-            {
-                usage();    /* wrong arg count */
-                exit(1);
-            }
-
             src_addr = optarg;
             break;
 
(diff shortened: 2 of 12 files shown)


N0NB:
What is the exit code with that code removed?

dforsi:
The exits are removed together with the entire `if` because that code is never executed; this can be seen adding a debug printf in the default case and issuing an incomplete command like `rigctl -m` 

A different change would be not applying this commit, and replacing only the checks with: `if (*optarg == '-')`
This would catch some user errors such as negative numbers and missing arguments between two options, but I'm not sure that it is worth the effort.

optarg is guaranteed to be non-NULL for mandatory arguments, `man optarg` says:
> If such a character is followed by a colon, the option requires an argument, so getopt() places a pointer
> to the following text in the same argv-element, or the text of the following argv-element, in optarg.

optarg would be NULL for optional arguments (those with two colons), but those aren't used in `tests/*`.

N0NB:
I wasn't clear.

When `get_opt` causes the program to exit, what value does it return to the shell?  Should we document that in the man page(s)?

dforsi:
It returns 1:
```
        default:
            usage();    /* unknown option? */
            exit(1);
```

The man pages for non-daemons already have it:
``` 
EXIT STATUS
       rigctl exits with:
       0      if all operations completed normally;
       1      if there was an invalid command line option or argument;
       2      if an error was returned by Hamlib.
```
and `rigsmtr.1` and `rigswr.1` also have:
```
       3      if the radio doesn't have the required capabilities.
```
which is different from error code 3 printed by `rigctl --help` and `rigctld --help`:
```
Error codes and messages
-0 - Command completed successfully
-1 - Invalid parameter
-2 - Invalid configuration
-3 - Memory shortage
-4 - Feature not implemented
-5 - Communication timed out
-6 - IO error
-7 - Internal Hamlib error
-8 - Protocol error
-9 - Command rejected by the rig
-10 - Command performed, but arg truncated, result not guaranteed
-11 - Feature not available
-12 - Target VFO unaccessible
-13 - Communication bus error
-14 - Communication bus collision
-15 - NULL RIG handle or invalid pointer parameter
-16 - Invalid VFO
-17 - Argument out of domain of func
-18 - Function deprecated
-19 - Security error password not provided or crypto failure
-20 - Rig is not powered on
-21 - Limit exceeded
-22 - Access denied
```

We could add codes 0, 1, 2 to the man  pages of daemons, but the other error codes aren't returned by tests because all `exit()`s have constant arguments.

N0NB:
Thanks, Daniele.

This looks good to me, but if you want to wait until 4.7, that is quite fine.  Probably better to wait to get some more testing rather than merging this before 4.6.3.

N0NB:
I think we can start working on merging this into `master`.

dforsi:
I think that this PR can be merged as-is (or I can rebase it to the current master to keep the history more straight).

The return code "3" in the man page is correct, however the man pages for the daemons are missing the section about return codes and some tools call `exit(2);` when they should call `exit(1);` for consistency, but these things can be fixed in a different PR because it isn't in the context of getopt  which was the target of this PR.

I checked that, as intended, this PR doesn't change the exit codes when the programs are called:

- without arguments
- with `--help` only
- with `--model` only (i.e. without a model number, which is an error)
- with `--model --version` (i.e. without a model number, which is an error)

I'm attaching to this comment the test scripts to be run from inside the `tests/` directory and I'm going to copy the results in the next comment.
[test-pr-1709-runner.sh.txt](https://github.com/user-attachments/files/20618724/test-pr-1709-runner.sh.txt)
[test-pr-1709-no-args.sh.txt](https://github.com/user-attachments/files/20618726/test-pr-1709-no-args.sh.txt)
[test-pr-1709-model-verbose.sh.txt](https://github.com/user-attachments/files/20618727/test-pr-1709-model-verbose.sh.txt)
[test-pr-1709-model.sh.txt](https://github.com/user-attachments/files/20618728/test-pr-1709-model.sh.txt)
[test-pr-1709-help.sh.txt](https://github.com/user-attachments/files/20618729/test-pr-1709-help.sh.txt)

dforsi:
Running the scripts in my previous comment gives the same return codes before and after this PS, as expected.

Collecting output on branch HEAD

test-pr-1709-help-HEAD.sh
```
ampctld Return Code=0
ampctl Return Code=0
rigctlcom Return Code=0
rigctld Return Code=0
rigctl Return Code=0
rigctlsync Return Code=0
rigctltcp Return Code=0
rigfreqwalk Return Code=1
rigmem Return Code=0
rigsmtr Return Code=0
rigswr Return Code=0
rigtestlibusb Return Code=1
rigtestmcast Return Code=1
rotctld Return Code=0
rotctl Return Code=0
```

test-pr-1709-model-HEAD.sh
```
ampctl Return Code=1
rigctlcom Return Code=1
rigctld Return Code=1
rigctl Return Code=1
rigctlsync Return Code=1
rigctltcp Return Code=1
rigfreqwalk Return Code=1
rigmem Return Code=1
rigsmtr Return Code=1
rigswr Return Code=1
rigtestlibusb Return Code=1
rigtestmcast Return Code=1
rotctld Return Code=1
rotctl Return Code=1
```

test-pr-1709-model-verbose-HEAD.sh
```
ampctl Return Code=2
rigctlcom Return Code=2
rigctld Return Code=2
rigctl Return Code=2
rigctlsync Return Code=2
rigctltcp Return Code=2
rigfreqwalk Return Code=1
rigmem Return Code=1
rigsmtr Return Code=2
rigswr Return Code=1
rigtestlibusb Return Code=1
rigtestmcast Return Code=1
rotctld Return Code=2
rotctl Return Code=2
```

test-pr-1709-no-args-HEAD.sh
```
ampctl Return Code=0
rigctlcom Return Code=2
rigctl Return Code=0
rigctlsync Return Code=1
rigctltcp Return Code=2
rigfreqwalk Return Code=1
rigmem Return Code=1
rigswr Return Code=1
rigtestlibusb Return Code=0
rigtestmcast Return Code=0
rotctl Return Code=0

N0NB:
I'm not too concerned about a straight line of history.  I gave up on that long long ago!

Thanks, Daniele.

PR #1715, opened 2025-04-26
Fix typos

dforsi:
Fixed with codespell.


N0NB:
As you can see, we're not always terrific spellers.  🤣

Thanks, Daniele.

PR #1718, opened 2025-04-30
Fix/typos

dforsi:
The first commit replaces a string in an error message that was copied to all simulators.
The second commit removes examples of compile commands that are wrong and/or incomplete.
The other commits fix Doxygen comments and in some cases (commit 56b075ab and 24eafbd2) move a definition to be below the relevant comment.


N0NB:
Just curious, other than the commit message for 3448c73 where is the compilation command for the simulators documented?

N0NB:
Thanks, Daniele.

PR #1719, opened 2025-05-01
Add support of setting IC-705 WFM mode

jj1bdx:
On IC-705, setting WFM mode via ICOM CI-V command 0x26 does not work.
This patch implements a workaround using CI-V command 0x06 instead via `icom_set_mode_without_data()` for IC-705 setting mode WFM.


N0NB:
Thank you for your contribution to Hamlib, Kenji.

PR #1723, opened 2025-05-02
Fix IC-705 COMP, VD, and ID meter calibration values

jj1bdx:
IC-705-specific values are ported from flrig-2.0.05.93, for Comp Level, Vd, and Id.
My observation of actual radio transmissions indicates that the fixed values are consistent with those of the IC-705 meters.


Tyrbiter:
Might be sensible to get these into Hamlib rather than freedv, which in any case uses Hamlib.

I'm not sure about these values in Hamlib, but it would be sensible to look at the Hamlib github for more information.

N0NB:
Thank you, Kenji.

PR #1724, opened 2025-05-03
Fix IC-705 filter selection and bandwidth handling for FM and WFM

jj1bdx:
This pull request (PR) solves the bug for IC-705 with rigctl when entering the command `M FM 0` after `M WFM 0` *did not* change the mode properly to (narrow) FM. This PR also explicitly defines WFM bandwidth to 200 kHz, which is the case for IC-705, and the only valid filter choice for WFM is 1 (one).

This PR includes the following fixes:

* A partial rollback of https://github.com/Hamlib/Hamlib/pull/1719/commits/a395b91be6bd19d760e57005ecb8079b15af9ede; IC-705 now uses `icom_set_mode_x26()` instead of `icom_set_mode_without_data()`
* Enable `.fm_filters` in `IC705_priv_caps`, which is mandatory for handling FM filter settings properly
* `icom2rig_mode()`: handle FM and WFM separately and correctly at least for IC-705, no changes for IC-7300 and IC-9700
* `icom_get_mode_without_data()`: add WFM to the code, which assumes that values from `icom2rig_mode()` are correct
* `icom_set_mode_x26()`: When `RIG_PASSBAND_NORMAL` or `RIG_PASSBAND_NOCHANGE` are passed to the parameter `width`, the command value `buf[2]` is forcefully set to 1 (the widest bandwidth (15kHz for IC-705, IC-7300, and IC-9700)). This prevents an invalid value of 0 from being passed to the radio as a filter selection value of the command 0x26.
* Add WFM freq to `ic705_caps.filters`.

### DIFF
--- a/rigs/dummy/aclog.c
+++ b/rigs/dummy/aclog.c
@@ -19,6 +19,7 @@
 *
 */
 
+#include <ctype.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -421,9 +422,14 @@ static rmode_t modeMapGetHamlib(const char *modeACLog)
 /*
 * aclog_get_freq
 * Assumes rig!=NULL, STATE(rig)->priv!=NULL, freq!=NULL
+*
+* string='<CMD><READBMFRESPONSE><BAND>23</BAND><MODE>SSB</MODE><MODETEST>PH</MODETEST><FREQ>1,296.171100</FREQ></CMD> '
 */
 static int aclog_get_freq(RIG *rig, vfo_t vfo, freq_t *freq)
 {
+    int i, j = 0;
+    char f_string[32];
+
     char value[MAXARGLEN];
     struct aclog_priv_data *priv = (struct aclog_priv_data *) STATE(rig)->priv;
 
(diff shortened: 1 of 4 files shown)


jj1bdx:
@N0NB
I am trying to squash multiple commits into a single commit. I will let you know here in another comment when it's ready to go.

jj1bdx:
@N0NB
I've squashed the old set of commits into a single commit of https://github.com/Hamlib/Hamlib/commit/abed4c8d964df5698df42c96b39c217735920818.
It's ready to merge. Please take a look and review.

N0NB:
Hi Kenji.

I manually merged commit abed4c8.  No question that `rebase -i` takes a little practice, especially with pushing to a remote repository.  I do try to avoid doing rebase after a push as things can get confusing and so do any rebase operations on a set of commits before I do any push operation.

Thanks again.  We appreciate your contribution, especially as I want to release 4.6.3 within the next several days and your fixes will be useful to a lot of users.

jj1bdx:
@N0NB Sorry for messing around with force updating. Thanks for merging!

PR #1729, opened 2025-05-08
Fix/build

dforsi:
This PR:
* adds some build rules to run make from inside of some subdirectories
* removes some unused/uneeded dependencies in makefiles
* adds a check to avoid a segfault
* fixes a syntax error in a define
* fixes some return values (errors should be negative)


src/Makefile.am
N0NB, line 1: Running `bootstrap.sh` results in this warning from this change:

```
src/Makefile.am:34: warning: '%'-style pattern rules are a GNU make extension
```

I'm unsure if this would cause an issue on platforms that don't use GNU make (Mac OS?).

dforsi: Yes, that could cause issues so I replaced it with the portable rule suggested in https://stackoverflow.com/a/27975701 that is also more readable because the list of dependencies is a little more explicit and that list will always be up to date, so it doesn't require any extra maintenance effort.
I'm not squashing this commit with the other one because there is a value in keeping this step, IMHO.

While testing this new rule I fixed two other issues.

N0NB:
It's likely this is unrelated to any of these changes as it has been a long time since I did test builds of any of the bindings.  When enabling the Perl bindings I get this from `make`:

``` console
Making all in bindings
make[1]: Entering directory '/home/nate/git/hlb/bindings'
  GEN      hamlibperl_wrap.c
../../hamlib/include/hamlib/rig.h:1872: Warning 314: 'next' is a perl keyword
../../hamlib/include/hamlib/rig.h:1880: Warning 314: 'last' is a perl keyword
make  all-am
```

It appears these Perl keywords are used as variable names for some structures in `rig.h`:

``` c
/**
 * Config item for deferred processing
 **/
struct deferred_config_item {
  struct deferred_config_item *next;
  hamlib_token_t token;
  char *value;                  // strdup'ed, must be freed
};
typedef struct deferred_config_item deferred_config_item_t;

struct deferred_config_header {
  struct deferred_config_item *first;   // NULL if none
  struct deferred_config_item *last;
};
typedef struct deferred_config_header deferred_config_header_t;
```

N0NB:
Going through and building the bindings individually and running `make check` shows no failures.  I think this can go into 4.6.3.

dforsi:
This can go in 4.6.3

I see that the warnings appeared also in 4.6.2:
```
git checkout 4.6.2
./bootstrap
./configure --with-perl-binding
make -C bindings 2>&1 | grep "perl keyword"
../include/hamlib/rig.h:1873: Warning 314: 'next' is a perl keyword
../include/hamlib/rig.h:1881: Warning 314: 'last' is a perl keyword
```

Adding some ignore lines to `ignore.swg` doesn't fix the warnings but prevents the definitions to appear in `hamlibperl_wrap.c`:
```
%ignore deferred_config_item;
%ignore deferred_config_item_t;
%ignore deferred_config_header;
%ignore deferred_config_header_t;
```
so I suggest to keep the warning for now and look into adding some `#ifndef SWIG` in `rig.h` instead for a future release that could contain more changes related to bindings.

N0NB:
Thanks, Daniele.

PR #1733, opened 2025-05-12
Add names to threads created in src/*.c, if possible.

GeoBaltz:
So I can tell them apart from application threads.
Also fix missing/misplaced '#include "config.h"' directives so they compile portably.

Started on this while fixing thread locking, but never got it fit to merge.


GeoBaltz:
I give up.  Some days I think macOS goes out of it's way to discourage portable code.

PR #1735, opened 2025-05-14
Fix usage and usage texts

### LINKED ISSUE #1728: verbose cumulative explanation
In Hamlib everywhere I see `verbose++` not `verbose=1`, example
https://github.com/Hamlib/Hamlib/blob/6dfa118dace16c3fff039d071ca87a4210c91793/tests/rigctld.c#L568

Because of this style user can tune verbose level, for example, double verbose `rotctld -vv`. But `rotctld --help` doesn't provide details, just
```
-v, --verbose                 set verbose mode, cumulative
```
But how many levels of verbose exists? Hamlib contributors plan to add such info in `--help`? Provide maximum available verbose level at least pls!

dforsi:
This PR changes some code:
- removes `-z` from rigctld because it isn't implemented
- fixes `--mapa2b` `--version` and `--help` in rigctlsync

and changes some strings:
- adds `-v to -vvvvv` to all programs that support the option `--verbose` (closes #1728)
- adds `-l` to all programs that support this option `--model`
- adds some missing option arguments
- adds the description of the option `-` to all programs that support it
- change some texts to be identical or more similar

### DIFF
--- a/tests/ampctl.c
+++ b/tests/ampctl.c
@@ -490,7 +490,9 @@ void usage()
         "  -v, --verbose                 set verbose mode, cumulative (-v to -vvvvv)\n"
         "  -Z, --debug-time-stamps       enable time stamps for debug messages\n"
         "  -h, --help                    display this help and exit\n"
-        "  -V, --version                 output version information and exit\n\n"
+        "  -V, --version                 output version information and exit\n"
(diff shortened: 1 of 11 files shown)


N0NB:
Thanks, Daniele.

PR #1739, opened 2025-05-17
Fix build when USRP is enabled

dforsi:
The USRP code may be old but these fixes are trivial (a missing include ad some missing parentheses and spaces).

This allows `make distcheck AM_DISTCHECK_CONFIGURE_FLAGS="--enable-usrp"` to succeed.

Compile-tested because I do not own the hardware.

### DIFF
--- a/rigs/kit/usrp.c
+++ b/rigs/kit/usrp.c
@@ -19,6 +19,8 @@
  *
  */
 
+#include <hamlib/config.h>
+
 /*
  * Compile only this model if usrp is available
  */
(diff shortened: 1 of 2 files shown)


dforsi:
On Linux, having installed libusrp in `$HOME`, the proper paths need to be passed to configure:
> ./configure --enable-usrp PKG_CONFIG_PATH=$HOME/lib/pkgconfig

or more explicitly:
> ./configure --enable-usrp USRP_CFLAGS="-I $HOME/include/usrp" USRP_LIBS="-l $HOME/lib"

PR #1744, opened 2025-05-26
Add locking around get/set levels

GeoBaltz:
Fixes CW sending with tlf.  Fixes conflicts between morse_data_handler() and rig_get_level(,,RIG_LEVEL_KEYSPD,), which caused various timeouts and protocol errors.

Should probably be done for any HAMLIB_API rig_[gs]et_<something> that may involve I/O to the rig.


N0NB:
George, do you think this is safe to apply to 4.6.3 as well?

PR #1745, opened 2025-05-26
Fix/typos

dforsi:
This PR:

- fixes the formatting of a man page (the following paragraph was rendered as verbatim text with hard line breaks)
- fixes a typing error in a debug message in the dummy driver

This could wait for 4.7.0


N0NB:
Thanks, Daniele.

PR #1747, opened 2025-05-27
Fix list of dependecies of aclocal.m4

dforsi:
This PR adds missing ax_lib_readline.m4 and ax_lua.m4 to the list of MACROS and sorts the list.


N0NB:
Thanks, Daniele.

PR #1751, opened 2025-05-30
Fix sign of RIG_E* return values

dforsi:
All constant error values RIG_E* should be negated when used.

The first patch is just for consistency, it's trivial because it changes a macro definition and where it's used.
The second patch is also trivial because it changes only Doxygen comments (this could be cherry-picked to 4.6.3).
The third patch is tricky because it's hard to test, so I'm marking this PR as draft.

Apparently the hunks like this:
```
         if (!q)
         {
-            return RIG_EINVAL;
+            return -RIG_EINVAL;
         }
```
make no difference because the value returned is compared against `RIG_OK` and if different it gets  passed to `rigerror()` that takes the absolute value of its argument.

### DIFF
--- a/rigs/icom/icom.c
+++ b/rigs/icom/icom.c
@@ -1232,7 +1232,7 @@ retry_open:
     {
         // maybe we need power on?
         rig_debug(RIG_DEBUG_VERBOSE, "%s trying power on\n", __func__);
-        retval = abs(rig_set_powerstat(rig, 1));
+        retval = rig_set_powerstat(rig, 1);
 
         // this is only a fatal error if powerstat is implemented
         // if not implemented than we're at an error here
@@ -1243,7 +1243,7 @@ retry_open:
             rig_debug(RIG_DEBUG_ERR, "%s: rig_set_powerstat failed: %s\n", __func__,
                       rigerror(retval));
 
-            if (retval == RIG_ENIMPL || retval == RIG_ENAVAIL)
+            if (retval == -RIG_ENIMPL || retval == -RIG_ENAVAIL)
             {
                 rig_debug(RIG_DEBUG_ERR, "%s: rig_set_powerstat not implemented for rig\n",
                           __func__);
@@ -1316,11 +1316,11 @@ int icom_rig_close(RIG *rig)
     {
         // maybe we need power off?
         rig_debug(RIG_DEBUG_VERBOSE, "%s trying power off\n", __func__);
-        int retval = abs(rig_set_powerstat(rig, 0));
+        int retval = rig_set_powerstat(rig, 0);
 
         // this is only a fatal error if powerstat is implemented
         // if not implemented than we're at an error here
-        if (retval != RIG_OK && retval != RIG_ENIMPL && retval != RIG_ENAVAIL)
+        if (retval != RIG_OK && retval != -RIG_ENIMPL && retval != -RIG_ENAVAIL)
         {
             rig_debug(RIG_DEBUG_WARN, "%s: unexpected retval here: %s\n",
                       __func__, rigerror(retval));
--- a/rigs/kenwood/ic10.c
+++ b/rigs/kenwood/ic10.c
@@ -152,7 +152,7 @@ static int get_ic10_if(RIG *rig, char *data)
 {
     const struct kenwood_priv_caps *priv = (struct kenwood_priv_caps *)
                                            rig->caps->priv;
-    int i, data_len, retval = RIG_EINVAL;
+    int i, data_len, retval = -RIG_EINVAL;
 
     rig_debug(RIG_DEBUG_TRACE, "%s: called\n", __func__);
 
--- a/rigs/kenwood/k3.c
+++ b/rigs/kenwood/k3.c
@@ -2352,7 +2352,7 @@ int k3_get_level(RIG *rig, vfo_t vfo, setting_t level, value_t *val)
         }
         else
         {
-            return RIG_EPROTO;
+            return -RIG_EPROTO;
         }
 
         // extended K22 format PCnnnx where 0=.1W units and 1=1W units
--- a/rigs/pcr/pcr.c
+++ b/rigs/pcr/pcr.c
@@ -589,7 +589,7 @@ pcr_open(RIG *rig)
 
     rig_flush(rp);
 
-    /* return RIG_ERJCTED if power is off */
+    /* return -RIG_ERJCTED if power is off */
     err = pcr_transaction(rig, "H1?");
 
     if (err != RIG_OK)
@@ -1939,7 +1939,7 @@ int pcr_get_powerstat(RIG *rig, powerstat_t *status)
     struct pcr_priv_data *priv = (struct pcr_priv_data *) STATE(rig)->priv;
     int err;
 
-    /* return RIG_ERJCTED if power is off */
+    /* return -RIG_ERJCTED if power is off */
     err = pcr_transaction(rig, "H1?");
 
     if (err != RIG_OK && err != -RIG_ERJCTED)
--- a/rigs/tentec/orion.c
+++ b/rigs/tentec/orion.c
@@ -425,7 +425,7 @@ int tt565_set_freq(RIG *rig, vfo_t vfo, freq_t freq)
      * matches the VFO A/B setting. c.f. rig_get_range().
      * Recall VFOA = ham only, VFOB = gen coverage for Hamlib.
      * (We assume VFOA = Main RXTX and VFOB = Sub RX.)
-     * If outside range, return RIG_ERJECTED for compatibility vs icom.c etc.
+     * If outside range, return -RIG_ERJECTED for compatibility vs icom.c etc.
      */
     in_range = FALSE;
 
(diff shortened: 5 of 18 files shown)


N0NB:
I think we can start working on merging this into `master`.

rigs/icom/icom.c
dforsi, line 1: These are ok because after removing `abs()` the errors are negative.

rigs/kenwood/ic10.c
```diff
@@ -152,7 +152,7 @@ static int get_ic10_if(RIG *rig, char *data)
 {
     const struct kenwood_priv_caps *priv = (struct kenwood_priv_caps *)
                                            rig->caps->priv;
-    int i, data_len, retval = RIG_EINVAL;
```
dforsi, line 155: This is consistent with the other return value of this function which is `-RIG_ERJCTED`.

rigs/kenwood/k3.c
```diff
@@ -2352,7 +2352,7 @@ int k3_get_level(RIG *rig, vfo_t vfo, setting_t level, value_t *val)
         }
         else
         {
-            return RIG_EPROTO;
```
dforsi, line 2355: This is consistent with the other return values of this function.

rigs/pcr/pcr.c
dforsi, line 1: These are comments, change them for consistency

rigs/tentec/orion.c
```diff
@@ -425,7 +425,7 @@ int tt565_set_freq(RIG *rig, vfo_t vfo, freq_t freq)
      * matches the VFO A/B setting. c.f. rig_get_range().
      * Recall VFOA = ham only, VFOB = gen coverage for Hamlib.
      * (We assume VFOA = Main RXTX and VFOB = Sub RX.)
-     * If outside range, return RIG_ERJECTED for compatibility vs icom.c etc.
```
dforsi, line 428: These is a comment change it for consistency.

rigs/yaesu/ft757gx.c
```diff
@@ -508,7 +508,7 @@ static int ft757gx_get_freq(RIG *rig, vfo_t vfo, freq_t *freq)
         return RIG_OK;
     }
 
-    return RIG_ENAVAIL;
```
dforsi, line 511: This function is called via `rig_caps` so it must return a negative error.

rotators/grbltrk/grbltrk.c
```diff
@@ -447,7 +447,7 @@ grbltrk_rot_set_conf(ROT *rot, hamlib_token_t token, const char *val)
         if (retval < 0)
         {
             rot_debug(RIG_DEBUG_ERR, "grbl_request [%s] fail\n", val);
-            return RIG_EIO;
```
dforsi, line 450: This function is called via `rot_caps` so it must return a negative error.

rotators/grbltrk/grbltrk.c
```diff
@@ -205,7 +205,7 @@ grbl_init(ROT *rot)
         if (retval != RIG_OK)
         {
             rot_debug(RIG_DEBUG_ERR, "grbl_request [%s] fail\n", grbl_init_list[i]);
-            return RIG_EIO;
```
dforsi, line 208: Return values are silently ignored by the caller. Looks like different bug (see ##1759).

rotators/meade/meade.c
```diff
@@ -310,7 +310,7 @@ static int meade_set_position(ROT *rot, azimuth_t az, elevation_t el)
     {
         rig_debug(RIG_DEBUG_VERBOSE, "%s: expected 110, got %s\n", __func__,
                   return_str);
-        return RIG_EINVAL;
```
dforsi, line 313: This function is called via `rot_caps` so it must return a negative error.

rotators/skywatcher/skywatcher.c
dforsi, line 1: These makes no difference because the return value is passed to the `ERROR_CHECK()` macro which checks for `!= RIG_OK`.

src/amp_conf.c
dforsi, line 1: In this file only Doxygen comments are changed adding a minus sign.

src/amplifier.c
dforsi, line 1: In this file only Doxygen comments are changed adding a minus sign.

src/cm108.c
dforsi, line 1: In this file only Doxygen comments are changed adding a minus sign.

src/event.c
dforsi, line 1: In this file only Doxygen comments are changed adding a minus sign.

src/extamp.c
dforsi, line 1: In this file only Doxygen comments are changed adding a minus sign.

src/locator.c
dforsi, line 1: In this file only Doxygen comments are changed adding a minus sign.

src/rot_conf.c
dforsi, line 1: In this file only Doxygen comments are changed adding a minus sign.

src/rig.c
```diff
@@ -7448,7 +7448,7 @@ int HAMLIB_API rig_send_morse(RIG *rig, vfo_t vfo, const char *msg)
 {
     const struct rig_caps *caps;
     struct rig_state *rs;
-    int retcode = RIG_EINTERNAL, rc2;
```
dforsi, line 7451: Consistent with the other return values of this function.

src/rig.c
```diff
@@ -984,8 +984,8 @@ RIG *HAMLIB_API rig_init(rig_model_t rig_model)
  * a negative value if an error occurred (in which case, cause is
  * set appropriately).
  *
- * \retval RIG_EINVAL   \a rig is NULL or inconsistent.
- * \retval RIG_ENIMPL   port type communication is not implemented yet.
```
dforsi, line 988: Doxygen comments.

src/rot_ext.c
dforsi, line 1: In this file only Doxygen comments are changed adding a minus sign.

src/rot_settings.c
dforsi, line 1: In this file only Doxygen comments are changed adding a minus sign.

src/rotator.c
dforsi, line 1: In this file only Doxygen comments are changed adding a minus sign.

tests/rigctl.c
dforsi, line 1: Ok because I'm also changing signs in the body of the macro.

tests/rigctld.c
dforsi, line 1: Ok because I'm also changing signs in the body of the macro.

tests/rigctltcp.c
dforsi, line 1: Ok because I'm also changing signs in the body of the macro.

tests/ampctl_parse.c
```diff
@@ -1393,7 +1393,7 @@ int ampctl_parse(AMP *my_amp, FILE *fin, FILE *fout, char *argv[], int argc)
     pthread_mutex_unlock(&amp_mutex);
 #endif
 
-    if (retcode == RIG_EIO) { return retcode; }
```
dforsi, line 1396: The new check is ok, but the return is a different bug (it should return `2` meaning Hamlib error, so I'm going to open a new PR).

tests/ampctl_parse.c
```diff
@@ -1608,7 +1608,7 @@ int set_conf(AMP *my_amp, char *conf_parms)
 
         if (!q)
         {
-            return RIG_EINVAL;
```
dforsi, line 1611: This change is ok because the code is parsing a line was supposed to contain an equal sign but it is not.

tests/rigsmtr.c
```diff
@@ -394,7 +394,7 @@ int set_conf_rig(RIG *rig, char *conf_parms)
 
         if (!q)
         {
-            return RIG_EINVAL;
```
dforsi, line 397: This change is ok because the code is parsing a line was supposed to contain an equal sign but it is not.

tests/rigswr.c
```diff
@@ -346,7 +346,7 @@ static int set_conf(RIG *rig, char *conf_parms)
 
         if (!q)
         {
-            return RIG_EINVAL;
```
dforsi, line 349: This change is ok because the code is parsing a line was supposed to contain an equal sign but it is not.

tests/rotctl_parse.c
```diff
@@ -1472,7 +1472,7 @@ int rotctl_parse(ROT *my_rot, FILE *fin, FILE *fout, const char *argv[],
     pthread_mutex_unlock(&rot_mutex);
 #endif
 
-    if (retcode == RIG_EIO) { return retcode; }
```
dforsi, line 1475: The new check is ok, but the return is a different bug (it should return `2` meaning Hamlib error, so I'm going to open a new PR).

dforsi (commented):
It seems that after I did a `git push --force` the previous comments were deleted, anyway I looked at all source and I think that all the changes are good.

The changes to flrig.c may cause different settings to be visible to the users, but they would be consistent to what flrig is reporting.

N0NB:
Thanks, Daniele.

PR #1752, opened 2025-06-02
Fix Icom spectrum data callback

GeoBaltz:
All commits go into master.
Commits 1 (Pass data), 2 (async_data_handler) and 5 (NEWS) should go into 4.6.x, assuming that icon Stuart added means what I think it means.


N0NB:
Thanks George.

Commits 1, 2, and 5 cherry-picked to Hamlib-4.6.3 as requested.

PR #1754, opened 2025-06-05
Changes for JRC NRD-535D and NRD-545

markjfine:
1. Removed RIG_FUNC_FAGC from 535D as erroneous.
2. Added RIG_FUNC_NB2 functionality to both 535D and 545. Since they both use the same command and affect each other, it is combined with the set and get functions for RIG_FUNC_NB.


N0NB:
It looks like these should also be cherry-picked into 4.6.3 as well.  Any objection?

markjfine:
I have no problem with that.

N0NB:
Thanks, Mark.

markjfine:
No problem.

markjfine:
@N0NB 
Unfortunately, this created a few warnings when built on Mac that I didn't see on Linux yesterday.
I created a separate PR to deal with it.

PR #1755, opened 2025-06-06
Kill warnings

markjfine:
Kills some warnings in jrc_set_func when compiling on Mac that weren't seen in Linux.


N0NB:
Thanks, Mark.

PR #1757, opened 2025-06-06
Fix exit statuses

dforsi:
This PS makes the error statuses consistent with the man pages (in other words: `1` for errors caused by user input, `2` for errors returned by rig/rotator/amplifier).


N0NB:
Thanks, Daniele.

PR #1758, opened 2025-06-06
Fix Doxygen syntax

dforsi:
This PR just fixes the markup used by Doxygen, not the contents.


N0NB:
I think this can go into 4.6.3 as well.

N0NB:
Thanks, Daniele.

PR #1760, opened 2025-06-08
Reduce output from cppcheck.sh

GeoBaltz:
Get rid of some more style nagging - add const/static modifiers, drop some unneeded code.
Make cppcheck.sh aware of out-of-tree building - try to find generated include files.

For master/4.7 only

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -14,6 +14,8 @@ Version 5.x -- future
 
 Version 4.7.0
         * 2025-12-01 (target)
+	* Reduce/repair excess output from cppcheck.sh - mostly cosmetic changes (WIP)
+	  Output from `wc -l cppcheck.log` - 4.6.2: 981  now: 732
         * Remove dead getopt code.  GitHub PR #1709. (TNX Daniele Forsi)
         * Move rig_cache to separate(calloc) storage. Prepare for other moves.
           Issue #1420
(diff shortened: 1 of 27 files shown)


N0NB:
George, have you seen my comments regarding `k3.c`?

GeoBaltz:
No, Nate.  How did you send them?

GeoBaltz:
Looking back at k3.c...

I do see some stuff out of order, and one func that cppcheck missed.  If that's what you were commenting on, I'll push another commit to this branch.  Let me know.

N0NB:
Yes, please push another commit when you can.

GeoBaltz:
I think I have k3.c back in order.  Still haven't seen your original comments, Nate, so I hope this is what you were writing about.

N0NB (approved):

N0NB:
Apparently my comments stayed in the GH review process.  No matter.  I've learned another of its quirks.

Regardless, it looks good now.

My guess is that `cppcheck` upon finding the duplicated comment has logic to remove the second rather than the former.  It did the right thing the wrong way for our purposes.

I like to scroll through the commits the commits before merging and I wondered why the comment was being removed, as if it might be out of date or some such.  Then I looked at the copy in the `master` branch and saw the duplication.  When you said these changes were from `cppcheck` I guessed that its algorithm led it to do this.

Thanks George.  Much appreciated.

GeoBaltz:
Close, but cppcheck only gave hints.  The selection of which to drop was my fault.  When I went back and really looked at the final code I saw what you saw - I dropped the wrong stuff.  Anyway it looks good now.
Thank & 73
n3gb

Issue #1761, opened 2025-06-08
IC-705: "unknown mode=WFM" when using rigctl to switch to WFM


tmiw:
When using `rigctl` to control my IC-705, I noticed that it doesn't recognize WFM mode (despite the radio actually supporting it; ref page 16 from [the CI-V manual for the IC-705](https://www.icomeurope.com/wp-content/uploads/2020/08/IC-705_ENG_CI-V_1_20200721.pdf)).

Version info:

```
MacBookPro:build_osx mooneer$ ../hamlib/bin/rigctl -m 3085 -r /dev/cu.usbmodem142201 -vvvv
rigctl.c(532) Startup: ../hamlib/bin/rigctl -m 3085 -r /dev/cu.usbmodem142201 -vvvv
rigctl Hamlib 4.6.2 2025-02-09T21:03:50Z SHA=870364 64-bit
Report bugs to <hamlib-developer@lists.sourceforge.net>
```

Debugging output when attempting to switch modes:

```
Rig command: m
rigctl_parse.c(2338):rigctl_get_mode entered
**2:rig.c(2979):rig_get_mode entered
icom.c(2898):icom_get_mode entered
icom_get_mode called vfo=currVFO
icom.c(2733):icom_get_mode_without_data entered
icom_get_mode_without_data called vfo=currVFO
icom_get_vfo_number_x25x26(10016): current_vfo=currVFO, vfo=currVFO -> vfo_number=0
***3:frame.c(525):icom_transaction entered
icom_transaction: cmd=0x26, subcmd=0x00, payload_len=0
****4:frame.c(145):icom_one_transaction entered
****4:frame.c(503):icom_one_transaction returning(0) 
***3:frame.c(556):icom_transaction returning(0) 
icom_get_dsp_flt called, mode=USB
***3:frame.c(525):icom_transaction entered
icom_transaction: cmd=0x1a, subcmd=0x03, payload_len=0
****4:frame.c(145):icom_one_transaction entered
****4:frame.c(503):icom_one_transaction returning(0) 
***3:frame.c(556):icom_transaction returning(0) 
from_bcd called
icom.c(2116):icom_get_dsp_flt returning2(3000) 
icom.c(2881):icom_get_mode_without_data returning2(0) 
icom_get_mode mode=4
icom_get_mode databuf[2]=0, mode=4
icom.c(3038):icom_get_mode returning2(0) 
***3:cache.c(40):rig_set_cache_mode entered
[...]

PR #1762, opened 2025-06-13
Fix 0x26 command in simic705.c - off by 1 byte

GeoBaltz:
Reduce usage of rig.h idioms - use Icom values.
Reduce/cleanup some debug output.
Add width set (1A 03)

Found/fixed while on a wild goose chase for issue #1761


N0NB:
Thanks, George.

PR #1765, opened 2025-06-15
Verbosity reduction

GeoBaltz:
Silence the batch of RIG_DEBUG_ERR messages produced by \dump_caps.  They're not really errors, but are rather jarring when you're checking out a new application, like I was with not1mm.
And the usual batch of cppcheck nags.


N0NB:
Thanks, George.

Issue #1767, opened 2025-06-17
kenwood_transaction() is erroneously stripping eol characters, breaking hamlib for tm-d710/tm-v71a devices.


larsks:
I was seeing the symptoms reported in #1698. I spent some time debugging the code, and the problem is that by the time we reach this block of code:

https://github.com/Hamlib/Hamlib/blob/aca0b2d1c63eb27b23ce376eaf630da169c17a0a/rigs/kenwood/kenwood.c#L483-L495

We have already called `remove_nonprint(buffer)` at line 448:

https://github.com/Hamlib/Hamlib/blob/aca0b2d1c63eb27b23ce376eaf630da169c17a0a/rigs/kenwood/kenwood.c#L446-L448

This strips the command termination character (`\r`), causing the "Check that command termination is correct" to always fail, effectively breaking hamlib for all TM-D710/TM-V71 type devices.

---

We can confirm this behavior by inspecting the value of `buffer` before and after the call to `remove_nonprint`:

```
Breakpoint 3, kenwood_transaction (rig=rig@entry=0x435d60, cmdstr=cmdstr@entry=0x7ffff740176e "ID", data=data@entry=0x7fffffffc300 "",
    datasize=datasize@entry=128) at kenwood.c:443
443         rig_debug(RIG_DEBUG_TRACE, "%s: read_string len=%d '%s'\n", __func__,
(gdb) p buffer
$1 = "ID TM-V71\r", '\000' <repeats 117 times>
```

This is the `rig_debug` call just before the call to `remove_nonprint`; we can see that the buffer has the `\r` command termination character. After calling `remove_nonprint`...

```
(gdb) until 450
kenwood_transaction (rig=rig@entry=0x435d60, cmdstr=cmdstr@entry=0x7ffff740176e "ID", data=data@entry=0x7fffffffc300 "", datasize=datasize@entry=128)
    at kenwood.c:450
450         if (retval < 0)
(gdb) p buffer
$2 = "ID TM-V71", '\000' <repeats 118 times>
```

...we see that the termination character has been dropped.

larsks:
It looks like this issue was caused by d1e0e3f204d3b0f5b3a17e83a50d360184136e0d, which was apparently never tested on these devices. I've created a pull request that reverts that commit so that we can un-break hamlib until someone can create a less disruptive solution to #1652.

N0NB:
Hi Lars.

[...]

Issue #1698, opened 2025-04-19
Kenwood TM-D710 Error


promo776:
Homebrew hamlib 4.6.2

Console output:

```
rig_init: backend for TM-D710(G) does not contain hamlib_check_rig_caps
kenwood_transaction: Command is not correctly terminated 'BC 0,0'
kenwood_transaction: Command is not correctly terminated 'BC 0,0'
kenwood_transaction: Command is not correctly terminated 'BC 0,0'
rig_get_vfo: returning -8(
STATE(rig)->tx_vfo: VFOA
**2:rig.c(3452):rig_get_vfo entered
rig_get_vfo: cache miss age=1000000ms
**rig.c(3488) trace
tmd710_get_vfo: called
tmd710_get_vfo_and_mode: called
tmd710_get_vfo_num: called
kenwood_transaction called cmd=BC
kenwood_transaction: cmdstr = BC
read_string_generic called, rxmax=4095 direct=1, expected_len=1
write_block(): TX 3 bytes
read_string_generic called, rxmax=11 direct=1, expected_len=1
read_string_generic(): RX 7 characters, direct=1
'enwood_transaction: read_string len=7 'BC 0,0
kenwood_transaction: Command is not correctly terminated 'BC 0,0'
kenwood.c(711):kenwood_transaction returning2(-8) Protocol error

Protocol error
Protocol error
)
kenwood_transaction: Command is not correctly terminated 'FO 0,0441500000,0,0,0,0,0,0,13,41,051,05000000,1'
Opened rig model 2034, 'TM-D710(G)'
```

promo776:
It seems like the issue is related to _transaction_read_ in **kenwood.c**:

```
transaction_read:
    /* allow room for most any response */
    // this len/expected stuff is confusing -- logic in some places includes the semicolon
    // so we add 1 to our read_string length to cover these cases
    // eventually we should be able to get rid of this but requires testing all Kenwood rigs
```

When I look at the output from a serial console, the D710 is not sending a semicolon:

BC 0,0

larsks:
@promo776 please try out the change in #1768; that should resolve the error.

N0NB:
Fixed by PR #1768.

Issue #1652, opened 2025-01-06
uSDX CAT error in WSJT-X


brightkill:
Greetings!

My transceiver - uSDX (uses Kenwood TS-480 protocol with hamlib)
I use FT232RL adapter with 3.5 jack to my transceiver.
After applying CAT settings in WSJT-X i have a hamlib error

>  read_string_generic called, rxmax=4095 direct=1, expected_len=1
> write_block(): TX 14 bytes, method=2
> 0000    46 41 30 30 30 31 34 30 37 34 30 35 35 3b           FA00014074055;  
> write_block(): TX 3 bytes, method=2
> 0000    49 44 3b                                            ID;             
> read_string_generic called, rxmax=51 direct=1, expected_len=1
> read_string_generic(): RX 30 characters, direct=1
> 0000    ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff     ................
> 0010    ff ff ff ff ff ff ff ff 49 44 30 32 30 3b           ........ID020;  
> kenwood_transaction: read_string(expected=51, len=30)='������������������������ID020;'
> kenwood_transaction: No data expected, checking ID; in ������������������������ID020;
> kenwood_transaction: wrong reply �� for command verification ID
> kenwood.c(624):kenwood_transaction returning2(-8) Protocol error
> 
> kenwood.c(1908):kenwood_set_freq returning2(-8) Protocol error
> 
> 0:rig.c(1814):rig_set_freq returning(-8) Protocol error
> 
> Protocol error
> Protocol error
>  while setting frequency
> 
> Timestamp: 2025-01-06T21:37:08.617Z

Test CAT button lights with green and Test PTT lights with red but PTT worked on transceiver.

mdblack98:
What version of WSJT-X?
What version of Hamlib?

brightkill:
> What version of WSJT-X? What version of Hamlib?

WSJT-X 2.7.0rc8
Hamlib 4.6~git

brightkill:
@mdblack98 now i have this bug while uSDX transmit at 28.074 MHz. i've built the git master last branch with this commit and seems to work fine

PR #1768, opened 2025-06-17
Un-break hamlib on TM-D710/TM-V71/etc

larsks:
This reverts commit d1e0e3f204d3b0f5b3a17e83a50d360184136e0d.

That commit introduced a `remove_nonprint` method that breaks hamlib on all
TM-D710/TM-V71A devices by erroneously removing the command termination
character.

Resolves: #1767 #1698

### DIFF
--- a/rigs/kenwood/kenwood.c
+++ b/rigs/kenwood/kenwood.c
@@ -231,6 +231,12 @@ struct confparams kenwood_cfg_params[] =
     { RIG_CONF_END, NULL, }
 };
 
+// This function removes non-printable characters from a buffer. This was
+// implemented to work around a problem reported with the uSDX transceiver
+// [1], which emulates the Kenwood TS-480 but apparently generates garbage
+// on the serial port in some situations.
+//
+// [1]: https://github.com/Hamlib/Hamlib/issues/1652
 static int remove_nonprint(char *s)
 {
     int i, j = 0;
@@ -443,9 +449,14 @@ transaction_read:
     rig_debug(RIG_DEBUG_TRACE, "%s: read_string len=%d '%s'\n", __func__,
               (int)strlen(buffer), buffer);
 
-    // this fixes the case when some corrupt data is returned
-    // let's us be a little more robust about funky serial data
-    remove_nonprint(buffer);
+    // This fixes the case when some corrupt data is returned; it lets us be a
+    // little more robust about funky serial data. If the terminator is
+    // printable(usually ';'), then there should be no nonprintables in the
+    // message; if it isn't (usually '\r') then don't touch the message.
+    if (isprint(caps->cmdtrm))
+    {
+        remove_nonprint(buffer);
+    }
 
     if (retval < 0)
     {


N0NB:
Rather than a complete reversion, would it be possible to not apply the `remove_nonprint` function to affected models or, better yet, make sure the command termination character is preserved?  Mike's commit references issue #1652 so it seems there is a reason Mike added this but may have inadvertently stripped the command terminator as well.

larsks:
> Rather than a complete reversion, would it be possible to not apply the remove_nonprint function to affected models or, better yet, make sure the command termination character is preserved?

Maybe! This revert will at least resolve the problem for people trying to use hamlib with impacted devices. We can certainly choose not to merge it, but now at least I can point people here who run into the same problem.

To be clear: at the moment, hamlib is now completely broken on a variety of Kenwood devices. I'm sure it's possible to create a more nuanced solution to #1652, but that will take time and given severe nature of the problem I wanted to make this fix available.

> Mike's commit references issue https://github.com/Hamlib/Hamlib/issues/1652 so it seems there is a reason Mike added this but may have inadvertently stripped the command terminator as well.

The change was made in the context of Kenwood devices that use `;` as the command terminator; it was never tested on devices that use `\r`.

GeoBaltz:
How about checking the expected command terminator(kenwood_priv_caps.cmdtrm) - if it is printable, remove the nonprintables. If it isn't printable, then don't.

N0NB:
George, that's a much more elegant solution than my example in #1767 which only tests for `\r` and is hard coded.

GeoBaltz:
Something like this -
``` diff
diff --git a/rigs/kenwood/kenwood.c b/rigs/kenwood/kenwood.c
index 1596c2dfe..88655f2dd 100644
--- a/rigs/kenwood/kenwood.c
+++ b/rigs/kenwood/kenwood.c
@@ -445,7 +445,13 @@ transaction_read:
 
     // this fixes the case when some corrupt data is returned
     // let's us be a little more robust about funky serial data
-    remove_nonprint(buffer);
+    // If the terminator is printable(usually ';'), then there should be
+    //   no nonprintables in the message; if it isn't(usually '\r') then
+    //   don't touch the message
+    if (isprint(caps->cmdtrm))
+    {
+        remove_nonprint(buffer);
+    }
 
     if (retval < 0)
     {

```

N0NB:
George, check my proposed fix in #1767.  The problem is outlined in #1652 and I think needs to be dealt with in the `remove_nonprint` function.

N0NB:
Then again, yours might be correct if radios that use `EOM_TH` (`\r`) need to pass unprintable characters to the host.

N0NB:
Lars, give George's patch preference over mine as it's likely correct in that radios that use `EOM_TH` should not have the message modified.

brightkill:
Since the previous commit works for some devices but breaks others, maybe instead of reverting it entirely, we could keep the old behavior behind a ./configure flag (e.g., --enable-removenonprint or similar).

This way, users who rely on the previous behavior can enable it explicitly, while the default remains compatible with the majority of devices.

N0NB:
Hi @brightkill 

I think George's submission will work for both cases.  Having to invoke `configure` time options for relatively common devices is not something we wish to do. This will be resolved in the library.

One question I have for you related to issue #1652 is whether the device firmware was updated by the authors to remove the unprintable characters?

larsks:
@GeoBaltz I've updated this PR to adopt your proposed solution.

N0NB:
Thanks, Lars and George.

This will be in tomorrow's daily snapshot.  I will also cherry-pick this over into the Hamlib-4.6.3 branch for a bug fix 4.6.4 release in a few weeks.

PR #1770, opened 2025-06-20
Fixed jrc_set_chan

markjfine:
set_chan() was correctly creating and sending the command, and returning RIG_OK. However, radio was actually ignoring it because command wasn't terminated with a CR. This is now corrected.


N0NB:
Thanks, Mark.

markjfine:
No problem. I may have done the original many moons ago, so I'm likely fixing my own bugs.

PR #1774, opened 2025-06-23
Correct powerstat check on Icom R75

markjfine:
The R75 for some reason rejects the powerstat query and returns an error. Commented out  .get_powerstat to correct that. Applications should initially assume it's on, then internally track power status, since you can still turn it off.

### DIFF
--- a/rigs/icom/icr75.c
+++ b/rigs/icom/icr75.c
@@ -245,7 +245,7 @@ struct rig_caps icr75_caps =
     .scan =  icom_scan,
     .set_ts =  icom_set_ts,
     .set_powerstat = icom_set_powerstat,
-    .get_powerstat = icom_get_powerstat,
+    //.get_powerstat = icom_get_powerstat,
 
     .set_channel = icr75_set_channel,
     .get_channel = icr75_get_channel,


N0NB:
So the radio rejects the powerstat query at any and all times?  If so, then no point in implementing the command.

N0NB:
Thanks, Mark.

markjfine:
@N0NB 
Also looking at AGC and will likely provide a separate issue/pr. Someone remapped all legacy agc sets/gets to the ic746 pro, making a small mess for the R75 (and likely others). The defines are located in icom-defs.h with a D_AGC_ prefix. Perhaps that should be a per-radio define rather than all icom radios?

Just looking for guidance before I go ahead and accidentally ruin someone's day.

Edit: i should note that I've tried redefining the D_ values back to normal AGC_ ones within icr75.c to no avail. I kinda expected this but gave it a shot.

N0NB:
Those defines have been in place a long, time.  Stephane, F8CFE, committed that back on April 14, 2010 as commit 16e6443.

We might have still been on CVS back then even though there is an SVN release link in the message.  Those were added when I converted the repository from SVN to Git.  Stephane had converted the repository on SourceForge from CVS to SVN a couple of years prior to my switching it to Git.

markjfine:
Ok. I'll do an issue to document it, with a pr that fixes it in icom.c explaining the rationale, and we'll see what falls out. lol.

PR #1783, opened 2025-06-25
More cppcheck messages handled

GeoBaltz:
Mostly minor warnings/errors/false positives

All for master, patch 2(TS-590) and 4 (rotctl dump_caps) could go in 4.6.4, but I doubt anyone would notice either way.


N0NB:
Thanks, George.

PR #1790, opened 2025-06-29
Allow to execute some tests with simulators or with real devices

dforsi:
This PR
* adds a way to pass the usual `--model` `--rig-file` `--serial-speed` arguments to the test runner to talk to a simulator or to a real rig (at this time the tests succeed with the dummy `--model 1` and fail with other drivers)
* changes the order of the arguments of some bindings to have the optional VFO at the end (more bindings need changes)
* makes some small changes to the dummy rig to make it more similar to other rigs (more changes may be needed, either to the dummy or to the simulators or to both)

The tests must be run manually to pass the arguments, for example:
* in a terminal run a simulator and note the path of the device file that it creates (eg. `/dev/pts/2`) that is the value to be used with the `--model` argument 
* in another terminal run the test script `test_rig.py` modifying PYTHONPATH if needed
  > PYTHONPATH=bindings/:bindings/.libs/ bindings/python/test_rig.py --model 1035 --rig-file /dev/pts/2

I think that this can be merged to master now because it works with the dummy rig, even if there are still things to finish:
* the tests `test_misc()` and `test_object_creation()` are executed only with the dummy rig and are skipped with other models because they would be a waste of time
* the order of the arguments of the bindings (and the order of the values that they return) is a problem to be solved
* the test would probably fail with most rigs/simulators, so it would need to 

The first 3 commits can go also to 4.6.x to avoid too many changes to be merged in 4.7.0, but they aren't too important; the 4th commit adds the functionality to run the tests; the commits after the 4th change the API of some bindings but more changes will be needed, but they only change the files in the `bindings` so it shouldn't be a problem if they are merged only to 4.7.0

### DIFF
--- a/bindings/python/conftest.py
+++ b/bindings/python/conftest.py
@@ -5,6 +5,11 @@ import sys
 
 def pytest_addoption(parser):
     # using long options only because short options conflict with pytest's
+    if sys.argv[1].endswith("amp.py"):
+        parser.addoption('--model', type=int, default=1,
+                        metavar='ID', help='select amplifier model number')
+        parser.addoption('--amp-file', default=None,
+                        metavar='DEVICE', help='set device of the amplifier to operate on')
     if sys.argv[1].endswith("rig.py"):
         parser.addoption('--model', type=int, default=1,
                         metavar='ID', help='select radio model number')
--- a/bindings/python/test_amp.py
+++ b/bindings/python/test_amp.py
@@ -9,14 +9,12 @@ import Hamlib
 
 Hamlib.rig_set_debug(Hamlib.RIG_DEBUG_NONE)
 
-AMP_MODEL = Hamlib.AMP_MODEL_DUMMY
-
 class TestClass:
     """Container class for tests"""
 
-    def test_without_open(self):
+    def test_without_open(self, model):
         """Call all the methods that do not depend on open()"""
-        amp = Hamlib.Amp(AMP_MODEL)
+        amp = Hamlib.Amp(model)
         assert amp is not None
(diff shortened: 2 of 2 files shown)


N0NB:
Is moving the VFO parameter to the end going to be well documented?  Also, how to run the tests should be documented.

dforsi:
Good questions, it doesn't make sense to maintain double documentation, double code in the bindings and in the tests to have the VFO as an optional last parameter and being different from the C library.

I added the optional VFO where it was missing because the *.swg files where written to support that, but I'm going to revert those changes to get the arguments with the same order as in the Hamlib library, keeping the hidden pointer to the object (Rig, Amp, Rot) like now

N0NB:
Thanks for your reconsideration, Daniele.

dforsi:
The main goal of this PR doesn't work so I'm going to split this so that some parts can be merged earlier

dforsi:
In this latest iteration, this PR allows to run the scripts `test_amp.py`, `test_rig.py`, `test_rot.py` from the command line passing to the pytest framework the parameters to select a model, a device and a baud rate.

By default those scripts open the dummy device and the tests succeed (they are the same tests run by the CI on github), however when running with different models some tests fail. Fixing the tests for simulators is a bigger task that can be done in steps and shouldn't block the reviews of this PR.

N0NB:
Thanks, Daniele.

dforsi:
This broke `make check` but we didn't notice because both I and the CI were running `make distcheck` which didn't have the files changed in this PR.

Proposed fix in PR  #1819.

PR #1792, opened 2025-06-30
Fix build failures when building bindings

dforsi:
This PR:
* fixes the issue reported by Walter Fey @dl8fcl in https://github.com/Hamlib/Hamlib/commit/f7d38f92cac73fb8b952f5dde3853f40ef910e14#commitcomment-161078321 by enabling the compilation of the Python code only for the Python bindings
* modifies the CI to build of all language bindings to avoid the same issue in future


N0NB:
Thanks for fixing this, Daniele.

PR #1795, opened 2025-07-05
Fix disabling rigs/winradio and rotators/indi and modify CI to build minimal or (mostly) complete

### LINKED ISSUE #1737: rigs/winradio and rotators/indi are compiled even if disabled in configure
After running `./configure --disable-winradio` or `./configure ---without-indi` the makefile still compiles them.

winradio is compiled because it is unconditionally added to `RIG_BACKEND_LIST` in `configure.ac`.

indi is compiled because it is unconditionally added to `DIST_SUBDIRS` in the top `Makefile.am`.

dforsi:
This fixes #1737 fixing the use of `--enable-winradio=yes/no` and `--with-indi=yes/no`

The main issue was that dependencies where created for both rigs/winradio and rotators/indi even when they where configured to not build, so the makefiles were building them in all cases, so I added new variables to hold the paths for the optional drivers, to exclude them from `make all` and similar targets but to include them in `make dist` and similar.

To verify that everything works as expected I modified the CI action:
* remove the simple `make check` because `make distcheck` does the same and more, but it only builds with the default options set
* repeat the configure+make steps setting all options to no and to yes, with some exceptions:
  * shared and static libraries are build in all cases
  * usrp is not built because there isn't a package in Ubuntu to be installed with apt (it needs to be checked out from a repository, but it's a different issue)
  * Python bindings fail to build in macOS so they are disabled for now
  * there is an inconsistency in the final output of `configure` where if pytest is installed it prints `Enable Python tests yes` even if the Python binding aren't built and the tests will not run
* reduce the clutter in the output on the CI (errors and warnings are still shown)

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -76,14 +76,12 @@ jobs:
       run: grep README Makefile.am
     - name: configure
       if: runner.os != 'macOS'
-      run: ./configure ${{ matrix.configure_args }}
+      run: ./configure ${{ matrix.configure_args }} --enable-silent-rules
     - name: configure on macOS
       if: runner.os == 'macOS'
-      run: ./configure ${{ matrix.configure_args }} --without-python-binding
+      run: ./configure ${{ matrix.configure_args }} --enable-silent-rules --without-python-binding
     - name: make
-      run: make -j 4
-    - name: make check
-      run: make check
+      run: make -j 4 V=0 --no-print-directory
     - name: make distcheck
-      run: make distcheck
+      run: make distcheck V=0 --no-print-directory
 
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -12,6 +12,41 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
+        configure_args:
+          - >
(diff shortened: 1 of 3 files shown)


dforsi:
I don't know how to fix the errors in the CI when using `--with-lua-binding`, after `brew install lua` on Mac:

```
checking for lua.h... no
checking for lualib.h... no
checking for lauxlib.h... no
checking for luaconf.h... no
```

dforsi:
I'm setting this PR as draft because sorting the as in commit [d065bcb](https://github.com/Hamlib/Hamlib/pull/1795/commits/d065bcb8ef0311ab21f85c8a7d1aec83db2c6cb1) breaks the build with:
>  rig.c:74:10: fatal error: 'hamlibdatetime.h' file not found

Like in the job https://github.com/Hamlib/Hamlib/actions/runs/16087329208/job/45400068817?pr=1795

I can't reproduce that on my local build.

dforsi:
I'm setting this PR ready for review now because the two things that I'm not able to fix (not installing Lua development on Mac and duplication of targets in top level `Makefile.am`) existed even before these changes.

N0NB:
It looks clean enough to merge to me.

N0NB:
Thanks, Daniele.

PR #1797, opened 2025-07-06
Revert commit ae95564 anche change the test to suit

dforsi:
This reverts commit https://github.com/Hamlib/Hamlib/commit/ae9556462a2989b8b97af0d6e320558ffa0e1f3e.

The change for ATTRIBUTE_FORMAT_PRINTF was good but I'm not re-adding it because it generates a duplicate definition warning when building the INDI driver because is is defined also in /usr/include/libindi/indidevapi.h with this different code:
```
/* enable warnings for printf-style functions */
#ifndef ATTRIBUTE_FORMAT_PRINTF
# ifdef __GNUC__
#  define ATTRIBUTE_FORMAT_PRINTF(A, B) __attribute__((format(printf, (A), (B))))
# else
#  define ATTRIBUTE_FORMAT_PRINTF(A, B)
# endif
#endif
```

### DIFF
--- a/bindings/python/test_startup.py
+++ b/bindings/python/test_startup.py
@@ -31,7 +31,7 @@ class TestClass:
         assert my_rig.get_conf("retry") == '5'
 
         assert my_rig.error_status == Hamlib.RIG_OK
-        assert Hamlib.rigerror(my_rig.error_status) == "Command completed successfully\n"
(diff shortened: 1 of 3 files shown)


N0NB:
Thanks, Daniele.

PR #1799, opened 2025-07-07
AF6SA WRC Rotor added

aa5sh:
I renamed it from my testing m.  I should have before I published.  Is there some information of what I should do/test before uploading?  I do successful make's and test with my application. But looks like it is failing when it does some tests on GitHub.  Is there something I can run on my end (MacOS) before I submit?


dforsi:
In this particular case you just need to update a test that is failing because at least a new define appeared, which is expected because you added a new backend:

>  E         +  'ROT_MODEL_AF6SA_WRC',

The code to update that test needs the Python bindings, so if can build them you can update that test yourself with `make -C bindings generate-pytests` and add it to this PR (if you can't, then we can find a workaround, but it's easier if you can).

The `generate-pytests` target is untested on Mac because I don't know what packages should be installed with brew; it would be great if you could tell what needs to be installed,  the only suggestion that I found is `brew install python` but it gives this error:
```
configure: error: 
  Could not link test program to Python. Maybe the main Python library has been
  installed in some non-standard library path. If so, pass it to configure,
  via the LIBS environment variable.
  Example: ./configure LIBS="-L/usr/non-standard-path/python/lib"
  ============================================================================
   ERROR!
   You probably have to install the development version of the Python package
   for your distribution.  The exact name of this package varies among them.
  ============================================================================
```

aa5sh:
Ok I think I got it to run finally.  Here are the steps I used for future reference:
```
brew install python@3.11
brew install swig

make distclean
./bootstrap
./configure PYTHON=python3 \
  CPPFLAGS="$(python3-config --includes)" \
  LDFLAGS="$(python3-config --ldflags)" \
  --with-python-binding

make -C bindings generate-pytests PYTHON=/opt/homebrew/opt/python@3.11/libexec/bin/python3

```

aa5sh:
Not sure why it made me create a new PR #1800

Issue #1746, opened 2025-05-26
Packages like `autoconf-archive` may cause `bootstrap` and `configure` issues


N0NB:
In reference to commit e641162, the installation of a `autoconf-archive` package (Debian name or similar) may cause Autconf macro (`macros/m4`) files to be overwritten when `bootstrap` is run.  The reason is due to there being some rather old macro files in our archive and the default behavior of `aclocal` (an Autoconf utility) is to replace them with newer versions it finds on the system.

### Why do we have old macro files?

They serve our purpose and we've not seen a reason to update.  This attitude may need to be adjusted.

### What are the symptoms?

Quite a number of files will be shown to be updated during a `bootstrap` run (Debian Stable):

``` console
$ ../hamlib/bootstrap 
Running 'autoreconf -i' to process configure.ac
and generate the configure script.
aclocal: overwriting 'macros/ax_cflags_warn_all.m4' with '/usr/share/aclocal/ax_cflags_warn_all.m4'
aclocal: installing 'macros/ax_compiler_vendor.m4' from '/usr/share/aclocal/ax_compiler_vendor.m4'
aclocal: overwriting 'macros/ax_cxx_compile_stdcxx.m4' with '/usr/share/aclocal/ax_cxx_compile_stdcxx.m4'
aclocal: overwriting 'macros/ax_lib_readline.m4' with '/usr/share/aclocal/ax_lib_readline.m4'
aclocal: overwriting 'macros/ax_pkg_swig.m4' with '/usr/share/aclocal/ax_pkg_swig.m4'
aclocal: installing 'macros/ax_prepend_flag.m4' from '/usr/share/aclocal/ax_prepend_flag.m4'
aclocal: overwriting 'macros/ax_pthread.m4' with '/usr/share/aclocal/ax_pthread.m4'
aclocal: overwriting 'macros/ax_python_devel.m4' with '/usr/share/aclocal/ax_python_devel.m4'
aclocal: installing 'macros/ax_require_defined.m4' from '/usr/share/aclocal/ax_require_defined.m4'
aclocal: installing 'macros/libtool.m4' from '/usr/share/aclocal/libtool.m4'
aclocal: installing 'macros/ltoptions.m4' from '/usr/share/aclocal/ltoptions.m4'
aclocal: installing 'macros/ltsugar.m4' from '/usr/share/aclocal/ltsugar.m4'
aclocal: installing 'macros/ltversion.m4' from '/usr/share/aclocal/ltversion.m4'
[...]

PR #1800, opened 2025-07-07
AF6SA WRC with Python Tests

aa5sh:


aa5sh:
Hey @dforsi thanks for your help.  I did have to make one change in generate_tests.py from " to ' on line 60 I think for it to run on the Mac.  I will try to step through the process at home on my other laptop because I did a lot of back and forth.  But I think I just had to install those two packages in homebrew python@3.11 and swig .

bindings/python/generate_tests.py
```diff
@@ -57,7 +57,7 @@ def generate_file(object_name, callables, properties):
 """
 import Hamlib
 '''
-    filename = f"test_{object_name.replace(".", "_")}_class.py"
```
dforsi, line 60: Your change looks good; the reason is that you are using Python 3.11 while I'm using 3.13 and my syntax is [allowed since 3.12](https://docs.python.org/3.12/whatsnew/3.12.html#pep-701-syntactic-formalization-of-f-strings)

rotators/gs232a/gs232.c
```diff
@@ -141,6 +141,33 @@ gs232_wo_transaction(ROT *rot, const char *cmdstr,
...
+
+    SNPRINTF(cmdstr, sizeof(cmdstr), "W%03u" EOM, u_az);
+    retval = gs232_wo_transaction(rot, cmdstr, NULL, 0);
+
+    if (retval != RIG_OK)
+    {
+        return retval;
+    }
+
+    return RIG_OK;
```
dforsi, line 169: Nitpicking: if you aren't doing anything in the `if`, remove it and do just:
```
return retval;
```

dforsi (commented):
You are modifying the m4 macros, but it seems unnecessary for your other changes and since the macros have caused problems in the past, it's better to leave them alone.

bindings/python/generate_tests.py
N0NB: The m4 macros can be unintentionally updated in situations like issue #1746 documents.

N0NB:
Before merging into `master`, it looks like I could drop commits 9fe9a00 bcb4406 and 8b7c803.  Does that look acceptable?

aa5sh:
Yes I believe so.  Those were trying to cleanup the macros.  Thank you.

N0NB:
Thanks, Michael.

N0NB:
I missed this before merging but I am seeing this GCC warning:

``` console
Making all in rotators/gs232a
make[2]: Entering directory '/home/nate/git/hlb/hamlib-4.7~git/_build/sub/rotators/gs232a'
  CC       gs232a.lo
  CC       gs232b.lo
  CC       gs232.lo
../../../../rotators/gs232a/gs232.c: In function 'wrc_rot_set_position':
../../../../rotators/gs232a/gs232.c:149:20: warning: variable 'u_el' set but not used [-Wunused-but-set-variable]
  149 |     unsigned u_az, u_el;
      |                    ^~~~
  CCLD     libhamlib-gs232a.la
make[2]: Leaving directory '/home/nate/git/hlb/hamlib-4.7~git/_build/sub/rotators/gs232a'
```

N0NB:
Thanks, Michael.

PR #1801, opened 2025-07-11
Add alpha FTX-1 radio support to Hamlib

jeremybox:
- Add FTX-1 rig model (ID 840) to riglist.h
- Create ftx1.c and ftx1.h with FTX-1 implementation
- Add FTX-1 to Yaesu Makefile.am
- Integrate FTX-1 into newcat.c command validation system
- Update yaesu.c and yaesu.h for FTX-1 support
- Based on FT-991A CAT protocol compatibility, but a review will happen when yaesu releases a valid CAT manual


N0NB:
Thanks, Jeremy.

Issue #1773, opened 2025-06-22
5.0.0 release goals


N0NB:
A list of feature enhancements for the 5.0.0 release:

- [ ] Set the minimum C version to C11 or C17 (C11 fixes).  ![0% done](https://markdone.org/progress/0.png) (@GeoBaltz)
- [ ] Split include/hamlib/rig.h so including rig.h gives the app access to the whole API, but none of the structures - and guarantee this stable. Allows app developers to not worry about Hamlib internal changes unless they explicitly need/include the structure definitions.. ![0% done](https://markdone.org/progress/0.png) (@GeoBaltz)
- [ ] Follow up on the cache move by breaking out the port and state structures.  Now I'm not sure we can make things transparent across 4.7<->5.0 - having cache inside state inside rig throws a monkey wrench into what I had hoped to do. ![0% done](https://markdone.org/progress/0.png) (@GeoBaltz)
- [ ] Make the tests that work with the dummy devices also work with the simulators. ![0% done](https://markdone.org/progress/0.png) (@dforsi)
- [ ] Add `dms2dmmm` and `dmmm2dms` to `locator.c` and `rotctl` using the `F` and `f` commands. ![0% done](https://markdone.org/progress/0.png) (@N0NB)
- [ ] Update values such as `int` that may be subject size limitations with `int32_t` or `uint32_t` as needed.
- [ ] Complete making structures pointers as discussed in #487.  ![0% done](https://markdone.org/progress/0.png) (@GeoBaltz)
- [ ] Remove rig_get_conf(), rot_get_conf() and amp_get_conf().  Already deprecated per Issue #924.  (@GeoBaltz)

N0NB:
Move `locator.c` changes here pending API discussion.

dforsi:
1. `src/token.h` should be part of the public API because functions an structures in `include/hamlib/*.h` use `token_t`

1. Several enumerations are preprocessor defines (`TOK_*`, `RIG_VFO_*`, `RIG_ANT_*`, maybe more) which doesn't allow the compiler to issue useful warnings

1. the arguments of the bindings are reversed wrt the C library, eg.
    bindings: `Rig_get_ant(Rig *self, ant_t *ant_rx, ant_t *ant_tx, ant_t *ant_curr, signed int *option, ant_t ant, vfo_t vfo)`
    C: `rig_get_ant(self->rig, vfo, ant, &value, ant_curr, ant_tx, ant_rx);`

[...]

Issue #1772, opened 2025-06-22
4.7.0 release goals


N0NB:
A list of feature enhancements for the 4.7.0 release

- [ ] Better coverage of bindings for Python and other languages.  ![0% done](https://markdone.org/progress/0.png) (@dforsi)
- [ ] Make P_THREADS a hard requirement. ![0% done](https://markdone.org/progress/0.png) (@GeoBaltz)

GeoBaltz:
For paving the road to 5.0:

1. create dummy/placeholder files for

- include/hamlib/port.h
- include/hamlib/rig_state.h
- include/hamlib/amp_state.h
- include/hamlib/rot_state.h

2. Fix the infrastructure to track and install these
3. Add  the required \#includes in hamlib source files
4. Move the struct definitions from rig.h to the dummy files, replaced with includes of them.  All these includes are marked for easy removal, possibly even ifdef'ed for testing.
5. Document, document, document

Should be possible with no effect on current apps, but allows them to any necessary changes with full source compatibility 4.7<->5.0
Doing the prep work for 4.7 should make things at least a little bit easier for 5.0

N0NB:
Move the `locator.c` work to 5.0.0 for more discussion and API change.

GeoBaltz:
I was looking through the doxygen output tonight, and came across the "Deprecated list".  There are some items(rig_get_trn, rig_set_trn and rig_caps::transceive) there that haven't been addressed. They could be marked in 4.7, and deleted in 5.0.

OK?

N0NB:
Everything has been rather quiet of late and that is probably as much my fault as anyone else's.  Are we ready to create the 4.7 branch?

As I see it, this will likely be something of a long term support branch considering the changes planned for the 5.0 branch.  I don't want to hold things up any longer as I'm sure work on 5.0 is ready to commence.

I suspect that we will be able to cherry-pick backend code from 4.7 to master (and vice versa) for some time.  I'm considering relaxing the rules a bit and permit new models in the 4.7 branch given that it may be some time before a 5.0.0 release (thoughts requested).  I'm anticipating perhaps as many as 10 releases from the 4.7 branch and there may be some releases after 5.0.0 is released given 4.7's coming legacy API status (thoughts requested).

I'm thinking that branching now may allow for a 4.7.0 release yet this year.

GeoBaltz:
[...]

PR #1802, opened 2025-07-11
Move port and state structures to their own includes

GeoBaltz:
Prep for 5.0 - allow adding includes for data access in 4.7, when it will be required in 5.0.

Issue #1772. #1773


N0NB:
Thanks, George.

PR #1808, opened 2025-07-20
Cleanup includes

dforsi:
This PR modifies:
* `.gitignore` sorting it to make it easier to see what's missing and to ignore more files (mostly for Windows builds)
* `configure.ac` to delete unused tests
* many include directives following the findings of IWYU (which means adding or removing an `#include` or swapping one file for another)
* the delimiters `< >'` or `" "` around filenames

There are still many include that don't make IWYU happy, but the main goal was to remove the includes that were obviously wrong (like `math.h` when you're not doing math) and then doing some cleanup.
One thing to consider in future is that IWYU suggests to include `serial.h` when the code is only opening/closing a device and `iofunc.h` when the code is using the device for reading and/or writing, but it can be configured to always suggest `serial.h` instead (the gain is unclear in either case).

Changing the quoting around filenames makes no difference when using gcc, ,but it make it clear when we want to use the includes coming with Hamlib and when we do not"


N0NB:
For commit 9b21dee it can probably be argued that any of the included files from `include/hamlib` can be surrounded by angle brackets rather than quotes.  For client applications compiling against an installed version of Hamlib, angle brackets should be used.

I think our build system ensures that the local includes are consumed rather than any installed in the filesystem so either way is probably fine for our purposes.

GeoBaltz:
The includes in hamlib/rig.h are going to be included by applications - they need <> since they will be using the installed hamlib-dev (or whatever it's called by the distro) /usr/include/hamlib.  I wasn't happy about the <> usage, but we only control the usage in our code - once it is packaged and installed, it is system code.
Besides, they will only be in rig.h until 5.0.

N0NB:
That should also extend to `rotator.h` and `amplifier.h` as well since they're essentially siblings of `rig.h`, right?

dforsi:
You are right.
I reverted the changes to the files in `include/hamlib/'

N0NB:
Thanks Daniele.

PR #1815, opened 2025-07-27
Make the dummy rig more similar to other rigs

dforsi:
This PR adds to the dummy rig:
* antenna number 5
* a check for the split value and a fixed test (I mistakenly thought that the `split` argument was a frequency offset, instead of an on/off switch)
* a define instead of a plain number to mean "no antenna"

These changes were initially in #1790 with the difference that now I set chan->ant to RIG_ANT_NONE instead of RIG_ANT_1.


N0NB:
Thanks, Daniel.

PR #1816, opened 2025-07-27
Fix bindings and use more constants in pytest

dforsi:
I started these changes with the idea of making all Python bindings accept an optional VFO argument with the default of `RIG_VOF_CURR` because the latest changes seemed to go that way, then I have seen that the order of arguments has changed over time in both directions, so I'm reverting some commits, instead of deleting them because they worked, but not necessarily in the more useful way, since the order of the arguments is swapped with respect to the underlying  C functions.

Changing the order of the arguments, if it's done, is better done in a specific pull request that allows easier checking of the changes.

In the end, the major change is to make the order of the argument of `set_dcs_sql()` the same as in the 3 similar methods `set_ctcss_sql() set_ctcss_tone() set_dcs_code()`, which I considered a bug.

These changes were initially in #1790.


N0NB:
I see this PR now shows conflicts that need resolving.  Quite possibly this is a side effect of merging #1790 and #1815.

N0NB:
Thanks, Daniele.

PR #1817, opened 2025-07-27
Fix make check

dforsi:
This PR changes `Makefile.am` to stop `make check` if an error happens when running the Tcl or the Lua tests (Perl and Python already stopped).

The failures in running the Tcl tests went unnoticed, such as in [a recent CI run](https://github.com/Hamlib/Hamlib/actions/runs/16550365122/job/46804305394?pr=1790#step:11:1644):
```
 couldn't load file "./.libs/hamlibtcl-4.7": ./.libs/hamlibtcl-4.7: cannot open shared object file: No such file or directory
    while executing
"load ./.libs/hamlibtcl-4.7 Hamlib"
    ("package ifneeded hamlib 4.7" script)
    invoked from within
"package require hamlib"
    (file "./tcltest.tcl" line 12)
Tcl test failed
```

The same could happen for Lua tests.


N0NB:
Thanks, Daniele.

PR #1818, opened 2025-07-27
Deprecate rig_get_conf, clean up switch fallthroughs

GeoBaltz:
Add macros for annotating fall through and deprecation, if the compiler is capable.
Repair or annotate all fallthroughs in `switch` statements.
Deprecate `rig_get_conf()` (from NEWS file)
Fix internal uses of `rig_get_conf()`

### DIFF
--- a/c++/rigclass.cc
+++ b/c++/rigclass.cc
@@ -87,11 +87,19 @@ void Rig::setConf(const char *name, const char *val)
 
 void Rig::getConf(hamlib_token_t token, char *val)
 {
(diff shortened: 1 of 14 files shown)


N0NB:
I've been wondering if we should update `uthash.h`.  It's quite old and there has been quite a bit of development on it since.

N0NB:
I do see the deprecation warnings when I mimic the CI locally and compile with all the bindings enabled with `distcheck`, so that is working.

Thanks, George.

Issue #924, opened 2022-01-09
get_conf function replace with get_conf2


mdblack98:
get_conf functions
rig_get_conf
rot_get_conf
amp_get_conf
Will all be deprecated in Hamlib 5.0 and replaced with get_conf2 versions which pass in the length of the the value.
This is to avoid buffer overruns
The old get_conf will produce a compilation warning about the upcoming change.

N0NB:
@GeoBaltz With yesterday's commit, can this be closed?

GeoBaltz:
Probably not.  They are still there, and won't be officially dead until 5.0.

(They're only pining for the fjords)

GeoBaltz:
Looking at this again, this only addresses the deprecation, so it is done.  Can be closed.

Added a comment to 5.0 goals so we don't forget.

N0NB:
Thanks, George.

PR #1821, opened 2025-07-29
Deprecate rot_get_conf and amp_get_conf, too

GeoBaltz:
Mark as deprecated - to be removed in 5.0 - Issue #924 
More cppcheck tweaks


N0NB:
Thanks, George.

Issue #1013, opened 2022-04-21
Rig model name changes


mdblack98:
```
  1004  Yaesu                  MARK-V FT-1000MP        20211014.0      Stable      RIG_MODEL_FT1000MPMKV
  1024  Yaesu                  FT-1000MP               20211113.0      Stable      RIG_MODEL_FT1000MP
  1025  Yaesu                  MARK-V Field FT-1000MP  20211014.0      Stable      RIG_MODEL_FT1000MPMKVFLD
```

Change models to
```
1004 FT-1000MP MARK-V
1025 FT-1000MP MARK-V Field
```

Review other models for similar inconsistences.

arekm:
Just a note that WSJT-X remembers rigs in its config not by rig number but by name:

WSJT-X.ini
````
Rig=FLRig
````

and after hamlib upgrade it gets confused and defaults to first on the list (ADAT). Unfortunately it does that silently.

To get things working it requires manual change in WSJT-X settings. Was just debugging why things work with hamlib 4.6.x and fail with git master.

````
Rig=W1HKJ FLRig
````

Not a hamlib problem but maybe worth giving a big note in NEWS about rig name changes and potential effects.

N0NB:
I just reverted this in commit 5b9ce8c where it will be FLRig for both the model and manufacturer.

arekm:
From wsjtx point of view that's just another change as previously it was "FLRig" and now "FLRig FLRig".

IMO the changes were fine on hamlib side just documenting that well was missing. And maybe mentioning to never rely on names but use IDs instead.

N0NB:
Trouble is that the WSJT-X maintainer wants it changed as setting the manufacturer name to W1HKJ put it at the bottom of the list.

arekm:
Nice addition to wsjt-x would be adding simple search bar to that setting. That would make order not that important.

Manually reviewing that list was always painful anyway.

N0NB:
People must change radios more often than I do.  I used that list once around 10 years or more ago and haven't looked at it since.  This is why I was puzzled that such a simple change became such a big deal!

arekm:
Not that people change radios but change software like use wsjtx with stable 4.6 hamlib then try newer wsjtx built with hamlib 4.7 devel version and so on (rig name changes in hamlib -> wsjtx silently switches to ADAT rig).

[...]

PR #1823, opened 2025-08-01
Sanitize radio model names and manufacturers

N0NB:
Reference GitHub issue #1013.


N0NB:
Also mentioned in https://sourceforge.net/p/hamlib/discussion/25919/thread/9859a3e545/#6f35

PR #1824, opened 2025-08-01
Various fixes and cleanups

dforsi:
This PR fixed two error messages in simulators and  removes some unused/unneeded code from some rigs and some simulators.

The functions defined with `DECLARE_PROBERIG_BACKEND()` are used by `rig_probe()` and are useful only if they actually check if they can handle the rig connected to the given port, so register the rig with the macro that doesn't declare the probe function..


brianjester (approved):

N0NB:
Thanks, Daniele.

PR #1827, opened 2025-08-03
Fix typos

dforsi:
This PR fixes some small errors (a couple of them are in user-visible strings, most are in comments, some are just adding a space between a number and its unit of measure).


N0NB:
Thanks, Daniele.

PR #1828, opened 2025-08-03
Fix initialization of comm_status

dforsi:
This PR fixes a problem in the lifecycle of a connection tracked with `rig->comm_status` that made appear that a connection was in progress even before `rig_open()` was called. It also updates the test which showed this issue.

The function `rig_init()` doesn't touch the hardware, it's `rig_open()` that initiates the connection and in fact the first use of comm_status in `rig_open()` is when it gets the value `RIG_COMM_STATUS_CONNECTING` and `rig_close()` sets it to `RIG_COMM_STATUS_DISCONNECTED`.

### DIFF
--- a/bindings/python/test_rig.py
+++ b/bindings/python/test_rig.py
@@ -47,10 +47,12 @@ class TestClass:
         assert rig is not None
 
         assert rig.state.comm_state == 0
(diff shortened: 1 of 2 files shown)


N0NB:
Thanks, Daniele.

PR #1826, opened 2025-08-02
Enhanced FTX-1 support with comprehensive improvements

jeremybox:
- Added RIT/XIT functionality with proper command handling
- Enhanced split operation support with get/set functions
- Improved error handling with dedicated error processing
- Updated VFO operations to include memory clear (MCL)
- Enhanced filter definitions for better mode support
- Added configuration parameters framework
- Improved frequency range and band handling
- Added proper format specifiers for shortfreq_t values
- Integrated with newcat backend for memory operations
- All commands tested and working with real FTX-1 hardware

PR #1830, opened 2025-08-03
Yaesu FTX1 Changes

aa5sh:
When I tried to connect rigctld to wsjtx for this rig I could get freq but not set.  Had to adjust the IF response for 30.  Then it would cause weird hangs which was resolved by increasing the FTX1_POST_WRITE_DELAY from 2 to 5.  Now seems to work fine with initial testing in WSJT and rigctld.  Also the band mapping was a little different for 2m 70cm.

### DIFF
--- a/rigs/yaesu/ftx1.h
+++ b/rigs/yaesu/ftx1.h
@@ -169,7 +169,7 @@
 
 /* Delay sequential fast writes */
 
-#define FTX1_POST_WRITE_DELAY               2
+#define FTX1_POST_WRITE_DELAY               5
 
 typedef struct
 {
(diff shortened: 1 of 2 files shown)


N0NB:
PR #1826 is also pending.  Please check those commits to avoid duplication.

aa5sh:
Sorry I didn't look at that one.  I will pull it down and close mine if needed.  Thanks @N0NB

N0NB:
Obviously both of you are intent on adding support.  Please collaborate if you can.  Unfortunately, I don't know if GitHub will allow multiple contributors adding commits to a given PR.

PR #1834, opened 2025-08-06
Another round of misc fixes

GeoBaltz:
More cppcheck stuff - suppress bogus uninitialized variables errors
Update CW buffer size
Some error case fixes found by -fanalyzer

All go in master, first and last (CW message and buffer overrun) could be picked for 4.6.x.


N0NB:
Thanks, George.

PR #1835, opened 2025-08-08
Improve Doxygen comments

dforsi:
Adds missing cross-references. Fix tagging of comment blocks in "aor" even if the rigs aren't included in the generated documentation.


N0NB:
Perhaps one day the main manual will be completed and we can assembled a "backend" manual.  That'll be a herculean task.

Thanks, Daniele.

dforsi:
That would be big for sure.

BTW I missed 1 cross reference for callbacks... I will add it to the next related PR.

PR #1836, opened 2025-08-08
Feature python callbacks

### LINKED ISSUE #481: freq callback in python
I would like to use the "freq_event" Callback with python. However, I
struggle getting this to work. That's how far I get:

def f():
    print("test")
    return 1

my_rig = Hamlib.Rig(Hamlib.RIG_MODEL_DUMMY)
my_rig.rig.callbacks.freq_event = f

However, I tried many things for f with ctypes but I'm not able to pass
this line (I have also tried f without ctypes, different set of
parameters, and so on) . That's the error message:

TypeError: in method 'rig_callbacks_freq_event_set', argument 2 of type
'freq_cb_t'

I have digged into the files Hamlib.py and hamlibpy_wrap.c and some SWIG
files but I can't get it working.

Any help would be appreciated.

Regards
Alex

dforsi:
This PR implements all the event callbacks in the Python bindings and in the Dummy rig:

* rig_set_freq_callback() (fixes #481)
* rig_set_dcd_callback() (NOT IN Dummy)
* rig_set_mode_callback()
* rig_set_vfo_callback()
* rig_set_pltune_callback() (NOT IN Dummy)
* rig_set_ptt_callback() (needed a fix in dummy.c)
* rig_set_spectrum_callback() (NOT IN Dummy)

In `dummy.c` I'm invoking the callbacks directly instead of using hte following functions that depend on `HAVE_PTHREAD`, while the bindings do not depend on it:
```
src/event.h:30:int rig_fire_freq_event(RIG *rig, vfo_t vfo, freq_t freq);
src/event.h:31:int rig_fire_mode_event(RIG *rig, vfo_t vfo, rmode_t mode, pbwidth_t width);
src/event.h:32:int rig_fire_vfo_event(RIG *rig, vfo_t vfo);
src/event.h:33:int rig_fire_ptt_event(RIG *rig, vfo_t vfo, ptt_t ptt);
```


dforsi:
This PR is adding all the code needed to set all the event callbacks from the Python bindings and execute Python code when an event is triggered; the code that sets the callbacks is executed for all 7 types, however only 3 callbacks are tested:
* rig_set_freq_callback()
* rig_set_mode_callback()
* rig_set_vfo_callback()
* rig_set_ptt_callback()

The other 3 are written in a similar way, so they should work, but are untested because testing them would require either a simulator or a rig that supports the needed features.
So I'm stopping here.

N0NB:
Thanks, Daniele.

PR #1837, opened 2025-08-09
Fix setting antenna number 3 for Yaesu FTDX3000

dforsi:
The argument ant_t ant is a bit mask, not a scalar, so the correct decimal value for choosing antenna 3 would be 4, not 3.
 Also use the defines instead of magic numbers.

### DIFF
--- a/rigs/yaesu/ft3000.c
+++ b/rigs/yaesu/ft3000.c
@@ -155,20 +155,21 @@ static int ft3000_set_ant(RIG *rig, vfo_t vfo, ant_t ant, value_t option)
 
     switch (ant)
     {
-    case 1:
+    case RIG_ANT_1:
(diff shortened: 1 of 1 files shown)


N0NB:
Good catch.  Given that `ant` is of type `ant_t`, this is the proper fix.

I wonder how many other magic numbers we'll find...

N0NB:
Scroll further down and fix the magic numbers in `ft3000_get_ant()`.  It looks like the same numbers were used, but maybe I'm mistaken.

N0NB:
As I read it again, it's likely the radio returns decimal 1, 2, or 3, and the code correctly assigns it to and `ant_t` value.

Never mind my noise...

N0NB:
Thanks, Daniele.

PR #1841, opened 2025-08-11
Fix deprecation warnings when compiling the bindings

dforsi:
This PR:
* replaces the deprecated `*_get_conf()` calls with the corresponding `*_get_conf2()` in the glue code
* it doesn't change the names of the Python methods
* hides the deprecated `cache` member of `rig_state`
* fixes an unrelated typo in a comment

I'm not sure about not having the methods with "2" in Python, it seems that they have never been available in Python but they aren't explicitly ignored.
In both cases the allocation of the string buffer isn't visible from Python, so the signature of the method doesn't change.

### DIFF
--- a/bindings/python/test_amp.py
+++ b/bindings/python/test_amp.py
@@ -27,11 +27,15 @@ class TestClass:
         assert amp.set_conf("", "") is None
         assert amp.get_conf("") == ""
         assert amp.get_conf(0) == ""
-        conf = amp.get_conf("mcfg")
-        assert isinstance(conf, str)
-        assert amp.set_conf("mcfg", "foo") is None
-        conf = amp.get_conf("mcfg")
-        assert conf == ""  # FIXME: should return "foo"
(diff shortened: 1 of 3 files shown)


GeoBaltz:
You want sizeof s, not sizeof *s.

dforsi:
Thank you George you are right, I fixed that.
The tests didn't catch my mistake because the dummy drivers do not implement the `*_conf2()` versions so the code falls back to the `*_conf()` versions so my length of 1 was ignored and the 3 chars string was returned (now I changed it to be more than 4 just in case the sizeof of a pointer gets used by mistake).

There are 99 drivers that set `.getconf` and 14 that set `get_conf2`:
```
grep -lr "\.get_conf " amplifiers/ rigs/ rotators/ | wc -l
99
grep -lr "\.get_conf2 " amplifiers/ rigs/ rotators/ | wc -l
14
```

N0NB:
Thanks, Daniele.

PR #1842, opened 2025-08-12
Fix segfault when executing the dump_caps command

dforsi:
Need to convert the level number 0..63 to the level bit mask.

The segfault happened when trying to print a string in `rig_sprintf_parm_gran()` because the data wasn't initialized properly in `netrigctl_open()` which is 676 lines long!

Steps to reproduce
```
tests/rigctld -m 1 &
tests/rigctl -m 2 dump_caps
```


N0NB:
I have confirmed the Segmentation Fault is present in 4.6.4 and likely earlier.  I'll be cherry picking this for 4.6.5.

Thanks much, Daniele.

PR #1845, opened 2025-08-13
Fixes and improvements for rigctl

dforsi:
This PR fixes some segfaults and adds the handling of `set_parm KEYERTYPE ?` in rigctl.

The dummy driver initializes its parameters to zero:
> memset(priv->parms, 0, RIG_SETTING_MAX * sizeof(value_t));

which crashed the code that tried to use the zero as a string pointer.

I'm confused about the fact that RIG_PARM_KEYERTYPE is a string `STRAIGHT BUG PADDLE` but `rigctl_parse.c` converts the string back and forth to an integer that is stored as ASCII string `"0", "1", "2"`. I didn't find any rig that uses a `RIG_KEYERTYPE_*` value, so I wonder if the code was finished, because some lines that instead stored an integer that were disabled with `#if 0`  and I had to disable identical code a few lines below.
(lines 3784-3790 and 3814-3820)
```
#if 0
            if (parm == RIG_PARM_KEYERTYPE)
            {
                val.i = atoi(arg2);
            }
            else
#endif
```

### DIFF
--- a/rigs/dummy/dummy.c
+++ b/rigs/dummy/dummy.c
@@ -1701,7 +1701,7 @@ static int dummy_set_parm(RIG *rig, setting_t parm, value_t val)
 
     if (RIG_PARM_IS_STRING(parm))
     {
-        strcpy(pstr, val.cs);
+        SNPRINTF(pstr, sizeof(pstr), "%s", val.cs);
     }
     else
     {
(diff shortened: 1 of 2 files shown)


N0NB:
I can confirm the segfault in 12ba55834 exists in 4.6.4.  The other one appears to be self inflicted.  😉

N0NB:
Thanks, Daniele.

PR #1846, opened 2025-08-13
Accumulated patches

GeoBaltz:
Fix some warnings found by cppcheck and gcc -fanalyzer
Mangle variables to avoid warnings in perl bindings
Update uthash.h to latest release version.

### DIFF
--- a/include/hamlib/rig.h
+++ b/include/hamlib/rig.h
@@ -1908,17 +1908,18 @@ struct rig_spectrum_line
 
 /**
  * Config item for deferred processing
+ *  (Funky names to avoid clash with perl keywords. Sheesh.)
  **/
 struct deferred_config_item {
-  struct deferred_config_item *next;
+  struct deferred_config_item *nextt;
   hamlib_token_t token;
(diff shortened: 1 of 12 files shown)


N0NB:
I've contemplated updating `uthash.h` a few times over the past few years but always chickened out!

N0NB:
Thanks, George.

Regarding the name change, I might have suggested a prefix like `hl_` but no need to change them again.

GeoBaltz:
When I saw the old uthash.h was released in 2011 I figured it couldn't hurt... It did fix the diagnostics I got from both cppcheck and gcc, and list everything correctly.

As for the name changes - I've probably written code like that a couple dozen times before, first time I've used anything other than first,next,last.

dforsi:
We can also rename or hide symbols with the file `bindings/ignore.swg`

GeoBaltz:
Thanks for mentioning that - it probably would have been OK to just drop this as it shouldn't be used outside Hamlib.

OTOH, there is a not-unreasonable expectation of consistency between C the other languages.

PR #1848, opened 2025-08-17
Add more files and directories for "make dist"

dforsi:
This PR adds existing files and directories to the archive created with `make dist` and `make distcheck` and removes some that weren't used.
It also builds more existing tests, but it doesn't execute them during `make check` because they would fail or are too simple:
* `example` (is an example)
* `func_chk` (checks Hamlib sources)
* `sendraw` (needs a rig or a simulator)
* `simple` (is an example)
* `testnet` (has hardcoded addresses that won't resolve)
* `testsecurity` (this could be run and could be added in another PR)


GeoBaltz:
-1 for this.  Very few (if any) of these files are in any shape to be distributed.  And are of very little use (again, if any) to anyone not already working with the full git repo. 

The distribution tarball is just that - what it takes for a distro to make a working hamlib.  Anything else is just part of the developer's toolkit.

dforsi:
In my opinion, a distribution tarball is for developers and users that do not use git.
It is true that github provides a way to download a snapshot of the repository, but it doesn't check if it actually builds.

GeoBaltz:
I don't think anyone can be called a developer without using git.  Maybe 5 years ago, but not today.

To revert back to my working days - the "deliverables" of hamlib is hamlib itself; not all the cruft that has accumulated in our repository.  We have enough to do to QA the product, let alone all the other flotsam and jetsam this adds.

Let's look at the list:
- example.c - uses a routine that is NOT marked as part of HAMLIB_API. There's a separate declaration just to get it to compile.
- func_chk.c - very prone to false positives. Would need to be re-written (lex/bison) to be reliable.
- testnet - A quick hack by Mike to test some things on his machine. I don't think even he meant for it to be generalized.

None are necessary to build hamlib, so the distributions don't need them.  And I don't think the WSJT-X folks would like to distribute them, either.

Make that -10.

N0NB:
A lot of commits to wade through, Daniele.  I'll get to looking at them later.

My philosophy aligns with George's.  The source tarball should be limited to that source that is of the greatest value to distribution maintainers, downstreams that incorporate the library, and end users that prefer to compile things.  In fact, we might be including too much, but Mike included them for a reason and it's likely that someone is using them.  As our software doesn't "phone home" we don't know what is being used or not.

I'll give further feedback as I go through the list.

dforsi:
I can split some commits or some changes to a different PR to make it
easier to review (but then it would be changes that are easier to
review anyway), there may be only 4 or 5, like:
* fix typos
* add files to .gitignore
* minor modifications to makefiles (ie NOT adding new files)

and then rebase this PR on top of the new one

N0NB:
Patience, please.  I had a long and rather exhausting weekend.  I'll work through this as noted above.

N0NB:
Regarding commit f78625b50.  I think the comment is a non sequitur to the flag.  It could probably be better stated as:

``` c
     RIG_PARM_KEYERTYPE =    (1 << 11),  /*!< \c KEYERTYPE -- Multiple Morse keyer types; 0=NO, 1=YES.  \sa rig_keyertype_e */
```

N0NB:
I think commit a8726cbb1 should be dropped entirely.  The additional files are not relevant to the dist tarball.

N0NB:
Regarding commit 2160d5017e, `README.python` should probably be edited to remove references to Python 2 as it is no longer supported by upstream.

N0NB:
Drop commit 5d0daa2cf.

N0NB:
Drop commit c07b1542.

N0NB:
Drop commit 60a23e45.

Nothing includes hamlib/winpthreads.h so I don't see value in distributing it.

N0NB:
Drop commits ad7d233, 226be75e, e9c75ec52.

Is there any value in distributing these files at all?  For the former commit, those files were added at a time when it was not so easy to find and view a repository's contents.  These days browsing a Git repository is easy and with the ability of anyone to create a fork or clone a copy locally, I don't see the point of distributing these files.

I'd say the only file to distribute in any of these is `Android.mk` and even then, I'm not sure that works any more.

N0NB:
Drop commits 7c291ce4, 884597e.

N0NB:
Drop commit 0a72108.

N0NB:
Drop commit 4a7a2e7.

Moving `hamlib_voice_cw_memory_4.6.0.xlsx` to `doc` is fine, but we will not distribute a proprietary formatted file.  I should see if this will open in Libre Office and save it as an Open Document Spreadsheet (`.ods`) instead.

Windows developers...

N0NB:
Drop commit befc234.  Of the additions, only `docker-build` might be of value.  The others are simply old code that no one maintains but remain in case someone wants to pick them up.

My working tree doesn't have a `perl` directory or file at the top level.  I see, removed by commit cd5296e4 which follows in sequence.

Kylix is a dead language as far as I know.

N0NB:
Drop commit ab46728.

N0NB:
I will do a rebase dropping the commits I noted and checking things before merging.

N0NB:
My rebase lost a bunch of commits.

@dforsi  Please create a new PR without the commits I noted to drop.  Also, please limit the scope of a PR to just typos, or source fixes, or modifying distributed files.  I think that 6 to 10 commits per PR is enough.

I'm not going to merge any of this PR.

N0NB:
I think I got it!  I didn't do a `git rebase --continue` when I should have.  I'll do some checking now.  Stay tuned.

N0NB:
I've completed the rebase and fix a couple of warnings and an error from MinGW.

I pushed it to my GitHub: https://github.com/N0NB/Hamlib/pull/new/dforsi-fix/make-dist

Look things over before I merge.

dforsi:
I compiled the branch from your repository without issues.

The commits look good, except maybe that in [tests/Makefile.am](https://github.com/Hamlib/Hamlib/compare/master...N0NB:Hamlib:dforsi-fix/make-dist?expand=1#diff-32103f666ff2fb42b025a47ccf1b959bbcc6db89f217e5943b1de73c81a4f9db)  you are keeping all the files that I added in commit a0f7408a01, is that on purpose? example.c func_chk and testnet were criticized by George.

> bin_PROGRAMS = example func_chk rigctl rigctld rigmem rigsmtr rigswr rotctl rotctld rigctlcom rigctltcp rigctlsync ampctl ampctld rigtestmcast rigtestmcastrx $(TESTLIBUSB) rigfreqwalk sendraw simple testnet testsecurity

GeoBaltz:
The last two of those are definitely not fit for distribution.  example.c could be cleaned up a bit and maybe added to doc/, but it still isn't needed by the distros to build Hamlib, so drop it for now.

N0NB:
Can I use the excuse that I lost track?

Well, I had some good exercise coming up with that fix.  Heh!

N0NB:
I dropped commit a0f7408 and modified my commit to leave the changes in `example.c` and `testsecurity.c` and commented the lines to build testsecurity as documentation.

As i deleted the former branch I had pushed, the current is the same link, https://github.com/N0NB/Hamlib/pull/new/dforsi-fix/make-dist with the changes.

N0NB:
No complaints?  I guess I'll merge it after while.

dforsi:
Ok for me.

N0NB:
Thanks, Daniele.

PR #1849, opened 2025-08-17
Fix output of dump_state

dforsi:
Fixes arguments to `*_IS_FLOAT` (it's levels, not indexes) and properly handles string parameters.

While it's hard to see it there is any difference in float printed as integers or vice versa, for strings 10 PARM_BANDSELECT and 11 PARM_KEYERTYPE it is easy.

Steps to reproduce:
> tests/rigctl chk_vfo dump_state

Before:
`parm_gran=[...]10=0,0,1.25587e+11;11=0,0,1.26077e+11;`

After:
`parm_gran=[...]10=BANDUNUSED,BAND70CM,BAND33CM,BAND23CM;11=STRAIGHT,BUG,PADDLE;`

~~I must be missing something because `dump_state` without `chk_vfo` shows a totally different output.~~
It is documented in the man page that `check_vfo` is needed because of this line in `rigctl_parse.c`:
>     chk_vfo_executed = 1; // this allows us to control dump_state version

PR #1850, opened 2025-08-17
Feature/tests

dforsi:
The final goal of this PR is checking that `--help` and `?` of the main test commands print the expected text for Dummy.
One issue is that the debug output is printed mixed to the output of `tests/ampctl --help`, like this:
```
usage_amp: nbspace left=3
F: set_freq    (Frequency(Hz))
usage_amp: nbspace left=16
f: get_freq    ()
usage_amp: nbspace left=11
l: get_level   (Level)
usage_amp: nbspace left=-2
L: set_level   (Level, Level Value)
usage_amp: nbspace left=13
w: send_cmd    (Cmd)
usage_amp: nbspace left=16
?: dump_state  ()
usage_amp: nbspace left=16
1: dump_caps   ()
usage_amp: nbspace left=16
_: get_info    ()
usage_amp: nbspace left=11
R: reset       (Reset)
usage_amp: nbspace left=4
?: set_powerstat(Power Status)
usage_amp: nbspace left=16
?: get_powerstat()
```
The easy fix would be deleting the line that prints that debug messages, which isn't present in rigctl and rotctl, however I think that it is a bug or a missing feature that `debug,c` sets the debug level to trace and the user could not change it before `getopt_long` loop ends (only `--help` and `--list` are affected).

Other test programs override the choice of the user with this logic:
> rigmem.c:    rig_set_debug(verbose < 2 ? RIG_DEBUG_WARN : verbose);
> rigsmtr.c:    rig_set_debug(verbose < 2 ? RIG_DEBUG_WARN : verbose);
> rigswr.c:    rig_set_debug(verbose < 2 ? RIG_DEBUG_WARN : verbose);

which IMHO is wrong because they can't run quietly, but they aren't the main programs.

so:
* I'm replacing `0` with `RIG_DEBUG_NONE`, `RIG_DEBUG_ERROR` would be more appropriate but we are missing a quiet option ('q' and 'Q' are reserved for `quit` in interactive mode, but maybe one can be reused in the command line)
* I'm setting the verbosity level each time that `-v` is handled so that the current behavior can be restored with `-vvvvv`
* I'm directing the output of `--help` to `stderr` when the program is about to do `exit(1)`
* I'm directing the output of error messages to `stderr`, missing any comment, I don't know if it was sent to `stdout` on purpose, but if running those program in a script cheching if stderr is not empty seems the only way to detect errors, maybe it's unneeded, so this part ch be dropped)
* I'm deleting some duplicated code and making some help test identical
* I'm adding 2 scripts
  *  one to compare the output of `--help` with a reference file (this allowed me to make some refactoring making sure that I didn't change the help texts)
  *  one to compare the output of all internal commands of ampctl, rigctl, rotcl, which without this PR changes every time because the debug message at trace level print memory addresses (pointer) that change at every run, and also some debug messages about the cache are printed in different order, that's why I though about redirecting debug messages to stderr

Setting as draft because it depends on #1849 an maybe other changes will be needed.

### DIFF
--- a/tests/ampctl.c
+++ b/tests/ampctl.c
@@ -116,7 +116,7 @@ int main(int argc, char *argv[])
     int retcode;        /* generic return code from functions */
     int exitcode;
 
-    int verbose = 0;
+    int verbose = RIG_DEBUG_NONE;
     int show_conf = 0;
     int dump_caps_opt = 0;
 
--- a/tests/ampctl.c
+++ b/tests/ampctl.c
@@ -132,6 +132,7 @@ int main(int argc, char *argv[])
     int serial_rate = 0;
     char conf_parms[MAXCONFLEN] = "";
 
+    rig_set_debug(verbose);
     while (1)
     {
         int c;
@@ -216,6 +217,7 @@ int main(int argc, char *argv[])
 
         case 'v':
             verbose++;
+            rig_set_debug(verbose);
             break;
 
         case 'L':
@@ -223,7 +225,6 @@ int main(int argc, char *argv[])
             break;
 
         case 'l':
-            rig_set_debug(0);
             list_models();
             exit(0);
 
@@ -240,8 +241,6 @@ int main(int argc, char *argv[])
             exit(1);
         }
     }
-
-    rig_set_debug(verbose);
 
     rig_debug(RIG_DEBUG_VERBOSE, "ampctl %s\n", hamlib_version2);
     rig_debug(RIG_DEBUG_VERBOSE, "%s",
--- a/tests/ampctld.c
+++ b/tests/ampctld.c
@@ -179,6 +179,7 @@ int main(int argc, char *argv[])
 #endif
 #endif
 
+    rig_set_debug(verbose);
     while (1)
     {
         int c;
@@ -245,6 +246,7 @@ int main(int argc, char *argv[])
 
         case 'v':
             verbose++;
+            rig_set_debug(verbose);
             break;
 
         case 'L':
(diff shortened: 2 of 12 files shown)


dforsi:
I dropped some commits for which I sent separated pull requests or will send and squashed some.

The first 2 commits just replace zeros with `RIG_ERR_NONE`, they do not change behavior in any way.
The 3rd commit changes the default behavior only for the following commands to avoid printing trace messages by default:
```
tests/ampctld -l >/dev/null
tests/ampctl -h >/dev/null
tests/rigctld -l >/dev/null
tests/rigctltcp -l >/dev/null
tests/rotctld -l >/dev/null
```
The debug messages for other commands and options are unaffected because `--list` and `--help` are the only options that are executed in the loop around `getopt_long`. If debug messages of any level are desired for code executed in the loop, it is sufficient to precede the commands with the usual `-v`, eg. `tests/ampctl -vvvvv -h` will show the same output that I copied in the first comment.

I'm attaching [list-help-stderr.txt](https://github.com/user-attachments/files/21900464/list-help-stderr.txt)  that shows the issue in all software, taken with these bash commands:
```bash
{
tests/ampctl -h
tests/ampctl -l
tests/ampctld -h
tests/ampctld -l
tests/rigctl -h
tests/rigctl -l
tests/rigctlcom -h
tests/rigctlcom -l
tests/rigctld -h
tests/rigctld -l
tests/rigctlsync -h
tests/rigctlsync -l
tests/rigctltcp -h
tests/rigctltcp -l
tests/rotctl -h
tests/rotctl -l
tests/rotctld -h
tests/rotctld -l
} >/dev/null 2>list-help-stderr.txt
```
Executing the same commands after the changes shows no output, as expected, because `--list` and `--help` do not raise errors (usually at least), adding more and more `-v` shows more messages.

N0NB:
It works as advertised here.

Thanks much for paring down the number of commits in this PR.  It is just so much easier to review and test a small set of commits that are related in scope.

Thanks, Daniele.

PR #1854, opened 2025-08-18
Fix man pages

dforsi:
This PR adds that `set_gpio` needs a parameter with value 0 or 1, and moves `freq_skip` to the proper section of the man page.


N0NB:
Thanks, Daniele.

Issue #1376, opened 2023-09-03
FLRig missing levesl


mdblack98:
Seems we don't have level queries to FLRig for
ATT
IF
AGC
PREAMP

dforsi:
flrig 2.0.06 added `rig.get_agc`

The currrent version 2.0.08 doesn't seem to support getting ATT IF PREAMP.
There is no `rig.get_att` `rig.get_if` `rig.get_preamp`; for ATT and PREAMP there are methods to get the label shown in the UI, (`rig.get_att_label` and `rig.get_pre_label` but not a value.

These are the methods implemented in flrig 2.0.08:
```
grep -oE 'XmlRpcServerMethod\("[^"]+"' src/server/xml_server.cxx | grep -oE '"[^"]+"' | sort
"main.get_version"
"main.set_frequency"
"rig.cat_priority"
"rig.cat_string"
"rig.client_string"
"rig.cmd"
"rig.cwio_get_wpm"
"rig.cwio_send"
"rig.cwio_set_wpm"
"rig.cwio_text"
"rig.freqA2B"
"rig.fskio_text"
"rig.get_60M_label"
"rig.get_60M_labels"
"rig.get_AB"
"rig.get_agc"
"rig.get_agc_label"
"rig.get_agc_labels"
"rig.get_an_label"
"rig.get_an_labels"
"rig.get_att_label"
"rig.get_att_labels"
"rig.get_bk_label"
"rig.get_bk_labels"
"rig.get_bw"
"rig.get_bwA"
"rig.get_bwB"
"rig.get_bws"
"rig.get_DBM"
"rig.get_info"
"rig.get_maxpwr"
"rig.get_micgain"
"rig.get_mode"
"rig.get_modeA"
"rig.get_modeB"
"rig.get_modes"
"rig.get_nb_label"
"rig.get_nb_labels"
"rig.get_notch"
"rig.get_nr_label"
"rig.get_nr_labels"
"rig.get_pbt"
"rig.get_pbt_inner"
"rig.get_pbt_outer"
"rig.get_power"
"rig.get_pre_label"
"rig.get_pre_labels"
"rig.get_ptt"
"rig.get_pwrmeter"
"rig.get_pwrmeter_scale"
"rig.get_rfgain"
"rig.get_sideband"
"rig.get_smeter"
"rig.get_split"
"rig.get_Sunits"
"rig.get_SWR"
"rig.get_swrmeter"
"rig.get_update"
"rig.get_vfo"
[...]

PR #1855, opened 2025-08-19
Implement get_agc for flrig 2.0.06+

dforsi:
Test case:
run: flrig configured for the rig in use
run: tests/rigctl -m 4 get_level AGC
expected: should print an enum agc_level_e (an integer in range 0..10)

Part of issue #1376.


N0NB:
Thanks, Daniele.

PR #1858, opened 2025-08-19
Fix help texts

dforsi:
This PR:
* makes the help texts more similar in the various test programs
* removes some code duplication

The help texts of rigctl* show also the list error codes, while the help texts for ampctl* and rotctl* do not.
Some errors apply only to rigs, but the same list could printed with all help texts, or similar lists where the word "rig" is replaced with amplifier or rotator where it makes sense.

Currently at least the following errors are returned for amplifiers and rotators
```
grep -hEro "RIG_E\\w+" amplifiers/ src/amp* 2>/dev/null | sort -u
RIG_EINVAL
RIG_EIO
RIG_ENAVAIL
RIG_ENOMEM
RIG_EPROTO

grep -hEro "RIG_E\\w+" rotators/ src/rot* 2>/dev/null | sort -u
RIG_EINTERNAL
RIG_EINVAL
RIG_EIO
RIG_ELIMIT
RIG_ENAVAIL
RIG_ENIMPL
RIG_ENOMEM
RIG_EPROTO
RIG_ERJCTED
RIG_ETIMEOUT
RIG_ETRUNC
```

### DIFF
--- a/tests/rotctl.c
+++ b/tests/rotctl.c
@@ -509,6 +509,4 @@ void usage()
     );
 
     usage_rot(stdout);
-
-    printf("\nReport bugs to <hamlib-developer@lists.sourceforge.net>.\n");
 }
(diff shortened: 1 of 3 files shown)


N0NB:
Thanks, Daniele.

PR #1859, opened 2025-08-19
Add missing string form for RIG_FUNC_SLICE

dforsi:
The define `RIG_FUNC_SLICE` doesn't seem to be use anywhere, though.


N0NB:
Thanks, Daniele.

Issue #1564, opened 2024-06-08
Modularize set_conf token processing


mdblack98:
Move set_conf token processing for rigctl, rigctld, rotctl, rotcld, ampctl, ampctld into the library.
This will reduce the footprint of each a bit and also allow reduction of the double calls in rotctl/rotctld.
See commit 1557ad70f752f7217c1eff963af1a1f3c4f96026

dforsi:
commit 1557ad70f752f7217c1eff963af1a1f3c4f96026 referenced above was reverted one week later with commit 341a205bd7e5765e64a2e5f35ed2264b2cf9e6e2 without explanation i the commit message.

I don't think that there is much gain in moving set_conf to the library because:
* each family of devices has its set of methods `amp/rig/rot_token_lookup()` `amp/rig/rot_set_conf()`
* the comma as separator works for the command line but may be unsuitable for other clients or other sources of configuration with structured formats (ini, JSON, XML) with key/value format or line based or CSV with different separators
* the code in `rigctl` has a special case for PTT:
   > ptt_type = my_rig->caps->ptt_type; // in case we set the ptt_type with set_conf
* only at this point 😒 I have understood what Mike meant with "reduction of the double calls in rotctl/rotctld" some is easy to fix just using the existing variable (need to write a test case)

PR #1861, opened 2025-08-20
Remove double call to rig_token_lookup() and rot_token_lookup()

dforsi:
As mentioned by Mike in issue #1564, rotctl/rotctld were doing double calls (in fact `set_conf()` of `rotctl_parse.c`) was calling `rot_token_lookup()`; a similar thing were doing rigctl/rigctld (in `set_conf()` of `rigctl_parse.c`). This PR fixes both cases, using an existing variable.

The code in `ampctl_parse.c` is very different and wasn't doing the token lookup twice.

Test case:
```
tests/rigctl --set-conf=rig_pathname=test,write_delay=1,timeout=2 --show-conf Q | grep --no-group-separator -A1 -E "(rig_pathname|^write_delay|^timeout):"

tests/rotctl --set-conf=rot_pathname=test,write_delay=1,timeout=2 --show-conf Q | grep --no-group-separator -A1 -E "(rot_pathname|^write_delay|^timeout):"
```

Output should be unchanged by the changes in this PR:
```
rig_pathname: "Path name to the device file of the rig"
        Default: /dev/rig, Value: test
write_delay: "Delay in ms between each byte sent out"
        Default: 0, Value: 1
timeout: "Timeout in ms"
        Default: 0, Value: 2
Command 'Q' not found!
rot_pathname: "Path name to the device file of the rotator"
        Default: /dev/rotator, Value: test
write_delay: "Delay in ms between each byte sent out"
        Default: 0, Value: 1
timeout: "Timeout in ms"
        Default: 0, Value: 2
```

### DIFF
--- a/tests/rigctl_parse.c
+++ b/tests/rigctl_parse.c
@@ -6102,7 +6102,7 @@ declare_proto_rig(set_conf)
     }
     else
     {
-        ret = rig_set_conf(rig, rig_token_lookup(rig, arg1), arg2);
+        ret = rig_set_conf(rig, mytoken, arg2);
     }
 
     return (ret);
--- a/tests/rotctl_parse.c
+++ b/tests/rotctl_parse.c
@@ -1738,7 +1738,7 @@ declare_proto_rot(set_conf)
     }
     else
     {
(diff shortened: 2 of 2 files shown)


N0NB:
Thanks, Daniele.

Issue #1730, opened 2025-05-10
configure fails if a C++11 compiler is not found even if no C++11 code is going to be built


dforsi:
In `configure.ac` there is a macro call added in commit dd376a1267 that makes a C++11 compiler a mandatory build dependency:
> AX_CXX_COMPILE_STDCXX([11],[noext],[mandatory])

However the main library and the utilities are written in C and in fact after commenting out that line, they can be built.

It is currently undocumented what parts of Hamlib need a C++ or C++11 compiler:
* C++ bindings (can be disabled at configure time `--without-cxx-binding`, fix proposed in PR 1729)
* USRP (can be disabled at configure time `--enable-usrp`)
* INDI (can be disabled at configure time `--without-indi`)
 
There is something else that hasn't an explicit option in configure?

N0NB:
Commit dd376a1 references Issue #740 which may shed some light on the C++11 question.  It seems at that time--mid 2021--Mike deemed this a critical change.

PR #1862, opened 2025-08-21
Make a C++11 compiler mandatory if INDI or C++ binding are requested

dforsi:
Otherwise make it optional.

Fixes issue #1730.

Regardless if a C++ compiler is installed or not, nothing changes when running `configure` without arguments or with explicit `--with-cxx-binding --with-indi`.

If a C++ compiler is NOT installed, before this PR the configure script fails even if called as:
> ./configure --without-cxx-binding --without-indi
```
configure: error: *** A compiler with support for C++11 language features is required.
```
We can simulate the absence of a C++ compiler passing `CXX=/usr/bin/false` to configure:
> ./configure  CXX=/usr/bin/false --without-indi --without-cxx-binding 

With this PR, configure succeeds and prints
```
    With C++ binding                no
...
    With INDI support               no
```
The build succeeds.

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -68,6 +68,7 @@ jobs:
       run: |
         brew install autoconf
         brew install automake
+        brew install libnova
         brew install libtool
         brew install grep
         brew install swig
@@ -86,7 +87,7 @@ jobs:
       run: ./configure ${{ matrix.configure_args }}
     - name: configure on macOS
       if: runner.os == 'macOS'
-      run: ./configure ${{ matrix.configure_args }} --without-lua-binding PYTHON=/opt/homebrew/bin/python3
+      run: ./configure ${{ matrix.configure_args }} --with-indi=detect --without-lua-binding PYTHON=/opt/homebrew/bin/python3
     - name: make
       run: make -j 4 V=0 --no-print-directory
     - name: make distcheck
@@ -94,4 +95,4 @@ jobs:
       run: make distcheck V=0 --no-print-directory AM_DISTCHECK_CONFIGURE_FLAGS="${{ matrix.configure_args }} --without-perl-binding"
     - name: make distcheck on macOS
       if: runner.os == 'macOS'
-      run: make distcheck V=0 --no-print-directory AM_DISTCHECK_CONFIGURE_FLAGS="${{ matrix.configure_args }} --without-perl-binding --without-lua-binding --without-tcl-binding PYTHON=/opt/homebrew/bin/python3"
+      run: make distcheck V=0 --no-print-directory AM_DISTCHECK_CONFIGURE_FLAGS="${{ matrix.configure_args }} --with-indi=detect --without-perl-binding --without-lua-binding --without-tcl-binding PYTHON=/opt/homebrew/bin/python3"
--- a/configure.ac
+++ b/configure.ac
@@ -111,10 +111,10 @@ AC_PROG_AWK
 AC_PROG_INSTALL
 AC_PROG_LN_S
 
-AS_IF([test -z "$CXX"], [
-	cf_with_cxx=yes,
-	cf_with_cxx=no
-])
+AS_IF([test -z "$CXX"],
+	[cf_with_cxx=no],
+	[cf_with_cxx=yes]
+)
 AM_CONDITIONAL([ENABLE_CXX], [test x"${cf_with_cxx}" = "xyes"])
 
 dnl Broke on older Automake, so test for its existence.
--- a/configure.ac
+++ b/configure.ac
@@ -111,7 +111,7 @@ AC_PROG_AWK
 AC_PROG_INSTALL
 AC_PROG_LN_S
 
-dml check whether CXX is functional
+dnl check whether CXX is functional
 AX_CXX_COMPILE_STDCXX([11],[noext],[optional])
 AS_IF([test "$HAVE_CXX11" = 1],
        [cf_with_cxx=yes],
--- a/configure.ac
+++ b/configure.ac
@@ -407,18 +407,23 @@ AC_MSG_CHECKING([whether to use INDI in rotctl/rotctld])
 
 AC_ARG_WITH([indi],
     [AS_HELP_STRING([--without-indi],
-        [disable INDI in rotctl/rotctld @<:@default=yes:>@])],
+        [disable INDI in rotctl/rotctld @<:@default=detect:>@])],
     [cf_with_indi_support=$withval],
-    [cf_with_indi_support=yes]
+    [cf_with_indi_support=detect]
 )
 
 AC_MSG_RESULT([$cf_with_indi_support])
 
-AS_IF([test x"$cf_with_indi_support" = "xyes"], [
-    AS_IF([test x"${cf_with_cxx}" != "xyes"], [
-        AC_MSG_ERROR([INDI support needs a C++ compiler.])
-    ])
-])
+case "${cf_with_indi_support}" in
+  detect)   cf_with_indi_support=$cf_with_cxx ;;
+  yes)
+            AS_IF([test x"${cf_with_cxx}" != "xyes"],
+                [AC_MSG_ERROR([INDI support needs a C++ compiler.])]
+            )
+            ;;
+  no)       ;;
+    *)      AC_MSG_ERROR([invalid value for --without-indi: ${cf_with_indi_support}]) ;;
+esac
 
 AS_IF([test x"$cf_with_indi_support" != "xno"], [
     # macros/ax_lib_nova.m4
--- a/configure.ac
+++ b/configure.ac
@@ -414,37 +414,35 @@ AC_ARG_WITH([indi],
 
 AC_MSG_RESULT([$cf_with_indi_support])
 
-case "${cf_with_indi_support}" in
-  detect)   cf_with_indi_support=$cf_with_cxx ;;
-  yes)
-            AS_IF([test x"${cf_with_cxx}" != "xyes"],
-                [AC_MSG_ERROR([INDI support needs a C++ compiler.])]
-            )
-            ;;
-  no)       ;;
-    *)      AC_MSG_ERROR([invalid value for --without-indi: ${cf_with_indi_support}]) ;;
-esac
-
 AS_IF([test x"$cf_with_indi_support" != "xno"], [
     # macros/ax_lib_nova.m4
     AX_LIB_NOVA
 
-    AS_IF([test x"$ax_cv_lib_nova" = "xno"], [
-	AC_MSG_WARN([libnova support not found, required by INDI.])
-	cf_with_indi_support=no
-	])
-
-    AS_IF([test x"$ax_cv_lib_nova" != "xno"], [
-	# macros/ax_lib_indi.m4
-	AX_LIB_INDI
-
-	AS_IF([test x"$ax_cv_lib_indi" = "xno"], [
-	    AC_MSG_WARN([INDI support not found.])
-	    cf_with_indi_support=no
-	])
-
-    ])
+    # macros/ax_lib_indi.m4
+    AX_LIB_INDI
 ])
+
+AS_IF([test x"$cf_with_indi_support" = "xyes"], [
+    AS_IF([test x"$cf_with_cxx" = "xno"],
+        [AC_MSG_ERROR([INDI support needs a C++ compiler.])]
+    )
+
+    AS_IF([test x"$ax_cv_lib_nova" = "xno"],
+        [AC_MSG_ERROR([libnova support not found, required by INDI.])]
+    )
+
+    AS_IF([test x"$ax_cv_lib_indi" = "xno"],
+        [AC_MSG_ERROR([libindi support not found, required by INDI.])]
+    )
+])
+
+AS_IF([test x"$cf_with_indi_support" = "xdetect"], [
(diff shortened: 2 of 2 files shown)


dforsi:
The Mac build was silently failing to build INDI, and there is a glitch in the output, eg. this is the PR #1861 of yesterday:

```
checking for readline/history.h... yes
configure: WARNING: INDI support needs a C++ compiler.
checking whether to use INDI in rigctl/rotctl... checking whether to build HTML rig feature matrix... checking for gd.h... no
no
```
https://github.com/Hamlib/Hamlib/actions/runs/17110489719/job/48530469799?pr=1861#step:9:239

N0NB:
Seems the test not seeing `x"$cf_with_cxx"` with the `"xyes"` value.  On Linux I sww that `config.log` has `cf_with_cxx='yes'`.  I'll have to do some checking to see if the `config.log` file can be read via the CI UI.

dforsi:
Yes, reading `config.log` would save time.

In this case the issue is caused by the line

> AC_CHECK_PROG([cf_with_cxx], ["${CXX}"], [yes], [no])

I added and `echo ${CXX}` on Mac is `g++ -std=gnu++11` not just the name of the program.

Comparing the output showed that `-std=gnu++11` was a difference:

Ubuntu
> checking for gcc option to enable C11 features... (cached) none needed
> ...
> checking for g++... yes

Mac
> checking for g++ option to enable C++11 features... -std=gnu++11
> ...
> checking for g++ -std=gnu++11... no

### Ubuntu
```
checking dependency style of gcc... gcc3
checking whether make supports nested variables... (cached) yes
checking for gcc... (cached) gcc
checking whether the compiler supports GNU C... (cached) yes
checking whether gcc accepts -g... (cached) yes
checking for gcc option to enable C11 features... (cached) none needed
checking whether gcc understands -c and -o together... (cached) yes
checking for g++... g++
checking whether the compiler supports GNU C++... yes
checking whether g++ accepts -g... yes
checking for g++ option to enable C++11 features... none needed
checking dependency style of g++... gcc3
checking how to run the C preprocessor... gcc -E
checking for gawk... (cached) gawk
checking whether ln -s works... yes
checking for g++... yes
```

### Mac
```
checking dependency style of gcc... gcc3
checking for gcc... (cached) gcc
checking whether the compiler supports GNU C... (cached) yes
checking whether gcc accepts -g... (cached) yes
checking for gcc option to enable C11 features... (cached) none needed
checking whether gcc understands -c and -o together... (cached) yes
checking for g++... g++
checking whether the compiler supports GNU C++... yes
checking whether g++ accepts -g... yes
checking for g++ option to enable C++11 features... -std=gnu++11
checking dependency style of g++ -std=gnu++11... gcc3
checking how to run the C preprocessor... gcc -E
checking for gawk... (cached) awk
checking whether ln -s works... yes
checking for g++ -std=gnu++11... no
```

N0NB:
I think Mark @markjfine develops on a Mac, maybe he can test this PR and help troubleshoot this?

markjfine:
I'm thinking this might be a clang thing masquerading as something gnu++11-compatible. Lemme pull this down and see what I end up with. I have gcc installed via Homebrew, but I don't think I have gnu-c++.

markjfine:
Mine won't build with INDI because it fails the libnova test. I don't have that installed.

Otherwise, here's what I'm getting with no configure switches:
```
checking dependency style of gcc... gcc3
checking for gcc... (cached) gcc
checking whether the compiler supports GNU C... (cached) yes
checking whether gcc accepts -g... (cached) yes
checking for gcc option to enable C11 features... (cached) none needed
checking whether gcc understands -c and -o together... (cached) yes
checking for g++... g++
checking whether the compiler supports GNU C++... yes
checking whether g++ accepts -g... yes
checking for g++ option to enable C++11 features... none needed
checking dependency style of g++... gcc3
checking how to run the C preprocessor... gcc -E
checking for gawk... (cached) awk
checking whether ln -s works... yes
*** CXX is 'g++'
checking for g++... yes
```
the output shows:
```
 Hamlib Version 4.7~git configuration:

 Prefix 	/usr/local
 Preprocessor	gcc -E 
 C Compiler	gcc -g -O2 
 C++ Compiler	g++ -std=c++11 -g -O2

 Package features:
    With C++ binding		    yes
    With Perl binding		    no
    With Python binding 	    no
    With TCL binding		    no
    With Lua binding		    no
    With rigmem XML support	    no
    With Readline support	    yes
    With INDI support		    no

    Enable HTML rig feature matrix  no
    Enable WinRadio		    yes
    Enable Parallel		    yes
    Enable USRP 		    no
    Enable USB backends 	    no
    Enable shared libs		    yes
    Enable static libs		    yes
    Enable Python tests		    no
```
Note that it says `c++11`, not `gnu++11`

if I add CXXFLAGS=-std=gnu++11 it says the same during configuration, but it just adds the `-std=gnu++11` to the C++ Compiler line in the output:
```
C++ Compiler	g++ -std=c++11 -std=gnu++11
```

Again, I'm thinking this may be a clang thing.
Let me know if there's anything specific you need me to look at and I'll be more than happy to oblige.

dforsi:
Yes, please @markjfine can you confirm that passing `false` as C++ compiler is a valid test case also for your system?

Running `./configure CXX=/usr/bin/false --without-indi --without-cxx-binding` should finish without errors, show that the C++ compiler is "/usr/bin/false" and that C++ binding and INDI are both "no".

It may be a clang thing because it needs the extra argument, but the error is that `AC_CHECK_PROG()` expects the name of a file instead it receives an entire command line on github: it gets `g++ -std=gnu++11` instead of `g++` and it tries to find a file named exactly `g++ -std=gnu++11`

markjfine:
@dforsi 

`./configure CXX=/usr/bin/false --without-indi --without-cxx-binding`
Completes with no errors:
```
Hamlib Version 4.7~git configuration:

 Prefix 	/usr/local
 Preprocessor	gcc -E 
 C Compiler	gcc -g -O2 
 C++ Compiler	/usr/bin/false 

 Package features:
    With C++ binding		    no
    With Perl binding		    no
    With Python binding 	    no
    With TCL binding		    no
    With Lua binding		    no
    With rigmem XML support	    no
    With Readline support	    yes
    With INDI support		    no

    Enable HTML rig feature matrix  no
    Enable WinRadio		    yes
    Enable Parallel		    yes
    Enable USRP 		    
    Enable USB backends 	    no
    Enable shared libs		    yes
    Enable static libs		    yes
    Enable Python tests		    no
```

markjfine:
I should add this is an M1 running macOS 15.6.1 (just updated this morning), using Apple's xCode 16.4 as well as Homebrew's gcc v15.1.0.

dforsi:
Thank you to both, there are still things that do not work, but it's getting late here, so I'm setting this PR as draft.

The trick with `CXX=/usr/bin/false` works for the checks that try to compile (such the check done by libnova), but doesn't work for the simpler test that I added:
```
AS_IF([test -z "$CXX"],
	[cf_with_cxx=no],
	[cf_with_cxx=yes]
)
```
`AC_PROG_CXX` defaults to the hardcoded value `g++` if it doesn't find a compiler, so that $CXX is not empty and configure creates a makefile that tries to build the C++ bindings with a non-working C++ compiler.

N0NB:
Thanks, Mark!

markjfine:
You're both certainly welcome.

dforsi:
Looks good now.

This PR:
* moves the check for C++11 towards the beginning of `configure.ac` and makes it optional
* uses a standard(?) variable set by that check to set our variable
* uses our variable to fail the configure step if `--with-indi` is set by default or by the user and a C++11 compiler is not found, changing the warning to a fatal error (users need to call configure adding `--with-indi=no`)
* the same for `--with-cxx-binding`

N0NB:
I see a typo in your last commit.  `dml` should be `dnl` on line 114.

It seems to me that if a C++ compiler is not found that the script should disable INDI support and continue on without error.  For example:

```
$ ../hamlib/configure CXX=/usr/bin/false
```

Should not error out as follows:

```
checking whether to use INDI in rigctl/rotctl... yes
configure: error: INDI support needs a C++ compiler.
```

I don't even have the Debian `libindi-dev` package installed and apparently never have had it installed.  Evidently it not being available is fine so long as the C++ compiler is found, but if there is no such compiler then it fails with a hard error.  Can we make it handle this case gracefully?

dforsi:
It was the goal of this PR to fail the configuration step if you run `./configure --with-indi --with-cxx-binding` (which is the default) and the prerequisites aren't met.

Why "yes" should be ignored?

We can change the default to `detect` or `auto` or some other word that we decide (it's just a string that must be checked to be different from "yes" and "no"):
* `detect` is the same as when no arguments are passed (the behavior is like before, ie build if the prerequisites are met, print a warning otherwise)
* `no` do not build (the behavior is like before)
* `yes` build if prerequisites are met, fail otherwise (the new behavior)

N0NB:
> It was the goal of this PR to fail the configuration step if you run `./configure --with-indi --with-cxx-binding` (which is the default) and the prerequisites aren't met.

That is reasonable.  Here failure should occur if the user is specifically telling `configure` to use INDI and C++ but doesn't have them installed.

> Why "yes" should be ignored?

Not at all.  It needs to be honored but in the scenario above, the user needs to be informed that optional packages requested are not available so his system needs to be fixed.  Hard failure here is perfectly acceptable.

> We can change the default to `detect` or `auto` or some other word that we decide (it's just a string that must be checked to be different from "yes" and "no"):
> 
> * `detect` is the same as when no arguments are passed (the behavior is like before, ie build if the prerequisites are met, print a warning otherwise)
> * `no` do not build (the behavior is like before)
> * `yes` build if prerequisites are met, fail otherwise (the new behavior)

I think that `detect` is the proper way to go when the command line options are not given.  When C++ is present, build those bindings and when INDI is present build its support.

If C++ is not present (rare but conceivable) then the INDI check should be bypassed.  This should probably be the case if the user passes `--without-c++-binding` to `configure`.  In both cases testing for INDI should be silently ignored and `cf_with_indi_support` set to `no`.

I hope I'm making sense!

dforsi:
I added the option `detect` to `--with-cxx-binding` and `--with-indi` and it works as said: `yes` causes a fatal error if something isn't available, `no` disables building the option and `detect` builds it if it finds the prerequisites.

I removed the old `AC_MSG_WARN()` because yes/no is printed anyway during the checks and it makes `configure.ac` a little simpler:

```
checking whether to use INDI in rotctl/rotctld... detect
checking for nova library... checking for libnova/libnova.h... yes
-lnova -lstdc++ -lz
checking for INDI library... checking for libindi/baseclient.h... yes
-lindiclient -lstdc++ -lz
```
I think that the line with `-lnova -lstdc++ -lz` is printed by the macro `AX_LIB_NOVA` and `-lindiclient -lstdc++ -lz` by `AX_LIB_INDI`, I don't know why they weren't visible before.

I used the following script to test with the C++ compiler enabled/disabled and libindi installed/uninstalled:
```
for cxx in "" "CXX=false"; do
    for options in \
        "" \
        "--with-cxx-binding=yes --with-indi=yes" \
        "--with-cxx-binding=no --with-indi=no";
    do
        echo ./configure $cxx $options
        ./configure $cxx $options 2>&1 | grep -iE "( indi | nova |c\+\+ binding)"
        echo
    done
done | tee output.txt
```
[output.txt](https://github.com/user-attachments/files/21957159/output.txt)
[output-no-libindi.txt](https://github.com/user-attachments/files/21957160/output-no-libindi.txt)

I also installed libnova in the CI for macOS and added `--with-indi=detect`, it doesn't find it (a path issue on Mac?) so the configure continues disabling INDI.

N0NB:
I noticed one digraph typo in the "disable INDI" message.

``` diff
diff --git a/configure.ac b/configure.ac
index fb1cf573e..81e045539 100644
--- a/configure.ac
+++ b/configure.ac
@@ -407,7 +407,7 @@ AC_MSG_CHECKING([whether to use INDI in rotctl/rotctld])
 
 AC_ARG_WITH([indi],
     [AS_HELP_STRING([--without-indi],
-        [disable INDI in rotctl/rotctld @<:@default=detect:>@])],
+        [disable INDI in rotctl/rotctld @<:@default=detect@:>@])],
     [cf_with_indi_support=$withval],
     [cf_with_indi_support=detect]
 )
```

N0NB:
Thanks for your attention to detail, Daniele.

N0NB:
@dforsi, @markjfine What do you think about splitting the Continuous Integration script, one for Ubuntu latest and the other for MacOS latest?  While there would be some duplication between the `.yaml` files, they would each be focused on the particular OS and easier to follow.

markjfine:
I'm inclined to agree if it will help make the test procedure better fit the intended OS. That said, I'm not well-versed in the way the CI system under github works (nor yaml for that matter) to have an educated opinion.

dforsi:
I think that's a good idea because we can see if a failure is only in one OS.
It will also make it easier to add another workflow for Windows or
Raspberry or something else.
In the current workflow I like the sharing of the `--with` and
`--enable` options, while overriding values for some steps, I don't
know how to share them between different scripts.

I see that workflows can be executed manually with the trigger `on
workflow_dispatch`, this can be useful to test the split scripts
before having them run automatically, there isn't much documentation
because it seems to work similar to `on push`/`on pull_request`:

https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax?s=09#onworkflow_dispatch

N0NB:
As I noted, there will be duplication but that's likely not of much concern as they're not a part of the source distribution.

Thanks for the links.

I'll open a feature issue (am I taking inspiration from Mike?) since this is a close PR.

PR #1863, opened 2025-08-22
Bug Fixes for FTX1

aa5sh:
Needed to adjust one of the timeouts, was causing issues on some Windows computers.  Also made some changes due to IF command response size being 30, it was erroring out.  And last band mapping was a little different on this rig for 2m, 70cm, 4m and Air Band.


N0NB:
Thanks, Michael.

PR #1864, opened 2025-08-22
Fix build failure on openSUSE Tumbleweed, Slowroll and Leap 16.0

dforsi:
Thanks to @dl8fcl Walter.


N0NB:
Thanks Walter and Daniele.

Issue #1759, opened 2025-06-07
grbl_net_open() ignores errors


dforsi:
`grbl_net_open()` always returns `RIG_OK` even if `rot_get_conf()` or `grbl_net_open()` or `grbl_init()` returned an error.

N0NB:
Is this fixed, Daniele?

dforsi:
It isn't fixed.
I need to better understand the code and think how to test it.
The missing checks are probably irrelevant because some of the code is no-op and some I don't know if it is ever executed if `.init` `.open` `.close` are done directly by `src/rotator.c`, so the fix may be just deleting code.

PR #1866, opened 2025-08-23
Remove unused code and reduce amount of debug output of grbltrk.c

dforsi:
The main goal of this PR was detecting errors during the initialization, to get there I modified `rotator/grbltrk.c` to set the TRACE debug level to the lines that don't show errors and to remove the functions that only printed debug messages (this deletes some messages).

Closes issue #1759.

The ports are opened and closed by the main library so the code doesn't need to open them.
I deleted `grbl_net_open()` because is unneeded and that function was doing nothing anyway.
I deleted `grbltrk_rot_init()` because was doing nothing, in fact the initialization is the only thing remaining in `grbltrk_rot_open()`, the rest were debug  messages.

The last 2 commits propagate the error code (it was issue #1759).

I don't own the hardware, but the goal were the errors during the initialization, so my tests were limited to launching `rotclt` and trying to cause an error, which I was able to cause only with the networked model 2402.

I tested with a serial port with nothing attached (less debug messages, but mostly the same):
* [grbltrk-2401-before.txt](https://github.com/user-attachments/files/21951689/grbltrk-2401-before.txt) [grbltrk-2401
* after.txt](https://github.com/user-attachments/files/21951690/grbltrk-2401-after.txt)

and with `nc -l 4500` (less debug messages and manually killing nc at the right moment an IO error, that ironically is was was hardcoded before i changed the code to propagate the error):
* [grbltrk-2402-before.txt](https://github.com/user-attachments/files/21951687/grbltrk-2402-before.txt)
* [grbltrk-2402-after.txt](https://github.com/user-attachments/files/21951688/grbltrk-2402-after.txt)

I thought the `rotctl` would always get an error and exit, but the error is returned and propagated only if `fail_count >= 10`, so I was able to trigger that condition killing nc early enough
`grbl_request read_string fail! (-6) fail_count 11`

The code to fail at the first error was commented out and I didn't change it.

### DIFF
--- a/rotators/grbltrk/grbltrk.c
+++ b/rotators/grbltrk/grbltrk.c
@@ -155,7 +155,7 @@ grbl_request(ROT *rot, char *request, uint32_t req_size, char *response,
         if (fail_count >= 10)
         {
             rot_debug(RIG_DEBUG_ERR, "%s too much xfer fail! exit\n", __func__);
-            return -RIG_EPROTO;
+            return retval;
         }
 
         rig_flush(rotp);
@@ -198,7 +198,7 @@ grbl_init(ROT *rot)
         if (retval != RIG_OK)
         {
             rot_debug(RIG_DEBUG_ERR, "grbl_request [%s] fail\n", grbl_init_list[i]);
-            return -RIG_EIO;
+            return retval;
         }
     }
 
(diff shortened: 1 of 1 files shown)


N0NB:
Thanks, Daniele.

PR #1872, opened 2025-08-30
Remove unused includes

dforsi:
This PR removes some of the includes that according to `iwyu` aren't used, but not all. It keeps `config.h` `rig_state.h` and some that are conditionally included (eg. on Windows), since `iwyu` replaces the compiler, it gets the preprocessed sources so I'm not sure if it can handle these cases right.

OTH `iwyu` removed some includes from simulators without adding `unistd.h` to `sim.h`, which made the build fail because `write()` wasn't defined.

I built this on linux and cross-built on mingw.

### DIFF
--- a/simulators/sim.h
+++ b/simulators/sim.h
@@ -1,6 +1,7 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <stdlib.h>
+#include <unistd.h>
 
 #include "../src/misc.h"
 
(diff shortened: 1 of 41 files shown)


N0NB:
Thanks, Daniele.

gvanem:
Why not remove `include/hamlib/winpthreads.h` ? 
Neither MinGW-w64, LLVM-MiNGW nor cl/clang-cl needs this AFAICS.

PR #1873, opened 2025-08-30
Fix include config.h

dforsi:
Fix unlikely ambiguity on which `config.h` should be included. There are 2 `config.h` but only the one under `include/` :

```
find -name config.h
./android/config.h
./include/hamlib/config.h
```

before:
```
grep -hEro 'include .(hamlib/)?config.h.' | sort | uniq -c
     19 include "config.h"
      1 include <config.h>
     82 include "hamlib/config.h"
      8 include <hamlib/config.h>
```

after:
```
grep -hEro 'include .(hamlib/)?config.h.' --exclude=*.txt | sort | uniq -c
      1 include "config.h"
      1 include <config.h>
     99 include "hamlib/config.h"
      9 include <hamlib/config.h>
```

The `1 include "config.h"` is in `simts890.c` and will be removed by PR #1872 because it is unused.
The `1 include <config.h>` is in `version_dll.rc` and I'm not sure which separator should be used `\` or `/`.
The `9 include <hamlib/config.h>` are in `bindings/` and `tests/`.

### DIFF
--- a/lib/precise_time.c
+++ b/lib/precise_time.c
@@ -20,7 +20,7 @@
 // along with fldigi.  If not, see <http://www.gnu.org/licenses/>.
 // ---------------------------------------------------------------------
 
-#include "config.h"
+#include "hamlib/config.h"
 #include <errno.h>
 #include <time.h>
 #include <sys/time.h>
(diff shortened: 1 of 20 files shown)


N0NB:
Thanks, Daniele.

PR #1874, opened 2025-08-30
Fix misc.c

dforsi:
This PR:
* replaces a literal `-1` with `-RIG_EINVAL` which has the same numeric value
* prints an error message to `stderr` instead of `stdout` (this means that the message isn't visible unless using `-v`, just like other BUG messages)

I'm using `RIG_DEBUG_BUG` because the function is about to return an error with a "Unhandled host" message that is too generic.
One way to trigger that error is adding trailing garbage to a host:port string:
`tests/rigctl -r "invalid:123:port" -m 2`
`tests/rigctl -r "invalid:123xyz" -m 2`


N0NB:
Thanks, Daniele.

PR #1875, opened 2025-08-30
Use rig_debug() instead of sending error messages to stdout

dforsi:
This PR replaces `printf()` with `rig_debug()` in 3 lines.

### DIFF
--- a/tests/ampctl_parse.c
+++ b/tests/ampctl_parse.c
@@ -1584,7 +1584,7 @@ void list_models()
 
     if (status != RIG_OK)
     {
-        printf("amp_list_foreach: error = %s \n", rigerror(status));
+        rig_debug(RIG_DEBUG_ERR, "amp_list_foreach: error = %s \n", rigerror(status));
         exit(2);
     }
 
--- a/tests/rigctl_parse.c
+++ b/tests/rigctl_parse.c
@@ -2154,7 +2154,7 @@ void list_models()
 
     if (status != RIG_OK)
     {
-        printf("rig_list_foreach: error = %s \n", rigerror(status));
+        rig_debug(RIG_DEBUG_ERR, "rig_list_foreach: error = %s \n", rigerror(status));
         exit(2);
     }
 
--- a/tests/rotctl_parse.c
+++ b/tests/rotctl_parse.c
@@ -1667,7 +1667,7 @@ void list_models()
 
     if (status != RIG_OK)
     {
-        printf("rot_list_foreach: error = %s \n", rigerror(status));
+        rig_debug(RIG_DEBUG_ERR, "rot_list_foreach: error = %s \n", rigerror(status));
         exit(2);
     }
 


GeoBaltz:
No.  In each case, the user has asked for a listing of all models.  If he doesn't get one, he is at least entitled to see the reason why.

Remember, rig_debug() is just what its name implies - a debugging tool for developers, not an error reporting mechanism.

dforsi:
Those look like debug messages: they print the name of a function and they use `rigerror()` that prints the whole trace, even of the internal functions that didn't return an error.

Anyway, there are other usages of `rig_debug()`; maybe the lines that print "...no such token..." should be changed to a plain `printf()`, because if I make a mistake such as `tests/rigctl --set-conf=foo=bar` I get no error message unless I add `-vv`

```
$ grep rig_debug\( -rn tests/rigctl.c tests/rotctl.c tests/ampctl.c

tests/rigctl.c:134:    rig_debug(RIG_DEBUG_VERBOSE, "%s: called\n", __func__);
tests/rigctl.c:185:        rig_debug(lvl, "%s: Network error %d: %s\n", msg, e, (char *)lpMsgBuf);
tests/rigctl.c:190:        rig_debug(lvl, "%s: Network error %d\n", msg, e);
tests/rigctl.c:195:    rig_debug(lvl, "%s: Network error %d: %s\n", msg, e, strerror(e));
tests/rigctl.c:241:    if (err) { rig_debug(RIG_DEBUG_ERR, "%s: setvbuf err=%s\n", __func__, strerror(err)); }
tests/rigctl.c:475:    rig_debug(RIG_DEBUG_VERBOSE, "%s\n", rigstartup);
tests/rigctl.c:478:    rig_debug(RIG_DEBUG_VERBOSE, "rigctl %s\n", hamlib_version2);
tests/rigctl.c:479:    rig_debug(RIG_DEBUG_VERBOSE, "%s",
tests/rigctl.c:522:            rig_debug(RIG_DEBUG_ERR, "%s: no such token as '%s'\n", __func__, mytoken);
tests/rigctl.c:565:            rig_debug(RIG_DEBUG_VERBOSE, "%s: defaulting to RTS PTT\n", __func__);
tests/rigctl.c:654:        rig_debug(RIG_DEBUG_TRACE, "%s vfo_opt=%d\n", __func__, vfo_opt);
tests/rigctl.c:657:    rig_debug(RIG_DEBUG_VERBOSE,
tests/rigctl.c:797:            rig_debug(RIG_DEBUG_WARN, "%s: rig_open again retcode=%d\n", __func__, retcode);
tests/rigctl.c:828:            rig_debug(RIG_DEBUG_ERR, "%s: i/o error\n", __func__);
tests/rigctl.c:834:                rig_debug(RIG_DEBUG_ERR, "%s: rig_close retcode=%d\n", __func__, retcode);
tests/rigctl.c:837:                rig_debug(RIG_DEBUG_ERR, "%s: rig_open retcode=%d\n", __func__, retcode);
tests/rotctl.c:266:    rig_debug(RIG_DEBUG_VERBOSE, "rotctl %s\n", hamlib_version2);
tests/rotctl.c:267:    rig_debug(RIG_DEBUG_VERBOSE, "%s",
tests/rotctl.c:303:            rig_debug(RIG_DEBUG_ERR, "%s: no such token as '%s', use -L switch to see\n",
tests/rotctl.c:380:    rig_debug(RIG_DEBUG_VERBOSE,
tests/ampctl.c:245:    rig_debug(RIG_DEBUG_VERBOSE, "ampctl %s\n", hamlib_version2);
tests/ampctl.c:246:    rig_debug(RIG_DEBUG_VERBOSE, "%s",
tests/ampctl.c:282:            rig_debug(RIG_DEBUG_ERR, "%s: no such token as '%s'\n", __func__, mytoken);
tests/ampctl.c:346:    rig_debug(RIG_DEBUG_VERBOSE,
```

GeoBaltz:
Maybe we need both.  As for the perror/perror2, if we had it to do all over again, perror2 should be perror, and perror should be hl_perror(or some such). Looks like perror2() would be preferable here.

These particular messages are followed by exit(); the lack of output would certainly be noticeable/confusing.

The info messages above (e.g., hamlib_version2, etc.) are where they belong - in the debug output, as they are useful for triage of later errors.  The user errors (CLI typos, etc.) should probably go directly to stderr. The status codes are supposed to be passed up the call stack so the ultimate caller can do the appropriate notification.

In The Best of All Possible Worlds, we should be able to define `rig_debug()` as null, and nobody would notice.

N0NB:
As I look at this PR I'm siding with Daniele.  The use of `rig_debug()` seems quite appropriate especially as it's wrapped in an `if` test based on the return value of *_list_foreach()`.  Also, I think it's appropriate to send such output to `stderr` rather than `stdout` by default.

GeoBaltz:
`rig_debug()` may be alright for the debugging info, but the user ought to get some indication that an error occurred, and why.  With this all he gets is the command prompt, as the next instruction is `exit()`.

N0NB:
George points out what I overlooked last night.  Sorry about that.

Now I must agree with George that the `printf()` is proper here.  It's too much to expect a user upon getting no output to know to add `-vv` and try again.

There might be an argument to be made that output should go to `stderr` rather than `stdout`.  99% of the time it doesn't matter, but there will be that one time where such a message would be expected to appear on `stderr` especially since an exit value of 2 is passed to the shell.

N0NB:
As a compromise, how about replacing each `printf()` with:

``` c
fprintf(stderr, "amp_list_foreach: error = %s \n", rigerror(status));
```

GeoBaltz:
Make that `rigerror2()` and I'm good.

N0NB:
Thanks, Daniele.

Issue #1018, opened 2022-04-29
Enhancement: conditional builds for radio classes


guitarpicva:
Concept is:

./configure --with-yaesu --with-icom

Which would build the main parts required by all and only radio types in the yaesu and icom radio classes.

Hamlib takes an eternity to build (particularly on an SOC) and most users have one or two brands of radios and don't need all the rest.

Also, consider --with-rotator, --with-antenna for those who need them, and then the rest of us can avoid the 'noid.

I don't expect any feedback.  I figure if you find the idea worth doing, you'll just do it. ;)

pvrose:
Who is the user who would use this? I would think that the main users of the compile flow are going to build hamlib into their generic application, and therefore want all backends.

The real end-user with one or two rigs will then use the generic application which will have hamlib already built-in for them.

N0NB:
I agree, Phil.

We're currently at 40 radio backends and adding configure conditionals for each would make our already quite large `configure.ac` even larger and less maintainable.  Then there is the question of whether the conditional options to enable all backends would at some point exceed the maximum command line length on some OS.

That said, I understand the time it takes to build current Hamlib, especially on lower end hardware.  A better solution, IMHO, would be to cross-compile on a faster host for the target.  As such recipes are developed we can add them to our documentation.

guitarpicva:
Using cmake, there could be CMakeLists.txt made up for major groups.  

ex: mkdir build && cd build && cp ../CMakeListsYaesu.txt ../CMakeLists.txt && cmake ../

No command line issues.

cmake is another "undertaking" the team may not wish to convert to.  that's a value judgement as well.

N0NB:
cmake has come up on the mailing list in the past.  Long ago before the library was this sophisticated I looked into it and redoubled my efforts on understanding the Autotools better.  A few years back a proposal floated through to convert to cmake.  My strong preferences as noted then have not changed, all current environments need to be built and the capability to distribute a source tarball as we do now that does not require the user to install cmake.  So far I've not seen any reference to any work that has been done by the proponents.

I've built cmake projects and I've nothing against it.  I'm not sure that cmake can generate a source archive like the Autotools can where all that is required is a shell to run `configure`, `make`, and any optional build dependencies.  It may be possible, but so far whenever I've encountered a cmake project I've had to have cmake installed.

[...]

PR #1876, opened 2025-08-30
Make POSIX threads mandatory

GeoBaltz:
Part of issue #1772 

As I was finishing this, I had to wonder - Do we really want to remove all of the '#if's?  Would they be useful if someone wanted to build an embedded version of Hamlib, perhaps along with #1018?

What say ye all?

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -14,6 +14,9 @@ Version 5.x -- future
 
 Version 4.7.0
         * 2025-12-01 (target)
+	* POSIX threads are required to build and run Hamlib. Note that it was
+	  actually the case for 4.6.x, but now the configuration step will
+	  fail instead of the compilation.
 	* Functions rig_get_conf, rot_get_conf, amp_get_conf deprecated;
           use *_get_conf2() instead.
         * Fix handling of unprintable characters in kenwood.c that broke radios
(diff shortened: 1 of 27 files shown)


dforsi:
Linux, Mac, Windows and supporting old and new radios is already a lot of work. Adding embedded systems is a risk, it would be nice, but we would need to state which embedded targets are supported, so that they can be tested for breakages, and maintaining compatibility takes time.

Something like a Raspberry Pi Zero gives a small form factor with hopefully less work to Hamlib developers. So I'd rather invest time in Linux Arm instead, should it need fixes.

People are welcome to send patches to make some parts Hamlib configurable, such as #1018 or to reduce memory usage or something else if it doesn't increase complexity too much for developers and users.

IMHO this PR could be merged.

N0NB:
Years back when I got an RPi B, I loved to torture it by compiling Hamlib.  I then moved on to Qemu images but I learned that unless doing a specific development/test for GPIO that there really wasn't any point in it as GNU is GNU.

I've long since thought that support for such systems is best done by cross-compiling.  As doing so saves installing the required tool chain and dependencies on the device and saves wearing of the (often, not always) SD/micro-SD card which really aren't designed for the heavy write cycles of compilation.

Having lost about 16 months worth of WeeWX weather data due to a dead SD card years back, I don't view them as reliable.  It's likely that OS have gotten better at minimizing writes but writing to a database or compilation will have its detrimental effect on the card.

N0NB:
Thanks, George.

PR #1878, opened 2025-08-30
Do not hardcode host and port number of INDI server

dforsi:
This allows to connect to the server running on another machine or on a different port.
Defaults to the old value of localhost:7624.
It is still possible to connect to the local domain, eg. with --rot-file @/tmp/indiserver


N0NB:
Your comments should be documented in the `rotctl[d]` man page.  Maybe a special section on INDI needs to be added detailing what it is and how to use it?

N0NB:
Are the `host` and `port` buffers at `256` and `6` respectively, sufficiently large?  While TCP ports at present are limited to 5 characters, I wonder if some future revision would raise that value?  Would `1024` and `16` be too large?

dforsi:
In this case `256` and `6` are ok because `parse_hoststr()` uses `255` and `5` and `sscanf()` terminates the string so it needs just an extra byte:
`n = sscanf(hoststr, "%255[^:]:%5[0-9]%1s", host, port, dummy);`

I added a one-sentence description of INDI, I don't know what to write. I see that there is a page in the wiki that could host more information: https://github.com/Hamlib/Hamlib/wiki/Rotator-Specific-Notes but still I don't know what to write.
The man page in the COLOPHON says that a wiki exists but it isn't clear what information could be found there.

N0NB:
Thanks, Daniele.

Issue #662, opened 2021-04-14
twiddle_rit not working


mdblack98:
twiddle_rit needs to be completed.

PR #1879, opened 2025-08-31
Fix definition of twiddle_timeout in struct confparam

dforsi:
Fixes the definition of config parameter `twiddle_timeout` and the description of `rig_set_twiddle()`.

This sets the maximum value to 10 seconds, however `rigctl` doesn't check the range so it can be set higher.

The value is in seconds with a minimum of 1, I wonder if a lower value would be better if this gets set by software; issue #662 says that `twiddle_timeout` needs to be completed so perhaps this can be changed to be milliseconds. How to test it?


N0NB:
Thanks, Daniele.

PR #1883, opened 2025-09-02
Baseline for ReleaseNotes (4.7 & 5.0)

GeoBaltz:
For most project releases, the release notes are a last minute chore that sees a lot of history searching to find out what happened. So I figured let's do something completely bass-ackwards and start them months before they're needed.  Or years.

These are meant as starting points and possibilities.  They are subject to change without notice, but eventually they should agree with the final product. They may even help decide what that final product is.

They do have more detail for things I have been working on - feel free to add your own.

Note that these are NOT meant to replace NEWS - that file should be the devel log, whereas RelNotes should be much more aimed at the end user.

### DIFF
--- a/ReleaseNotes_4.7.md
+++ b/ReleaseNotes_4.7.md
@@ -46,7 +46,7 @@ Along with the moves to separate storage, the definitions of these data structur
 will be moved out of hamlib/rig.h to their own include files. These `.h` files also
 define macros to get the address of said structures.
 
-Prelinary versions of these files are also part of 4.7 for pre-emptive use.
+Preliminary versions of these files are also part of 4.7 for pre-emptive use.
 
 ### The good news
 If your application only calls the Hamlib API routines, then nothing needs to change.


N0NB:
An idea I am toying with are the gnulib [VCS To ChangeLog](https://www.gnu.org/software/gnulib/manual/html_node/VCS-To-ChangeLog.html) and  [gitlog-to-changelog](https://www.gnu.org/software/gnulib/manual/html_node/gitlog_002dto_002dchangelog.html) modules and see how they compare to `git2cl` which is rather poor, for populating the ChangeLog file before releases.

I will have a look at this PR in depth tomorrow morning.  I like the idea.

GeoBaltz:
I've always thought the full CMS/VCS log or re-hashes thereof was much too verbose for documentation.  And with git it's already available in the repo.

N0NB:
The *VCS To ChangeLog* module claims that it operates on a range of `<from_ref> <to_ref>` where this would limit the output, say, from the commit that creates a branch until the tag commit created just before release.  If the output format is acceptable, I would hook this into the `dist` target so it would be included in the generated source archive and not ever held in the repository.  Also, it is claimed to only recognizes changes to C sources.

This idea might not see the light of day.  It's just something I'm considering looking at.

N0NB:
Line 49 of `ReleaseNotes_4.7.md` change to "preliminary".

N0NB:
For Markdown and text files I've come to prefer tabs as spaces and a text width of 80 characters.  In my `$HOME/.vim/after/ftplugin/markdown.vim` I have:

``` vim
" Set 4 space "tabs" in Markdown.
set shiftwidth=4 softtabstop=4 expandtab

" Set text width to 80 characters
set textwidth=80
```

I think I'll put such comments into `README.coding_style`.

GeoBaltz:
Since this was a practice session for learning Markdown, I did this with emacs, ghostwriter, and even nano.  Spaces were the common tool. I haven't worried too much about 80 character lines since I stopped using punch cards :-); usually I have a window about 100 wide and let it flow.

N0NB:
My window is much wider than 80, but I think it is a nice value that aids in reading.  As an example, I find that limiting manual pages to 80 characters aids in reading.  Likewise I find Web pages that have text going from border to border more difficult reading than ones where the text is limited.

Maybe that's just me.  😂

N0NB:
Thanks, George.

GeoBaltz:
I find 80 to be too limiting - maybe I'm still used to 132 character line printer output, 80 lines per page.  I like to get as much as possible on a screen; blank lines also seem like a waste of real estate in many cases.

I do think man pages look nicer when fit to 8.5x11 form factor; I have been known to refer to printed copies ;-)

To each his own.  I'll try to remember to format appropriately.
73

GeoBaltz:
One more comment - like all Markdown/HTML, the rendering can vary by engine.  When I was writing this, the output of ghostwriter looked great, okular was so-so.  Github is somewhere in between.

N0NB:
I was actually looking over `README.coding_style` yesterday before you submitted your PR as I was thinking over adding a section Markdown.  I forgot to PR the changes and just uploaded it.  The suggestions are flexible.

I have a Firefox plugin to view them but it doesn't work as nicely as it used to.  Progress.

Issue #1881, opened 2025-09-02
rigerror2() should be documented - and perhaps given a better name


peterbmarks:
We were looking for a less intimidating error message to show users than the error plus stack trace that rigerror() returns and was informed of rigerror2().

This is exactly what we want but is not documented in the rig API https://hamlib.sourceforge.net/manuals/4.3/group__rig.html

It might be good to give these better names, perhaps rigerror_stack_trace() and rigerror_user_message().

Peter
VK3TPM

N0NB:
I've uploaded a recently generated manual for 4.7~git to https://hamlib.sourceforge.net/manuals/4.7/

`rigerror2` is in there, but not well documented at all.  I'm working on that.

N0NB:
I've uploaded an updated 4.7 manual generated by Doxygen 1.14 with the changes in PR #1889.

dforsi:
Since the same error numbers and strings are used also for amplifiers and rotators, if the name is changed, I suggest using `hamlib_` as prefix instead of `rig`.

N0NB:
That's likely a sensible idea, Daniele.  Perhaps a link to the old name should remain at least for one major release, something like:

``` c
typedef hamlib_error_full rigerror;
typedef hamlib_error_simple rigerror2;
```

peterbmarks:
That would be fine although I would add that the "simple" error string should actually be the primary case and the "full" version with the stack trace should be used by developers. Is there a DEBUG #define used somewhere?
Thanks for documenting rigerror2().

PR #1889, opened 2025-09-06
Update Doxygen comments for rigerror() and rigerror2()

N0NB:
Add a new topic of "Hamlib internals" with the group of lib_internal.

Address GitHub issue #1881.

@GeoBaltz Please check this for accuracy.


GeoBaltz:
Your changes looks good to me.  There are a couple of helpers from a while back that now can be added to that category.

The only thing I noticed is that some of the current code is a little shaky vis-a-vis thread safety.

PR #1890, opened 2025-09-07
Fix typo

dforsi:
There were two 'R' instead of one in `DECLARE_INITROT_BACKEND`, however the output of `./cppcheck.sh rotators` is the same because the drivers include `register.h` that defines that macro.


N0NB:
Thanks, Daniele.

PR #1893, opened 2025-09-07
lib/precise_time.c: Remove dead code

df7cb:
The last branch of `#if _POSIX_TIMERS` was missing a declaration of "showme". Since that variable is not used anywhere, just drop it.

Discovered because the hurd-i386 Debian build was failing: https://buildd.debian.org/status/fetch.php?pkg=hamlib&arch=hurd-i386&ver=4.6.5-1&stamp=1757241757&raw=0

### DIFF
--- a/lib/precise_time.c
+++ b/lib/precise_time.c
@@ -53,15 +53,8 @@ double monotonic_seconds()
 
 #include <time.h>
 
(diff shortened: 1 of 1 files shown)


N0NB:
Thanks, Christoph.

PR #1902, opened 2025-09-09
Handle --amp-file and --rotfile for pytests

dforsi:
This PR allows to run some tests with different drivers using the well-known `--model` argument.

Examples:
```
PYTHONPATH=bindings/:bindings/.libs bindings/python/test_amp.py --model 2
PYTHONPATH=bindings/:bindings/.libs bindings/python/test_rot.py --model 2
```
As for the rigs, at this time only the simpler tests that do not open the device are enabled for all models; the tests that open the device are enabled only for the dummy model number 1 for two reasons: they would probably fail because some arguments aren't meaningful for actual devices and there is also the risk of activating functions that the user didn't want.

### DIFF
--- a/bindings/python/conftest.py
+++ b/bindings/python/conftest.py
@@ -10,7 +10,7 @@ def pytest_addoption(parser):
                         metavar='ID', help='select amplifier model number')
         parser.addoption('--amp-file', default=None,
                         metavar='DEVICE', help='set device of the amplifier to operate on')
-    if sys.argv[1].endswith("rig.py"):
+    elif sys.argv[1].endswith("rig.py"):
         parser.addoption('--model', type=int, default=1,
                         metavar='ID', help='select radio model number')
         parser.addoption('--rig-file', default=None,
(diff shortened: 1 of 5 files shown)


dforsi:
These arguments allow to do some tests on all devices.
For example, I found the segfault in a rig driver reported in issue #1905 using bash to extract all model numbers and run the tests:

```
export PYTHONPATH=bindings/:bindings/.libs

for MODEL in $(tests/ampctl -l | grep -E "^[ 0-9]+ " -o); do cmd="bindings/python/test_amp.py --model $MODEL"; echo $cmd; ./$cmd >/dev/null; done
for MODEL in $(tests/rigctl -l | grep -E "^[ 0-9]+ " -o); do cmd="bindings/python/test_rig.py --model $MODEL"; echo $cmd; ./$cmd >/dev/null; done
for MODEL in $(tests/rotctl -l | grep -E "^[ 0-9]+ " -o); do cmd="bindings/python/test_rot.py --model $MODEL"; echo $cmd; ./$cmd >/dev/null; done
```

N0NB:
Thanks, Daniele.

PR #1908, opened 2025-09-13
N0NB issue 1892

N0NB:
Trying after rebasing to master.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -347,39 +347,6 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[]],
 "Have you considered GCC lately?."])
 		  ])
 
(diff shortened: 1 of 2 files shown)


N0NB:
@dforsi If you look at this failed workflow run:

https://github.com/Hamlib/Hamlib/actions/runs/17694420185/job/50292203944?pr=1908

You'll see it fails to build `rigtestlibusb` due to `configure` not finding `libusb.h`.  I want you lo look at the workflow file, `https://github.com/Hamlib/Hamlib/blob/master/.github/workflows/macos-latest-features.yml` and tell me where I got this wrong.  When the `make discheck` step fails, it fails to run the remaing steps.  Does this need to be broken out into its own step in some way so the tar file will be generated and uploaded even with a failure?

Issue #1892, opened 2025-09-07
Configure script fails to detect libusb-1.0 on FreeBSD - hardcoded library name issue


techn0mad:
**Describe the bug**

- FreeBSD provides libusb-1.0 API through `-lusb` library name, not `-lusb-1.0`
- Configure script at line ~23326 hardcodes test for `-lusb-1.0` which fails on FreeBSD
- `pkg-config` correctly reports the library as `-lusb` but configure ignores this
- One workaround requires creating symlinks to existing libusb libraries: `ln -sf /usr/lib/libusb.so /usr/lib/libusb-1.0.so`, etc.

**To Reproduce**
1. `./bootstrap`
2. Run configure and examine the configure log for libusb issues:
```
$ ./configure --prefix=$HOME/local --with-python-binding --enable-pytest > /dev/null 2>&1
$ grep -A 20 -B 5 "libusb_init.*lusb-1.0" config.log
configure:23279: cc -c -g -O2   conftest.c >&5
configure:23279: $? = 0
configure:23281: result: yes
configure:23292: checking whether to build USB dependent backends
configure:23306: result: yes
configure:23326: checking for libusb_init in -lusb-1.0
configure:23355: cc -o conftest -g -O2    conftest.c -lusb-1.0   >&5
ld: error: unable to find library -lusb-1.0
cc: error: linker command failed with exit code 1 (use -v to see invocation)
configure:23355: $? = 1
configure: failed program was:
| /* confdefs.h */
| #define PACKAGE_NAME "Hamlib"
| #define PACKAGE_TARNAME "hamlib"
| #define PACKAGE_VERSION "4.7~git"
| #define PACKAGE_STRING "Hamlib 4.7~git"
| #define PACKAGE_BUGREPORT "hamlib-developer@lists.sourceforge.net"
| #define PACKAGE_URL "http://www.hamlib.org"
| #define HAVE_STDIO_H 1
| #define HAVE_STDLIB_H 1
| #define HAVE_STRING_H 1
| #define HAVE_INTTYPES_H 1
| #define HAVE_STDINT_H 1
| #define HAVE_STRINGS_H 1
| #define HAVE_SYS_STAT_H 1
| #define HAVE_SYS_TYPES_H 1
```

**Expected behavior**
configure runs to successful completion and Hamlib is ready to be compiled on the FreeBSD.

**Desktop (please complete the following information):**
 - OS: FreeBSD 14.3-RELEASE-p2
[...]

Issue #1870, opened 2025-08-26
Split GitHub workflow script into multiple scripts


N0NB:
Currently `.github/workflows/c-cpp.yml` run Continuous Integration (CI) on Ubuntu "latest" and MacOS "latest".  While this works and most configuration settings are shared, there are exceptions, particularly for MacOS which makes its stanzas a bit less straightforward than they could be.

I floated the idea to @dforsi and @markjfine in PR #1862 and both thought there is some merit to the idea.  [Daniel commented as follows](https://github.com/Hamlib/Hamlib/pull/1862#issuecomment-3222094525):

> I think that's a good idea because we can see if a failure is only in one OS. It will also make it easier to add another workflow for Windows or Raspberry or something else. In the current workflow I like the sharing of the `--with` and `--enable` options, while overriding values for some steps, I don't know how to share them between different scripts. 

> I see that workflows can be executed manually with the trigger `on workflow_dispatch`, this can be useful to test the split scripts before having them run automatically, there isn't much documentation because it seems to work similar to `on push`/`on pull_request`: 

> https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/
> https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax?s=09#onworkflow_dispatch

With this approach I presume that each build will be self-contained with some apparent duplication between scripts.  This really isn't an issue as these files just sit in the repository and are not a part of the source distribution.

N0NB:
I have pushed a pair each of Ubuntu and MacOS workflow files.  At the moment each pair consists of a build with features and no features with the configure options from `c-cpp.yml`.

To cut down on used resources and maybe speed up execution time they only do `make dist`.

To improve things each pair should be merged with both features and no-features for each OS.  I ran into a shortage time to figure out how to give the resulting `tar` files unique names.

It seems GitHub only runs a new workflow file if it exists in the master branch (once there it can be changed in a feature branch and it will be run, at least it did for me).  Constantly pushing changes that didn't work was getting tedious so I found a program called [act](https://github.com/nektos/act) that allows for mimicking the GitHub workflow system locally using a Docker image.  Unfortunately, only a rather old version of Ubuntu is available, so MacOS can't be tested locally.

Anyway, they are there now and maybe @dforsi can tweak them up.

dforsi:
Now the workflows are run for any branch not only for `master` or `feature` (PR #2179 and #2180)

PR #1911, opened 2025-09-14
Fix setting the variable HAVE_LIBUSB for makefiles by adding a missing AC_SUBST

dforsi:
This fixes a build failure when running `./configure --with-libusb' (which is the default when no libusb argument is given) and the needed files weren't found.

The relevant fragment of `configure` output looks like this:
```
checking whether to build USB dependent backends... yes
checking for libusb-1.0... yes
checking for libusb.h... no
checking for libusb-1.0/libusb.h... no
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -357,7 +357,6 @@ AC_ARG_WITH([libusb],
     )
 
 AC_MSG_RESULT([$cf_with_libusb])
-AM_CONDITIONAL([HAVE_LIBUSB], [test x"${cf_with_libusb}" = "xyes"])
 
 LIBUSB=""
 AC_ARG_VAR([LIBUSB_CFLAGS], [C compiler flags for libusb, overriding configure defaults])
@@ -376,6 +375,9 @@ AS_IF([test x"${cf_with_libusb}" = "xyes"], [
 		   [1],
 		   [Define if libusb-1.0 is available])
 	LIBUSB="libusb-1.0"])])
+
+AM_CONDITIONAL([HAVE_LIBUSB], [test x"${cf_with_libusb}" = "xyes"])
+AC_SUBST([HAVE_LIBUSB])
 
 # Only used in hamlib.pc.in
 AC_SUBST([LIBUSB])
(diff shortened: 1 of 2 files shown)


dforsi:
Not sure what changed in the workflow, now the output messages are different.

The Mac workflow that is passing with this PR has:
```
checking whether to build USB dependent backends... yes
configure: WARNING: libusb_init was not found in libusb-1.0--USB backends will be disabled
checking for libusb_init in -lusb-1.0... no
```
and the summary:
```
Enable USB backends 	    no
```
https://github.com/Hamlib/Hamlib/actions/runs/17713000840/job/50334205542?pr=1911#step:9:229

The workflow that failed in PR #1908 had:
```
checking whether to build USB dependent backends... yes
checking for libusb-1.0... yes
checking for libusb.h... no
checking for libusb-1.0/libusb.h... no
```
and the summary:
```
Enable USB backends 	    yes
```
https://github.com/Hamlib/Hamlib/actions/runs/17694420185/job/50292203944?pr=1908#step:7:549

N0NB:
I didn't change anything in the c-cpp.yml file.  I added four more workflow files as noted in issue #1870.  They likely could use attention from a more experienced eye.  As I had to upload them to `master` for GitHub to recognize them, they are being run with every PR since.

I also have PR #1908 where I modified `condigure.ac` to use `pkg_config` to search for `libusb` that is likewise failing on MacOS to find `libusb.h` in both the present `c-cpp.yml` and the dedicated MacOS CI workflow files.

I'm attempting to get a VM with MacOS going.  So far the first method has been failing and I've now found a Docker method that I'll try next.

N0NB:
To follow up.  I think we need to use `pkg_config` to detect the presence of `libusb` as the current detection is not working on that OS (issue #1892).  After making sure it worked on Linux and FreeBSD, it then failed MacOS.  Sigh...

dforsi:
Yes, `pkg_config` could make the detection more portable.

I think that this PR can be merged because the missing `AC_SUBST` would be an issue also when using `pkg-config` and the message in the summary would be confusing when it prints `yes` but later code changes the variable to `no`.

N0NB:
Thanks, Daniele.

I'll rebase this into my branch working on issue #1892 and see how things go.

PR #1912, opened 2025-09-14
Remove unneeded readline libraries

dforsi:
The only sources that use readline functions are:
$ grep -l readline tests/*.c
tests/ampctl.c
tests/ampctl_parse.c
tests/rigctl.c
tests/rigctl_parse.c
tests/rotctl.c
tests/rotctl_parse.c


N0NB:
What will be the replacement for Readline?

N0NB:
Oops, sorry, I misread the PR description, posted and then looked at your patch.  None of those three are interactive (I've never tried using any of them)?

dforsi:
Yes those 3 aren't interactive, but it turns out that they link to the `.o` compiled from `*_parse.c` that needs to be linked to libreadline on Github, while it works for me on Debian native and with `act` on Ubuntu (thanks @N0NB).

N0NB:
I think this PR needs to be dropped.  I get this:

```
  CCLD     rigctlcom
/usr/bin/ld: rigctlcom-rigctl_parse.o: in function `rp_getline':
/home/nate/git/hlb/tests/../../hamlib/tests/rigctl_parse.c:541: undefined reference to `readline'
/usr/bin/ld: rigctlcom-rigctl_parse.o: in function `rigctl_parse':
/home/nate/git/hlb/tests/../../hamlib/tests/rigctl_parse.c:1730: undefined reference to `add_history'
collect2: error: ld returned 1 exit status
make[1]: *** [Makefile:1151: rigctlcom] Error 1
make: *** [Makefile:666: all-recursive] Error 1
```

While it might be a bit of wasted bytes in the `Makefile`, having an unused library on the linker command line is harmless, as I understand it.  Here it is clearly required due to the dependency on `rigctl_parse.o` as you note.

PR #1917, opened 2025-09-18
Fix build

dforsi:
This PR fixes build errors when running `make` from inside the `simulators/` or `tests/` directories.
In my tests running `make` inside the `c++/` never failed, however the rule for build `libmisc.la` was missing, so I'm adding it there too.


N0NB:
Thanks, Daniele.

Issue #1813, opened 2025-07-24
SDR#/gpredict


markjfine:
Do we know if the current SDR#/gpredict backend is working with SDR# 1921?

Was just testing a mingw64 build of Smart Gnome Control on Win11 and was unable to connect to SDR# using the their gpredict plugin sitting on 127.0.0.1:4532. Attempting to connect gives this error:
<img width="589" height="720" alt="Image" src="https://github.com/user-attachments/assets/6a133d08-e4b2-482c-9f3d-fdd62ab4b118" />

However, if I keep that connection set and switch to one of the two Dummy backends the gpredict module in SDR# shows as being connected (although none of the commands do anything). So it appears as if the port isn't firewalled, just a possible issue with the SDR#/gpredict backend itself.

markjfine:
So I just cleaned up the sdrsharp backend to look like a limited version of GQRX, but only with a get/set_freq capability. Still receiving the same kinds of errors.

So, I'm going to guess that SDR#'s gpredict plugin broke when Youssef moved away from Telerik last year.

N0NB:
Is there a way to communicate without the gpredict plugin?

markjfine:
Don't think so, since the plugin is what was sitting on the port to read command traffic for SDR#.

That said, the plugin hasn't been updated since 2018, and I know that Youssef's completely redone SDR# in the past year or so. It might be easier for me to just rebuild the plugin using an updated SDK (if it's available).

Edit: I should say that the sitting on the port and detecting a connection works, just not any commands to SDR#

markjfine:
Following up: Finally getting back to looking into this after being diverted by other things (then forgetting about it 🤦🏻‍♂️). I contacted Youssef Touil some time ago to find if anyone has reported a problem with this plugin. He recommended running it through their plugin SDK, which I am finally getting to. More to come.

PR #1919, opened 2025-09-19
Update to SDR#/gpredict backend

markjfine:
Ref Issue #1813.
Circling back on this backend update. Still haven't gotten SDR# to talk to it in Windows. However, the backend does work in SDR++ (using the same port# no less). Pretty sure the SDR# issue is with its interface to the gpredict plugin. It will listen and recognize the port connection but not handle commands, causing the backend to think there's a connection failure. The backend is fine, SDR# is broken.


N0NB:
I think all of the merge commits can be avoided by the use of `git rebase master` in your feature branch.  But it does make for a lot of colored lines in `tig`!

markjfine:
Lol... I think was trying to synch up my fork with the fork master after I had synched that from the official master. Eventually realized it was just easier and cleaner just doing it from the button on the github web site. It got further complicated because I also have two locals of everything - macOS and Fedora - the Fedora one being used to cross-compile to Windows/SDR#.

N0NB:
Thanks, Mark.

markjfine:
Any time.

markjfine:
I've already told Youssef that I'd look at the plugin and see what the problem is there, but I keep getting sidetracked.

GeoBaltz:
Another request to do as Nate mentioned, and use `git rebase master`.  All the pretty colored lines in gitk are hard to follow for the other commits, besides the extra noise in the commit message.

73

N0NB:
I probably could have rebased it before merging into master but then GitHub doesn't see that it was merged.  This project isn't known for a straight line of history anyway!

markjfine:
I probably misunderstood this. I thought those merges would be transparent because the only difference between the two baselines at this point was only the one file (sdrsharp.c).

GeoBaltz:
The end result is the same, but each of your merges copied the master into your branch (and your history).  Then when Nate did the merge, all of the histories got traced back to their own origin and were listed twice - once for the master, and once in its respective merge.  It also makes the pull request(and the mail message) look a lot hairier than it really is.

Rebasing(if no conflicts) just moves the pointer to where the base of your code starts; merging into your branch does a full copy into that branch.

(sorry if I got carried away)

markjfine:
@GeoBaltz 
Good explanation. Apologize if I made a mess. If I have to sync a feature branch in the future I'll be sure to flatten it locally with a rebase and push it back to github before submitting a PR. Hopefully I won't take 2 months to test a feature where it gets synced to this extent. This was kind of an unusual case, where I wanted to be absolutely certain the problem was with a 3rd party application before submitting, and there were no repercussions from changes made elsewhere within Hamlib.

PR #1920, opened 2025-09-21
Add more function to bindings

dforsi:
This PR adds more function bindings to the Hamlib and Rig classes and increases test coverage.


N0NB:
Thanks, Daniele.

PR #1921, opened 2025-09-21
Add yet more functions to bindings 

dforsi:
This PR adds more function bindings to the Amp and Rot classes and increases test coverage.

The average function coverage is now above 50% and the line coverage is above 30%.


N0NB:
Thanks, Daniele.

PR #1922, opened 2025-09-24
Remove duplicated define

dforsi:
`ROT_MODEL_RT21` is defined with the same value two lines before that one.


N0NB:
Thanks, Daniele.

Issue #1923, opened 2025-09-25
TT599 Eagle does not return passband control to the knob


peterdulowsky:
**Describe the bug**
TT599 eagle does not return passband control to the knob

**Additional context**
Code:       `        SNPRINTF(mdbuf, sizeof(mdbuf), "*R%cM%c" EOM "*R%cF%d" EOM "R%cF0" EOM,
                 ttreceiver,
                 ttmode,
                 ttreceiver,
                 (int)width,
                 ttreceiver
                );
`
is missing asterisk before R in "R%cF0"
Correct to:
`        SNPRINTF(mdbuf, sizeof(mdbuf), "*R%cM%c" EOM "*R%cF%d" EOM "*R%cF0" EOM,
                 ttreceiver,
                 ttmode,
                 ttreceiver,
                 (int)width,
                 ttreceiver
                );
`

PR #1924, opened 2025-09-25
Fix command to put bandwidth control back to bandwidth knob

dforsi:
Adds a missing *. Thanks to @peterdulowsky.

Fixes issue #1923.


N0NB:
Thanks, Daniele.

PR #1926, opened 2025-09-27
N0NB issue 1895

N0NB:


GeoBaltz:
Question: does this fix the annoyance that running 'make distcheck' in a build subdirectory creates hamlibdatetime.h in the main source?

N0NB:
> Question: does this fix the annoyance that running 'make distcheck' in a build subdirectory creates hamlibdatetime.h in the main source?

No.  What I can figure is that with out-of-tree builds, the `dist/distcheck` options copy it there to create the source distribution.  I think that is just inherent to its operation.  But that is okay as `.gitignore` already takes it into account and a `git clean -dxf` will remove it.

GeoBaltz:
OK.  It's just an annoyance because the build date in the debug log header doesn't change after that.  I just have to remember to `rm ../src/hamlibdatetime.h`

N0NB:
Adding it to a `clean` rule wouldn't be wise as it would probably get deleted when `make clean` is run.  There are other "clean" rules that do various levels of deletion.  Perhaps it would fit there?

I really don't like it being in `$(builddir)/src`  I would rather it be in `$top_builddir/include/hamlib` but that is an exercise for another time.

PR #1930, opened 2025-10-01
Dummy Rotator Fix Move Command

dgbalharrie:
Use rot->caps min_az, max_az, min_el, max_el instead of numbers. Fix not being able to move beyond 180 using move CW command.


N0NB:
Thanks, David.

PR #1932, opened 2025-10-03
Move fifo stuff out of rig.h

GeoBaltz:
Move FIFO_RIG definition out of rig.h and into fifo.h.

FIFO_RIG is strictly internal to Hamlib and shouldn't be exposed to the app.  Move definition to src/fifo.h.  Also add
'hl_' prefix to internal routine names to avoid conflict with app.

This also adds a dummy definition of spaces(), since it is used by rig_debug() in rigctl/rigctld/<others> even though
it is not part of the API.  Another landmine in the road to a  stable yet flexible interface.


N0NB:
Thanks, George.

PR #1935, opened 2025-10-08
Various fixes, mostly for error handling

GeoBaltz:
Don't let one error lead to others - don't call routine on uninitialized buffer, don't go on if we get bad input, don't forget to clean up if we're exiting early, yada, yada.

Fix bad use of string literal.


N0NB:
Thanks George.  Apologies for being tardy as it's harvest season and that leaves little free time unless it rains, it seems.

PR #1948, opened 2025-11-10
Minor changes

GeoBaltz:
Minor optimization, and some typos.  The VFO changes were noticed while I was looking for all combinations of VFO configurations.  These didn't really make sense; looked and found some were using modes where vfos belonged.


N0NB:
Thanks, George.

PR #1952, opened 2025-11-19
Repair and revise Kenwood voice memory handling

GeoBaltz:
This series was born out of a very disappointing Sweepstakes weekend, when my plans were beset by both software and hardware gremlins.

First were errors trying to access the voice memories from Hamlib - that needed the enabling 'PB01' command restored for the newer rigs.  Then I could hear the messages, but they weren't being transmitted; and some time during this phase the headset mike stopped driving the transceiver.  Aaaargh!  I had visions of the TS-890 looking for input on the wrong port, blown input drivers, all sorts of horrible things.  At this point I game up and took the XYL out for dinner.

After dinner I started a slightly more methodical triage of the problems.  The mike turned out to be simple - a bad case of cruddy contacts.  Unplugging and plugging a couple of times got it working.  The audio memory not being transmitted was a little more subtle - the 890 was doing exactly what Hamlib asked it to do - playback the selected memory; i.e. feed it to the audio output of the rig for QA.  After reading the manual VERY carefully, changing one character (1 -> 5) made it work.

By then I had noticed a few things that were missing or just plain ugly, so set about fixing them or making them pretty. And here is the result.  Sorry for being so long-winded.

### DIFF
--- a/doc/man1/rigctl.1
+++ b/doc/man1/rigctl.1
@@ -760,7 +760,7 @@ Have rig transmit internal message
 .RI \(aq Msgnum \(aq
 .
 .TP
-.BR 0xab ", " stop_voice_mem "
+.BR 0xab ", " stop_voice_mem
 Stop transmision of internal message
 .
 .TP
--- a/doc/man1/rigctld.1
+++ b/doc/man1/rigctld.1
@@ -779,7 +779,7 @@ Have rig transmit internal message
 .RI \(aq Msgnum \(aq
 .
 .TP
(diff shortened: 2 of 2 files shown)


N0NB:
I much rather see a thorough explanation than a few words.  Git places no limit on the size of a commit message, so message away!

No SS for me this year as too much farm work needed doing.  Here's hoping I get a good chance at ARRL 160 in a couple of weeks.

N0NB:
Thanks, George.

I've been intrigued by Not1MM but haven't tried it yet.

PR #1957, opened 2025-12-06
Remove last vestiges of non-ANSI c

GeoBaltz:
Get rid of non-ANSI C support

For some reason, HAMLIB_PARAMS bothered me since the first time I looked at Hamlib code.  It was probably that similar things were prevalent when I first became a full-time UNIX SysAdmin(1992), and were mostly gone by the time I retired(2007).  It was like I had regressed 30 years.

There may be more to enforce `-std=c11`, but at least this won't bug me anymore.

### DIFF
--- a/include/hamlib/rig.h
+++ b/include/hamlib/rig.h
@@ -126,14 +126,10 @@
 #endif
 //! @endcond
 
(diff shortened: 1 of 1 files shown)


GeoBaltz:
Obviously master only
Issue #1773

N0NB:
Thanks, George.

And onward we go!

Issue #1653, opened 2025-01-07
Hamlib handles FLRig modes incorrectly


radio-miskovice:
Hamlib handles information received from FLRig in an incorrect way.

When I connect QRP Labs QMX transceiver to js8Call (which uses Hamlib to control CAT, even to when accessing through FLRig), I can see two types of errors reported from Hamlib to stderr:

```
Hamlib: flrig_openL Unknown mode (new?) for this rig='DIGI-L'
Hamlib: flrig_openL Unknown mode (new?) for this rig='DIGI-U'
```

The connected rig, QMX, has four possible modes (as of fw 1.0.27): CW, CW-R, DIGI-U and DIGI-L. These are mapped to MD3, MD7, MD6 and MD9 in Kenwood-like command set. The reason why QMX reports DIGI-U and DIGI-L instead of the conventional FSK ans FSK-R is probably due to the fact that QMX is a SDR rig, mostly used for FT8, JS8 and similar modes (single-frequency two-tone FSK or multi-tone MFSK) and has nothing to do with the archaic method of direct keying two-tone FSK. 

So my first objection to Hamlib behavior is that it should accept mode name as-is as it is received from FLRig. It is not up to Hamlib to invent or enforce this. This is correct and true information provided by FLRig.

Later on, Hamlib reports every couple of seconds the following:
```
Hamlib: modeMapGetFLRig: FlRig does not have mode: USB
```
Of course, this is true because QMX indeed has no USB mode (nor it has LSB for that matter), it has DIGI-U instead. This is another thing that Hamlib inapproprietly meddles in - it reports to JS8Call that USB mode is available on the transceiver for JS8 communication, but this is not true, and so when JS8Call requires - via Hamlib to FLRig - that the rig switches to USB mode, it obiously must fail.

I would like to propose the following:

- when talking to FLRig, Hamlib should accept whatever modes are reported by FLRig. The main reason is that Hamlib omits a number of transceiver models while FLRig is pretty up to date, even considering many nuances (e.g. QRP Labs transceivers use enhanced Kenwood command set,  but, for instance, QDX differs from QMX, and FLRig handles them both with astonishing accuracy).
- in general, mode setting (at least), if performed through Hamlib, should be absolutely transparent between the ham software and FLrig because in the current state the Hamlib creates phantom errors that would otherwise not occur.

mdblack98:
[...]

PR #1958, opened 2025-12-07
QRPLabs QMX: Set max baud rate to 230400.

tmiw:
I've gotten reports from other FreeDV users that the default baud rate for the QMX (256000) doesn't work. In my experience with my own QMX, I had to force 115200 in order for FreeDV to properly use it for CAT control as the OS (both macOS and Linux) don't like 256000 and throw an error. This PR updates the rig caps for the QMX to ensure that the maximum baud rate supported is 115200.


dforsi:
I agree that the value 256000 is invalid, however for native USB connections it shouldn't matter (it matters for converters to TTL or RS-232C).
Does it work if you use 230400 or other value greater than 115200 in the list below taken from `man cfgetispeed`?
```
       cfsetospeed() sets the output baud rate stored in the termios structure pointed
       to by termios_p to speed, which must be one of these constants:

              B0
              B50
              B75
              B110
              B134
              B150
              B200
              B300
              B600
              B1200
              B1800
              B2400
              B4800
              B9600
              B19200
              B38400
              B57600
              B115200
              B230400
              B460800
              B500000
              B576000
              B921600
              B1000000
              B1152000
              B1500000
              B2000000

       These constants are additionally supported on the SPARC architecture:

              B76800
              B153600
              B307200
              B614400

       These constants are additionally supported on non-SPARC architectures:

              B2500000
              B3000000
              B3500000
              B4000000

       Due to differences between architectures, portable applications should
       check if a particular Bnnn constant is defined prior to using it.
```

N0NB:
The serial value in question is a result of issue #1653 and commit 33adef9.

N0NB:
While this PR can be accepted as is, please test with `230400` and if that is successful, that should probably be the new default value.

tmiw:
> While this PR can be accepted as is, please test with `230400` and if that is successful, that should probably be the new default value.

Can confirm that 230400 works. 👍 I'll also update FreeDV to show the larger baud rates (we cap out at 115200 right now).

N0NB:
Thanks, Mooneer.

Issue #1954, opened 2025-11-25
Who will work with me to finish the ICOM IC-F8101 support?


guitarpicva:
**Describe the bug**
All six digital modes are un-implemented.  These would simplify life for users because (like the IC-7610 and others) they encapsulate configurable mode parameters owned by the user at the radio.

Additionally, but seems globally, there seems to be no "Data PTT" function available.  Most radios have this now and it seems sensible to make this a global option, which is defaulted to MIC PTT if there is no such thing for the radio.

**To Reproduce**
M USBD3 3000 is an error condition.  Same for LSBD1, USBD1, LSBD2, USBD2, LSBD3, USBD3

**Expected behavior**
Changes to the named mode.
"Data PTT" function triggers the "data" port's special PTT if exists, else mic PTT.

**Additional context**
The full CI-V documentation for the IC-F8101 is in hand, and I have years of experience using the radio and it's CI-V (including the oddities).  Happy to pitch in, but I don't have time to learn the entire Hamlib ecosystem.

I have gone through the icf8101.c and related source and tried to add in the appropriate modes, but there are LOTS of places in the code to put this information and it seems that hamlib/rigs.h does not have these ICOM style digital modes implemented.  That's with current git download.

GeoBaltz:
I'm not an Icom expert, or even an Icom user, but maybe I can give a couple of pointers.  Unfortunately, the Hamlib Icom expert passed away last spring, so you may be the reigning pro.

I'd start with adding RIG_MODE_USBD1|RIG_MODE_USBD2|RIG_MODE_USBD3 et al to the #define of ICF8101_MODES. Then in icf8101_set_mode() add cases for RIG_MODE_USBD1, etc, setting modbuf[1] to the appropriate value.  When that compiles, I'd give your `M USBD1 3000` command a try with -vvvvv on rigctl. If it makes it through to the rig then we can look at the width.

73 & GL
n3gb

guitarpicva:
Many thanks! I will give that a go. The pointers to all of the nooks and crannies for modes is a big leg up.

This radio is not hard to deal with, but when Mike Black and I worked on it a few Decembers ago, we just got things basically functional so it could be used with FLDIGI for a special project. Just seems a sad waste not to get it fully covered, so I will dig in a bit this month.

--
Mitch, AB4MW

Sent with [Proton Mail](https://proton.me/mail/home) secure email.

guitarpicva:
[...]

PR #1959, opened 2025-12-08
First round of IC-F8101 updates

GeoBaltz:
Thanks to AB4MW for driving this
See Issue #1954

Nate - you can merge this or let AB4MW resend or I can send you the patch(diff) that I posted

I won't be available Monday and maybe part of Tuesday (guests arriving)


N0NB:
Thanks, George.

PR #1965, opened 2025-12-11
Plug memory leaks

GeoBaltz:
rigctld and rigctltcp are leaking memory every time the select() timer expired, i.e. every 5 seconds.

Move the code to only allocate storage when needed, and when it is sure to be free()ed.

### DIFF
--- a/tests/rigctltcp.c
+++ b/tests/rigctltcp.c
@@ -950,6 +950,7 @@ int main(int argc, char *argv[])
             if (arg->sock < 0)
             {
                 handle_error(RIG_DEBUG_ERR, "accept");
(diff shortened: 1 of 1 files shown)


N0NB:
Thanks George.

Playing with `valgrind` was something that crossed my mind from time to time, but I never got around to it.  Glad you did so.

GeoBaltz:
The valgrind memcheck only showed those errors in half an hour running(slowly!) rigctld.  The thread checker (helgrind) has a lot more to say - that's going to be a real job.

N0NB:
Thanks for your efforts.

PR #1970, opened 2025-12-21
Handle the echo of commands from Skywatcher

### LINKED ISSUE #1968: SkyWatcher AZ-GTi (MC029) does not echo commands – SkyWatcher backend blocks waiting for echo
### Description

When using Hamlib with SkyWatcher AZ-GTi mounts equipped with the MC029 motor controller, communication fails due to an incorrect assumption in the SkyWatcher rotator backend that the mount echoes commands.
This issue prevents rotctl / rotctld from working correctly with newer SkyWatcher Alt-Az mounts.

### Environment

-  Mount: SkyWatcher AZ-GTi
-  Motor Controller: MC029
-  Firmware: v3.54
-  OS: Windows 11
-  Hamlib version: 5.0.0~git (master)
-  Connection: USB (EQDir cable)
-  Baud rate: 9600
-  Flow control: None

### Problem Summary

The current SkyWatcher backend (rotators/skywatcher/skywatcher.c) assumes that all SkyWatcher mounts echo commands, and therefore performs two consecutive reads for every command:

-   Read echoed command
-   Read actual response

However, the MC029 motor controller does NOT echo commands.
As a result, the first read blocks until timeout, causing all commands to fail.

### Evidence

Using serial monitoring tools (AccessPort) and with direct confirmation from SkyWatcher SynScan App Team, it was verified that:

-   MC029 responds with only one response, starting with "=" or "!"
-   No command echo is sent
-   SynScan Pro works correctly because it does not expect an echo

### Temporary Workaround (Confirmed Working)

As a test, I modified skywatcher_cmd() to skip the echo read and read only the actual response.
Specifically, this block was removed/commented out:

>    // echo from the device
>    read_string(port, response, response_len, "\r", 1, 0, 1);

After this change:

-   rotctl p returns correct azimuth/elevation
-   rotctl P <az> <el> moves the mount correctly
-   No timeout, retry, or delay tuning is required

This confirms that the echo assumption is the root cause.

### Recommended Proper Fix (from SkyWatcher)

SkyWatcher confirmed that both echoing and non-echoing mounts exist, and provided two recommended approaches for compatibility.

**Option A (Preferred, robust parsing)**
This is the method currently used by SynScan Pro.

1. Send the command
2. Read incoming data byte-by-byte
3. Ignore all characters until "=" or "!" is received
4. Continue reading until "\r" (0x0D) is received
5. Characters between "=" (or "!") and "\r" form the valid response

This method:

- Works with legacy mounts that echo commands
- Works with MC029 and newer controllers that do not echo
- Does not require detecting echo behavior explicitly

**Option B (Echo-detection at initialization)**
 SkyWatcher also suggested an alternative approach:

1. During initialization, send a single test character (e.g. ":")
2. Check whether the mount echoes back that character
3. If echoed → enable echo-reading logic
4. If not echoed → skip echo reads

This approach also supports both legacy and newer mounts, but requires maintaining an additional per-device state flag.

### Why Timeouts / Delays Are Not the Cause

I tested modifications to:

- timeout
- write_delay
- post_write_delay
- retry logic

None resolved the issue unless the echo read was removed.
Therefore, this is a protocol parsing issue, not a timing issue.

### Request

Please update the SkyWatcher rotator backend to remove the unconditional echo read and adopt either:

- Option A: "=" / "!" anchored response parsing (preferred), or
- Option B: Echo-detection during initialization

I am happy to:

- Provide serial logs
- Provide my temporary patch
- Test any proposed fix on real MC029 hardware

Thank you for maintaining Hamlib.

### Additional Notes

SkyWatcher confirmed:

- Older Alt-Az mounts (Merlin, Virtuoso, Dob GOTO, etc.) may echo commands
- Newer controllers (MC029) do not echo commands
- Official SkyWatcher software already supports both cases correctly

dforsi:
Some firmwares of the Skywatcher rotators echo the commands sent by Hamlib and some don't; this PR makes the presence of the echo optional as suggested in #1968 and fixes an unrelated small error.

Closes #1968.

### DIFF
--- a/rotators/skywatcher/skywatcher.c
+++ b/rotators/skywatcher/skywatcher.c
@@ -66,7 +66,7 @@ static int skywatcher_cmd(ROT *rot, const char *cmd, char *response,
 
     if (code < 0)
     {
-        return -code;
+        return code;
     }
 
     // the actual response
@@ -75,7 +75,7 @@ static int skywatcher_cmd(ROT *rot, const char *cmd, char *response,
 
     if (code < 0)
     {
-        return -code;
+        return code;
     }
 
     // nullify last \r
--- a/rotators/skywatcher/skywatcher.c
+++ b/rotators/skywatcher/skywatcher.c
@@ -60,7 +60,8 @@ static int skywatcher_cmd(ROT *rot, const char *cmd, char *response,
     rig_flush(port);
     size_t cmd_len = strlen(cmd);
     ERROR_CHECK(write_block(port, (unsigned char *) cmd, cmd_len));
-    // echo from the device
+    // Some firmwares echo the commands and some don't, so this may be the echo
+    // of the command sent above or the response
     int code = read_string(port, (unsigned char *) response, response_len, "\r", 1,
                            0, 1);
 
@@ -69,13 +70,15 @@ static int skywatcher_cmd(ROT *rot, const char *cmd, char *response,
         return code;
     }
 
-    // the actual response
-    code = read_string(port, (unsigned char *) response, response_len, "\r", 1, 0,
+    // If we got echo from the device, then get the actual response
+    if (response[0] == ':') {
+        code = read_string(port, (unsigned char *) response, response_len, "\r", 1, 0,
                        1);
 
-    if (code < 0)
-    {
-        return code;
+        if (code < 0)
+        {
+            return code;
+        }
     }
 
     // nullify last \r


dforsi:
@JF9SOM can you test if the changes in this pull request work for you?

JF9SOM:
Hi,

I tested this PR (#1970) on my AZ-GTi mount with MC029 motor controller
(Firmware v3.54) on Windows (MSYS2 / MinGW64).

- rotctl p works correctly
- rotctl P az el works correctly
- No timeouts observed
- Echo-less controller is handled properly

This patch also matches the protocol behavior confirmed by SkyWatcher
(SynScan App Team).

Tested-by: JF9SOM



------ 元のメッセージ ------
差出人 "dforsi" ***@***.***>
宛先 "Hamlib/Hamlib" ***@***.***>
Cc "JF9SOM" ***@***.***>; "Mention" ***@***.***>
日付 2025/12/21 22:26:12
件名 Re: [Hamlib/Hamlib] Handle the echo of commands from Skywatcher (PR 
#1970)

N0NB:
Thanks, Daniele.

PR #1976, opened 2025-12-30
More cppcheck mitigations and code cleanups

GeoBaltz:
Clean up some gripes from cppcheck
Add some Elecraft rigs to Kenwood voice memory handler
Redo parts of neverused.c - needs more relics removed.

For master branch only


N0NB:
Thanks, George.

PR #1980, opened 2026-01-03
Update versions of drivers

dforsi:
I should have updated the version numbers in the original pull requests because these drivers where improved/fixed.


N0NB:
Thanks, Daniele.

PR #1981, opened 2026-01-06
Fix manpages

dforsi:
This PR adds missing commands, fixes some errors and makes `man rigctl` and `man rigctld` more similar to each other.

Still the man pages look confused and confusing to me.

### DIFF
--- a/doc/man1/rigctl.1
+++ b/doc/man1/rigctl.1
@@ -12,8 +12,8 @@
 .
 rigctl \- control radio transceivers and receivers
 .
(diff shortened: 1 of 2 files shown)


N0NB:
Can these updates apply to the Hamlib-4.7 branch as well?  I don't recall any divergence yet, but I could be mistaken.

dforsi:
Yes they can be applied also to the Hamlib-4.7 branch.

N0NB:
Thanks, Daniele.

PR #1984, opened 2026-01-28
Fix buffer overruns in tests/???ctl_parse.c

GeoBaltz:
As per a message on hamlib-developer, there are some stack buffer overruns in the network-enabled control daemons.  Thanks to Vlatko Kosturjak with Marlink Cyber for reporting them.

This fixes the holes I could find in a day's searching.

Should fit into both master and 4.7


N0NB:
Thanks much, George.

Issue #1985, opened 2026-01-30
USB_AF command has changed from IC-7300 to IC-7300Mk2


chuckc192000:
**Describe the bug**
The CI-V command to get or set the USB_AF level has changed from 0x60 to 0x70 in the new IC-7300Mk2.  This has not been updated in HamLib.

**To Reproduce**
Steps to reproduce the behavior (please use `rigctl` with the `-vvvvv` option if at all possible and post the output under number 4 below):
1. `rigctl` model: 3094 (IC-7300Mk2)
2. The function that fails, i.e. `rig_open`, `rig_set_freq`, `rig_get_freq`, etc.: I think it's rig_set_level and rig_get_level
3. Hamlib version: 5.0.0~git
4. Post `rigctl` output: N/A

**Expected behavior**
Should be able to change and view the USB_AF level but the command fails silently due to the wrong CI-V command.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Linux (Fedora 43)
 - Client application: Custom written application (RadioControl) 
 - Client application version: N/A

**Additional context**

N0NB:
Closed by PR #1986.

PR #1986, opened 2026-01-30
IC-7300MK2 update

GeoBaltz:
Fix CI-V commands for the extcmds of IC-7300MK2(and the IC-7300), based on the CI-V manual.

I don't have any way to test these, but they compile and match the table in the manual.

Started by issue #1985, which they should fix.


N0NB:
Thanks George.

PR #1992, opened 2026-02-14
N0 nb multicast

N0NB:
Provide manual pages for installed binaries.

Don't build `rigfreqwalk` as it's more or less broken, IMO.

Will be applied to 4.7.0.

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -8,7 +8,7 @@ EXTRA_DIST = \
 dist_man_MANS = man1/ampctl.1 man1/ampctld.1 \
 	man1/rigctl.1 man1/rigctld.1 man1/rigmem.1 man1/rigsmtr.1 \
 	man1/rigswr.1 man1/rotctl.1 man1/rotctld.1 man1/rigctlcom.1 man1/rigctlsync.1 \
-	man1/rigctltcp.1 man1/rigtestlibusb.1 \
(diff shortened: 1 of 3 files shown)


doc/man1/rigtestlibusb.1
```diff
@@ -0,0 +1,108 @@
...
+Print information on a specific USB device.
+.
+.IP
+.BR Note :
+On a current Debian 12 system this fails with:
+.
+.IP
+.EX
+$ rigtestlibusb -v -d /dev/bus/usb/002/009 
+Testing wrapped devices is not supported on your platform
```
dforsi: WRT this failure, it fails because the original `testlibusb` has `#ifdef __linux__` at https://github.com/libusb/libusb/blob/a61afe5f/examples/testlibusb.c#L237 while `rigtestlibusb` has `#if defined(ANDROID) && defined(LIBUSB_API_VERSION) && (LIBUSB_API_VERSION < 0x01000107)`

N0NB:
Okay, can you work up a quick patch?  I likely won't tag 4.7.0 until after 1900z today.

dforsi:
I'm too late...

This `rigtestlibusb.c` looks like a copy of an older `testlibusb.c` with added compatibility with FreeBSD, the message for when libusb wasn't available at build time and some differences of formatting. The current version of `testlibusb.c` decodes and additional super speed type.

As you wrote, it may be useful when lsusb is missing, but since there isn't anything specific for radio, I'm not sure if Hamlib should provide generic tools.

PR #2001, opened 2026-02-22
Initialize rp->timeout in rig_init

dj5qv:
Timeout setting, e.g. provided by rig_set_conf(), is not effective.
It gets set in rig_open(), overwriting any value set from config, as well as the caps->timeout/2 logic in rig_init().

This change may cause issues for users, that have unreasonable timeouts configured, which did not take effect, but would now.


N0NB:
Thanks, Matthias.

PR #2002, opened 2026-02-22
Update ampctl/ampctld man page documentation with the new commands recently introduced

mikaelnousiainen:
Document changes done in #1996


N0NB:
Thanks, Mikael.

PR #2005, opened 2026-02-28
Update 4.7 rel notes

GeoBaltz:
Nate, I copy'n'pasted your support text from the announcement into ReleaseNotes_4.7.md, hope that's OK.

Then while looking around I realized I had never added it to the tar manifest - oops. Not too useful if Release Notes aren't in the release.


N0NB:
Thanks, George.

PR #2008, opened 2026-03-05
Potential fix for incorrect return-value check for a 'scanf'-like function

mikekppp:
**CodeQL in my branch detected a potential issue with the sscanf starting on line 386 in rigs/codan/codan.c,  Here is the analysis**:

In general, code using scanf-like functions should compare the return value to the exact number of expected successful conversions (here, 1) and treat any other value (including 0 and EOF) as an error. Chaining multiple parse attempts should be done by explicitly checking for success (retval == 1) and only then using the parsed values.

For this specific code in rigs/codan/codan.c, we should change the logic so that:

    We first try to parse "FREQ: %lg".
    If that fails (i.e., retval != 1), we then try "CHAN: %lg".
    Only if one of those calls returns 1 do we scale *freq by 1000 and return RIG_OK.
    If both fail (including EOF), we log an error and return -RIG_EPROTO.

This can be implemented by replacing the current pattern:

retval = sscanf(response, "FREQ: %lg", freq);

if (retval == 0)
    retval = sscanf(response, "CHAN: %lg", freq);

*freq *= 1000; // returned freq is in kHz

if (retval != 1)
{
    ...
    return -RIG_EPROTO;
}

with logic that:

    Only tries the "CHAN: %lg" parse if the first did not succeed (retval != 1).
    Only multiplies *freq after confirming retval == 1.

No new functions or headers are required; we only adjust the control flow in codan_get_freq around the sscanf calls.

### DIFF
--- a/rigs/codan/codan.c
+++ b/rigs/codan/codan.c
@@ -384,17 +384,18 @@ int codan_get_freq(RIG *rig, vfo_t vfo, freq_t *freq)
     }
 
     retval = sscanf(response, "FREQ: %lg", freq);
-
-    if (retval == 0)
-    retval = sscanf(response, "CHAN: %lg", freq);
-
-    *freq *= 1000; // returned freq is in kHz
+    if (retval != 1)
+    {
+        retval = sscanf(response, "CHAN: %lg", freq);
+    }
 
     if (retval != 1)
     {
         rig_debug(RIG_DEBUG_ERR, "%s: Unable to parse response\n", __func__);
         return -RIG_EPROTO;
     }
+
+    *freq *= 1000; // returned freq is in kHz
 
     return RIG_OK;
 }


N0NB:
Thanks for being clear in the use of Copilot in your commit message.

mpapple-swift:
You are welcome.  I would add:
1. Copilot explained the rationale for noting an issue with the existing code and how the proposed changes addressed said issue
2. Copilot's response stated these various points far more succinctly than I would normally write this up
3. It would appear we are on the same page in that the use of AI to generate code and/or create PRs should be noted explicitly.  I would note that I reviewed the proposed change manually and find that it addresses the issue both logically and with code that aligns with the general coding style of Hamlib.

KJ5HST:
> 3\. It would appear we are on the same page in that the use of AI to generate code and/or create PRs should be noted explicitly.

AI is just a logical progression from macros and code completion in IDEs. When has anyone ever pointed out what IDE or punch card machine they used unless it was to promote that product? This argument is the same as what the music industry is arguing. I don't care if it's AI music. If I like it, I'll listen to it. I also don't care how the code was written. I just care that it provides a reliable solution, and AI is much better and faster at it than humans.

gdt:
> AI is just a logical progression from macros and code completion in IDEs. When has anyone ever pointed out what IDE or punch card machine they used unless it was to promote that product?

That's not a fair argument, because it neglects the two core points: copyright infringement  of training data, and wrong answers.

gdt:
I have a few concerns about this PR, but I think they are fixable.  For background, there has recently been extensive discussion on the list.

First, the actual changes are very small: changing a few if conditions, and moving a line.  So it's below the threshold of copyrightability and the legal issues surrounding stolen training data are moot.

I find the PR text extremely wordy.   The key point is pedantically one should check for 1, raher than not 0, and presumably -1 could be returned, and that chaining should be careful.   That's fine, but it reads like a scanf lecture, which I find to be very AI.

The second is that the commit message is bizarre and off point.   "Potential fix"?  Does the human believe it is a fix, or does the human not?  If not, why is it being submitted?  If so, why is it labeled "potential".   Then "code scanning alert no. 1:".  That's just noise, some labeling output from a tool that is totally not relevant.    And yes, incorrect return-value check for scanf.

What should have happened instead is that the human should have seen the alert, understood the issue, made the fix (or taken it), and then written a proper commit message.    Something like "Fix return value checking of scanf when reading freq"  with a body that indicated that checking for 0 is wrong because it could also return -1, chaining, and that freq should only be used if it was read successfully, etc.   The minimal explanation for someone competent to read/write this kind of code.  As always, the commit message should describe the change, and any insert comments should describe the code -- but in this case the new code appears straightforwardly correct.

In my view this is an example of an LLM contribution that has quality problems and takes up extra maintainer time.  All the text in a PR should be de-AI'd before submission.  (It's fine to credit codeql, and proper to disclose LLM use.  I'm not objecting to it being mentioned that codeql helped/found/etc.   Just the LLM verbose/off-point text that came along with it, seemingly not re-edited.)

As it stands, I wouldn't merge this, were I the maintainer.

GeoBaltz:
I wouldn't merge it either, but for a different reason - the code as it stands is correct, with no missing retval checks. It is the "potential issue" warning that is bogus.

The check for `retval == 0` is checking for a specific condition; that the fixed text match before the first conversion fails.  If that is the case, try a different text match, and **reset the retval**.  Let any other retval fall into the final check for `retval != 1`.  Try all the possible values and see if there is any way an unchecked error condition can get through.

I wonder if the code scan would be different if the multiply were moved to right before the `return RIG_OK` even though it's a NOP in the error case.

KJ5HST:
> > AI is just a logical progression from macros and code completion in IDEs. When has anyone ever pointed out what IDE or punch card machine they used unless it was to promote that product?
> 
> That's not a fair argument, because it neglects the two core points: copyright infringement of training data, and wrong answers.

I was referring specifically to just one of your statements, and for that statement, I do believe it is a fair argument.

Your comment after this one is something I am in more agreement with you on, and I've stated it in other discussions regarding AI tools as well: AI requires checks and balances just like hiring a subcontractor or leveraging off-shore teams. Whoever is using AI to contribute, must also be competent and responsible enough to know when AI has failed. If they don't understand the code that was written, then they have no business submitting the code. Same with commit comments, training data and wrong answers. If they don't know these boundaries, their contributions are not valuable.

One small caveat to commit comments and other communication related things: You still have to be careful because they may be good coders, and use crumby AI for crumby translations. Either way, it still requires a competent human with experience to build complex systems -- at least for now.

N0NB:
* On 2026 09 Mar 21:18 -0500, GeoBaltz wrote:
> GeoBaltz left a comment (Hamlib/Hamlib#2008)
> 
> I wouldn't merge it either, but for a different reason - the code as
> it stands is correct, with no missing retval checks. It is the
> "potential issue" warning that is bogus.

The AI stuck "potential" in there as it yet apparently doesn't understand
the "why" of the code.  It's a machine that lack reasoning.

Based on George's comments I won't merge this.

Thanks, Mike.  While it might seem like wasted effort on your part, it
really is not.  It has given me a chance to get greater insight into LLM
submissions.  Please don't hesitate to bring other catches to our
attention for discussion.  Such is valuable.

73, Nate

-- 
Web: https://www.n0nb.us

gdt:
> I wouldn't merge it either, but for a different reason - the code as it stands is correct, with no missing retval checks. It is the "potential issue" warning that is bogus.

I accept your argument that the code was not wrong*, but I do think there is room for improvement.  scanf can return EOF in addition to 0 or 1 (for an invocation with one %).    It's not clear what that means for sscanf, and I suspect the standards are not adequately clear.    I lean to "code should be as simple as possible, so that the argument of correctness is not tricky".

* except that I'd say the use of freq when it was not written is wrong, even if harmless.

In the end I think we agree, that anything using an LLM needs the same standards of solid understanding on the part of the submitting human as non-LLM.

PR #2031, opened 2026-04-13
ftx1: wire SH bandwidth command into set_mode/get_mode (#2027)

### LINKED ISSUE #2027: FTX-1: get_mode returns 0 passband, set_mode ignores width — SH not wired into mode path
## Problem

The FTX-1 driver delegates mode handling to `newcat_set_mode` / `newcat_get_mode`, which do not use the `SH` (bandwidth) CAT command. This means:

1. **`m` (get_mode)** always returns 0 for passband width, even though the radio has a specific filter width set via `SH`
2. **`M MODE WIDTH`** sets the mode via `MD` but silently ignores the width parameter — the radio's filter bandwidth is not changed

The FTX-1 firmware treats `MD` (mode) and `SH` (bandwidth) as completely independent commands. `MD` never touches bandwidth, and querying `MD` never returns it. The `SH` command works correctly for both set and read — it just isn't called.

## Workaround

Users can set bandwidth by sending `SH` directly via raw CAT (`w SH0019;` for code 19 = 3200 Hz in SSB mode), and read it back with `w SH0;`. But this bypasses the standard Hamlib mode/passband interface.

## SH Lookup Table (from CAT reference)

SH uses mode-dependent codes:

- **SSB/DATA**: codes 0–21 (200 Hz to 4000 Hz)
- **CW**: codes 0–17 (50 Hz to 3000 Hz)
- **RTTY**: codes 0–12 (50 Hz to 2400 Hz)
- **AM**: codes 0–3 (3000 Hz to 9000 Hz)
- **FM**: code 0 only (no selectable bandwidth)

## Expected Fix

- `newcat_get_mode` (or FTX-1 override): query `SH` alongside `MD`, map the SH code to Hz for the passband return
- `newcat_set_mode` (or FTX-1 override): when width != 0, find the nearest valid SH code for the target mode and send `SH` after `MD`

## Environment

- Radio: Yaesu FTX-1, firmware v1.08+
- Hamlib driver: `rigs/yaesu/ftx1/ftx1.c` (delegates to newcat framework)

KJ5HST:
## Summary

Fixes #2027 — FTX-1 `get_mode` always returned `passband=0` and `set_mode` silently dropped the width argument, because the driver delegates to `newcat_set_mode` / `newcat_get_mode` and those helpers had no FTX-1 branch.

This adds dedicated `is_ftx1` branches to `newcat_set_rx_bandwidth` and `newcat_get_rx_bandwidth`, and extends the SH command-string builder at line 9671 to emit the 7-char `SH00NN;` form for FTX-1. Follows the reviewer guidance on the issue (N0NB: "test for `RIG_MODEL_FTX1` and do what is needed in `newcat.c` to support SH") and structures the gating so a future Yaesu model sharing this MD/SH pattern is a one-line extension.

## Why a dedicated branch instead of piggybacking on FT-710

The FTX-1 bandwidth table is per the FTX-1 CAT Operation Reference Manual Table 5 (doc 2508-C, firmware v1.08+). CW/DATA/RTTY/PSK codes 01–21 are identical to the FT-710 table, but the **SSB column diverges at codes 12/13/14**:

| Code | FT-710 (existing) | FTX-1 (spec) |
|------|-------------------|--------------|
| 12   | 2200 Hz           | **2250 Hz**  |
| 13   | 2300 Hz           | **2400 Hz**  |
| 14   | 2400 Hz           | **2450 Hz**  |

Sharing the branch would silently misreport 2400 Hz on the FTX-1 — exactly the canonical SSB filter width — because the radio would return `w=13` (= 2400 Hz on FTX-1) but the FT-710 decoder would translate that to 2300 Hz. PSK is also included in the CW/DATA/RTTY case group because Table 5 groups it there, and live probing confirmed PSK mode defaults to SH code 19 (the DATA default).

## Verification

Built clean against current master on macOS (Apple Silicon). Verified end-to-end against a live **FTX-1 field, firmware v1.12**, at 115200 bps through rigctl:

```
M USB 2400  → m → USB 2400   (was broken: returned 0)
M USB 2700  → m → USB 2700   (was broken: returned 0)
M USB 2250  → m → USB 2250   (corrected code 12)
M USB 2450  → m → USB 2450   (corrected code 14)
M USB 3000  → m → USB 3000
M CW 500    → m → CW 500     (was broken: returned 0)
M CW 2400   → m → CW 2400
M PKTUSB 3000 → m → PKTUSB 3000
M AM 9000   → m → AM 9000
M FM 16000  → m → FM 16000
```

Cross-checked raw-CAT afterwards to confirm the radio received the *corrected* SH codes on the wire (not the stale FT-710 codes):

```
USB 2250 Hz → SH0012;
USB 2400 Hz → SH0013;   ← proof the corrected branch is in effect
USB 2450 Hz → SH0014;
```

## Firmware note

The existing `newcat_get_rx_bandwidth` 7-byte response parser at line 9922 (`SH%*1d%1d%3d`) already handles the FTX-1 `SH00NN;` response shape unchanged. The read-command builder at line 9908 (`SH%c;`) also already produces the correct query form. Only the decode tables and SET branch needed new code.

SH set was empirically confirmed to work and persist in-session on firmware v1.12 across 16 sweep writes in LSB and CW-U, plus the full rigctl test above.

## Test plan

- [x] Build clean on macOS ARM
- [x] `rigctl -m 1051` set/get round-trip for SSB, CW, DATA, AM, FM
- [x] Raw-CAT confirmation of corrected SH code mapping
- [x] Out-of-range behavior: firmware silently clamps (CW code 22 → 21), hamlib pre-clamps per mode
- [ ] Regression on FT-710 / FTDX10 (no hardware available; existing branch is untouched, `is_ftx1` false path)

### DIFF
--- a/rigs/yaesu/newcat.c
+++ b/rigs/yaesu/newcat.c
@@ -9413,6 +9413,107 @@ int newcat_set_rx_bandwidth(RIG *rig, vfo_t vfo, rmode_t mode, pbwidth_t width)
             RETURNFUNC(RIG_OK);
         }
     } // end is_ftdx101d || is_ftdx101mp || is_ftdx10 || is_ft710
+    else if (is_ftx1)
+    {
+        // Bandwidth table per FTX-1 CAT Operation Reference Manual Table 5
+        // (doc 2508-C). Differs from the FT-710/FTDX10 table at SSB codes
+        // 12/13/14 (2250/2400/2450 Hz vs 2200/2300/2400), so do not merge.
+        switch (mode)
+        {
+        case RIG_MODE_PKTUSB:
+        case RIG_MODE_PKTLSB:
+        case RIG_MODE_RTTY:
+        case RIG_MODE_RTTYR:
+        case RIG_MODE_CW:
+        case RIG_MODE_CWR:
+        case RIG_MODE_PSK:
+            if (width == RIG_PASSBAND_NORMAL) { w = 0; }
+            else if (width <= 50) { w = 1; }
+            else if (width <= 100) { w = 2; }
+            else if (width <= 150) { w = 3; }
+            else if (width <= 200) { w = 4; }
+            else if (width <= 250) { w = 5; }
+            else if (width <= 300) { w = 6; }
+            else if (width <= 350) { w = 7; }
+            else if (width <= 400) { w = 8; }
+            else if (width <= 450) { w = 9; }
+            else if (width <= 500) { w = 10; }
+            else if (width <= 600) { w = 11; }
+            else if (width <= 800) { w = 12; }
+            else if (width <= 1200) { w = 13; }
+            else if (width <= 1400) { w = 14; }
+            else if (width <= 1700) { w = 15; }
+            else if (width <= 2000) { w = 16; }
+            else if (width <= 2400) { w = 17; }
+            else if (width <= 3000) { w = 18; }
+            else if (width <= 3200) { w = 19; }
+            else if (width <= 3500) { w = 20; }
+            else { w = 21; } // 4000 Hz
+
+            break;
+
+        case RIG_MODE_LSB:
+        case RIG_MODE_USB:
+            if (width == RIG_PASSBAND_NORMAL) { w = 0; }
+            else if (width <= 300) {  w = 1; }
+            else if (width <= 400) {  w = 2; }
+            else if (width <= 600) {  w = 3; }
+            else if (width <= 850) {  w = 4; }
+            else if (width <= 1100) {  w = 5; }
+            else if (width <= 1200) {  w = 6; }
+            else if (width <= 1500) {  w = 7; }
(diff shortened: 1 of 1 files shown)


N0NB:
@GeoBaltz LGTM, what do you think?

GeoBaltz:
Looks OK, and for both branches.  Definitely correct about the dedicated branch.

During the time I spent on the FT-710 issues, I found that all of the if()s, cases and branches in newcat.c need a lot of TLC.  I've been working on a set of changes to move a lot of the code into tables off of priv_caps (+ other random cleanups) that I'll have for 5.0, probably with the infrastructure ready in a week or so.  Moving the data might take a while.

N0NB:
Thanks for your review, George.

I'll get this into both branches.

N0NB:
Thanks, Terrell.

PR #2035, opened 2026-04-15
ftx1: rewrite set_channel using VFO-B + BM workaround (#2034)

### LINKED ISSUE #2034: FTX-1: ftx1_set_channel uses MW, which firmware rejects on all channels
## Problem

`rig_set_channel` on an FTX-1 always fails because the driver's `ftx1_set_channel` implementation builds and sends the `MW` (Memory Write) CAT command, which the FTX-1 firmware rejects universally with `?;` — on every channel, in every mode, regardless of parameters. The channel never gets written.

## Empirical evidence

Tested directly against a live FTX-1 (field head + Optima/SPA-1, firmware MAIN v1.12) at 115200 bps. Sent the exact MW command string the driver would build:

```
tx:  MW00005014250000+0000002120000;
rx:  ?;
```

MR readback of the target channel afterward confirms nothing was written:

```
tx:  MR00005;
rx:  ?;
```

This behavior is reproducible across every channel number and every mode I tested. `MW` appears to be completely non-functional on current FTX-1 firmware.

## Current broken code

`rigs/yaesu/ftx1/ftx1_mem.c`

`ftx1_set_channel()` at line 147 builds a 30-char MW command with freq, clarifier, mode, tone, and shift parameters, then calls `newcat_set_cmd()` at line 246. The firmware rejects the command and the function returns `-RIG_EPROTO`. No channel state is modified.

Related: the `DC` (DCS code) command is similarly firmware-broken — all variants return `?;` — and must be replaced by `CN` with `P2=1`. The FTX-1 driver already handles DCS via `CN` correctly in `ftx1_ctcss.c`, so the memory-write rewrite will follow the same approach.

## Proposed fix — VFO-B + BM workaround

The FTX-1 supports memory write via an indirect sequence that empirically works:

1. `MC<ch>;` — point the "last-selected channel" cursor at the target slot. Ignore `?;` if the slot is currently unprogrammed; the cursor still updates.
2. `VM000;` — force VFO mode (MC may have entered memory mode).
3. Configure **VFO-B** to the target state:
   - `FB%09.0f;` for frequency
   - `MD1<mode_char>;` for mode
   - `CT1<mode>;` + `CN10<idx>;` for CTCSS (FM channels only)
   - `CT13;` + `CN11<code>;` for DCS
   - `OS1<shift>;` for repeater shift
4. `BM;` — copy VFO-B state to the last-selected memory channel.

VFO-B is used rather than VFO-A so the operator's primary RX VFO is untouched. A production version should snapshot VFO-B at entry and restore it on exit (including error paths).

## Validation of the workaround

Verified end-to-end on the same live radio:

```
# Starting state: channel 1 = 7.000.000 LSB (existing memory)
tx:  MC000001;                  rx:  ''        (select ch 1, enter memory mode)
tx:  VM000;                     rx:  ''        (back to VFO mode)
tx:  FB007050000;               rx:  ''        (VFO-B = 7.050 MHz)
tx:  MD11;                      rx:  ''        (VFO-B = LSB)
tx:  BM;                        rx:  ''        (commit)

# Verify
tx:  MC000001;                  rx:  ''
tx:  FA;                        rx:  FA007050000;   ← channel 1 now 7.050
tx:  MD0;                       rx:  MD01;          ← mode preserved

# Restore channel 1 to 7.000 (reverse with the same sequence)
# Final: MR reads 7.000 LSB again, VFO-B restored to original state
```

Channel 1 was successfully rewritten and restored — the workaround is proven on hardware.

## Known scope gap

The existing MW path also handles clarifier/RIT state for the channel. Whether `BM` propagates clarifier state to the memory slot hasn't been empirically verified yet. The planned rewrite handles freq + mode + CTCSS + DCS + shift; clarifier persistence for channels is a potential follow-up if users report it missing.

## Environment

- Radio: Yaesu FTX-1 field head + Optima/SPA-1, firmware MAIN v1.12
- Hamlib: current master (post-merge of #2031)

I have a working rewrite ready and will submit it as a follow-up PR.

KJ5HST:
## Summary

Fixes #2034 — `rig_set_channel` on an FTX-1 always fails because the driver's `ftx1_set_channel` builds and sends the `MW` command, which is firmware-broken on v1.08+ and returns `?;` universally. Replace the MW path with the sequence the firmware actually accepts: `MC` + `VM000` + VFO-B configuration + `BM`.

## The sequence

```
1. MC<ch>;         point the "last-selected channel" cursor at the target
2. VM000;          force VFO mode in case MC entered memory mode
3. FB<freq>;       program VFO-B frequency
   MD1<mode>;      program VFO-B mode
   CT1<n>; CN10<tone>; CN11<dcs>; OS1<shift>;   (FM-family channels only)
4. BM;             copy VFO-B to the last-selected memory channel
```

VFO-B is used rather than VFO-A so the operator's primary RX VFO is not disturbed. VFO-B's frequency and mode are snapshotted at entry and best-effort restored on exit (including the error path), so callers see minimal side effects.

## Why a dedicated branch rather than fixing MW

The `MW` command is rejected by firmware on every FTX-1 channel in every mode — not a driver bug we can patch around. Confirmed against fw v1.12:

```
tx:  MW00005014250000+0000002120000;
rx:  ?;
```

The `BM` commit path is the documented Yaesu alternative and works empirically.

Related firmware bug: `DC` (DCS code) is similarly broken and must be replaced with `CN` where `P2=1`. The new code uses `CN11<code>;` for DCS, matching what `ftx1_ctcss.c` already does for the non-channel CTCSS/DCS path.

## Verification

Built clean against current master on macOS (Apple Silicon). End-to-end tested against a live **FTX-1 field head + Optima/SPA-1, firmware MAIN v1.12**, at 115200 bps:

```
# Starting state: channel 1 = 7.000.000 MHz LSB (existing memory)

$ rigctl -m 1051 -r /dev/cu.usbserial-01AE33F90 -s 115200
Rig command: H 1
Frequency: 7050000
mode (FM,LSB,etc...): LSB
rptr shift (+-0): 0
rit (Hz,0=off): 0
xit (Hz,0=off): 0
ctcss tone freq in tenth of Hz (0=off): 0
ctcss sql freq in tenth of Hz (0=off): 0
Rig command: q

# Confirm channel 1 now holds 7.050 MHz
tx:  MC000001;    rx:  ''
tx:  FA;          rx:  FA007050000;
tx:  MD0;         rx:  MD01;

# Round-trip restore via the same rigctl path back to 7.000 MHz
$ rigctl -m 1051 ... H 1 / 7000000 / LSB / 0 / 0 / 0 / 0 / 0

tx:  MC000001;    rx:  ''
tx:  FA;          rx:  FA007000000;    ← restored
tx:  MD0;         rx:  MD01;

# VFO-A and VFO-B both unchanged from pre-test state
tx:  FA;          rx:  FA144260000;    (was 144.260 FM)
tx:  FB;          rx:  FB028074000;    (was 28.074 USB)
```

The debug trace at `-vvvv` shows the full 7-step CAT sequence (save FB + MD1 → MC → VM000 → FB → MD1 → BM → restore FB + MD1), all returning `RIG_OK`.

## Known scope gap

The previous MW path also programmed clarifier (RIT/XIT) state into the channel. Whether `BM` propagates clarifier state to the stored memory slot has not been empirically verified, and getting it wrong silently (apparently-successful write but missing clarifier) is worse than not programming it at all. This rewrite drops clarifier handling; it's a reasonable follow-up once clarifier-via-BM is confirmed.

Also: the existing `ftx1_get_channel` discards the CTCSS P8 byte (`ftx1_mem.c:380` has `(void)ctcss_mode;`), so reading back a channel after this write won't report tone state yet. That's a separate read-side gap tracked as a follow-up.

## Minor refactor

Exposes `ftx1_freq_to_tone_num()` in `ftx1.h` — previously file-static in `ftx1_ctcss.c`. The new `ftx1_set_channel` calls it to convert CTCSS frequencies (in tenths of Hz) to tone indices for the `CN` command, rather than duplicating the 50-element lookup table.

## Test plan

- [x] Build clean on macOS ARM
- [x] `rigctl H 1` round-trips channel 1 between 7.000 MHz LSB and 7.050 MHz LSB
- [x] VFO-A state preserved across the operation
- [x] VFO-B state preserved across the operation (save/restore works)
- [x] Raw-CAT readback confirms channel 1 was actually written
- [ ] Test with an FM-family channel + CTCSS tone (no FM memory currently programmed on my test radio; left unverified but code path is straightforward)
- [ ] Test with a DCS channel (same as above)
- [ ] Regression on other newcat rigs (no code outside ftx1_mem.c/ftx1_ctcss.c/ftx1.h touched)

### DIFF
--- a/rigs/yaesu/ftx1/ftx1.c
+++ b/rigs/yaesu/ftx1/ftx1.c
@@ -72,6 +72,7 @@ extern int ftx1_set_att_helper(RIG *rig, vfo_t vfo, value_t val);
 extern int ftx1_get_att_helper(RIG *rig, vfo_t vfo, value_t *val);
 
 /* Wrappers from ftx1_func.c for rig caps */
+extern int ftx1_set_mode(RIG *rig, vfo_t vfo, rmode_t mode, pbwidth_t width);
 extern int ftx1_set_ptt_func(RIG *rig, vfo_t vfo, ptt_t ptt);
 extern int ftx1_get_ptt_func(RIG *rig, vfo_t vfo, ptt_t *ptt);
 extern int ftx1_set_powerstat_func(RIG *rig, powerstat_t status);
@@ -863,6 +864,43 @@ static int ftx1_close(RIG *rig)
 }
 
 /*
+ * ftx1_ensure_vfo_mode - exit Memory mode if the driver last saw MC enter it
+ *
+ * FTX-1 firmware rejects FA/OS sets when the Main side is in Memory mode
+ * (VM011) and silently discards CT/CN/MD sets as transient memory-tune
+ * overlays. Callers that set VFO-side state should call this first so the
+ * operation actually lands.
+ *
+ * We track memory-mode entry in priv->ftx1_in_memory_mode (set by
+ * ftx1_set_mem after a successful MC). If the flag is clear, this is a
+ * no-op — we don't spam the rig with VM000 on every VFO-state set.
+ *
+ * Best-effort: on failure the flag is cleared unconditionally so we
+ * don't loop.
+ */
+void ftx1_ensure_vfo_mode(RIG *rig)
+{
+    struct newcat_priv_data *priv;
+
+    if (!rig)
+    {
+        return;
+    }
+
+    priv = STATE(rig)->priv;
+    if (!priv || !priv->ftx1_in_memory_mode)
+    {
+        return;
+    }
+
+    rig_debug(RIG_DEBUG_VERBOSE, "%s: exiting memory mode via VM000;\n",
+              __func__);
+    SNPRINTF(priv->cmd_str, sizeof(priv->cmd_str), "VM000;");
+    (void)newcat_set_cmd(rig);
+    priv->ftx1_in_memory_mode = 0;
+}
+
+/*
  * ftx1_has_spa1 - Check if Optima/SPA-1 amplifier is present
  *
  * Returns 1 if Optima/SPA-1 detected, 0 otherwise.
@@ -1157,7 +1195,7 @@ struct rig_caps ftx1_caps = {
     .rig_close = ftx1_close,  /* FTX-1 specific close - resets detection state */
     .set_freq = ftx1_set_freq,
     .get_freq = ftx1_get_freq,
-    .set_mode = newcat_set_mode,
+    .set_mode = ftx1_set_mode,
     .get_mode = newcat_get_mode,
     .set_vfo = ftx1_set_vfo,
     .get_vfo = ftx1_get_vfo,
--- a/rigs/yaesu/ftx1/ftx1.h
+++ b/rigs/yaesu/ftx1/ftx1.h
@@ -222,6 +222,14 @@ extern int ftx1_has_spa1(RIG *rig);
 extern int ftx1_get_head_type(RIG *rig);
 
 /*
+ * ftx1_ensure_vfo_mode - force the radio out of Memory mode if the driver
+ * knows it entered via a prior set_mem.  Call at the top of any VFO-state
+ * setter that would otherwise fail or behave transiently in Memory mode.
+ * See ftx1.c for full rationale.
+ */
+extern void ftx1_ensure_vfo_mode(RIG *rig);
+
+/*
  * FTX-1 band select functions (defined in ftx1_scan.c)
  */
 extern int ftx1_set_band_select(RIG *rig, vfo_t vfo, int band);
--- a/rigs/yaesu/ftx1/ftx1.h
+++ b/rigs/yaesu/ftx1/ftx1.h
@@ -231,5 +231,6 @@ extern int ftx1_get_band_select(RIG *rig, vfo_t vfo, int *band);
  * FTX-1 CTCSS helpers (defined in ftx1_ctcss.c)
  */
 extern int ftx1_freq_to_tone_num(unsigned int freq);
+extern unsigned int ftx1_tone_num_to_freq(int num);
 
 #endif /* _FTX1_H */
--- a/rigs/yaesu/ftx1/ftx1.h
+++ b/rigs/yaesu/ftx1/ftx1.h
@@ -227,4 +227,9 @@ extern int ftx1_get_head_type(RIG *rig);
 extern int ftx1_set_band_select(RIG *rig, vfo_t vfo, int band);
 extern int ftx1_get_band_select(RIG *rig, vfo_t vfo, int *band);
 
+/*
+ * FTX-1 CTCSS helpers (defined in ftx1_ctcss.c)
+ */
+extern int ftx1_freq_to_tone_num(unsigned int freq);
+
 #endif /* _FTX1_H */
--- a/rigs/yaesu/ftx1/ftx1_ctcss.c
+++ b/rigs/yaesu/ftx1/ftx1_ctcss.c
@@ -62,7 +62,7 @@ int ftx1_freq_to_tone_num(unsigned int freq)
 }
 
 /* Convert tone number (0-based per spec) to frequency (in 0.1 Hz) */
-static unsigned int ftx1_tone_num_to_freq(int num)
+unsigned int ftx1_tone_num_to_freq(int num)
 {
     if (num < FTX1_CTCSS_MIN || num > FTX1_CTCSS_MAX)
     {
--- a/rigs/yaesu/ftx1/ftx1_ctcss.c
+++ b/rigs/yaesu/ftx1/ftx1_ctcss.c
@@ -46,7 +46,7 @@ static const unsigned int ftx1_ctcss_tones[] = {
 };
 
 /* Convert CTCSS frequency (in 0.1 Hz) to tone number (0-based per spec) */
-static int ftx1_freq_to_tone_num(unsigned int freq)
+int ftx1_freq_to_tone_num(unsigned int freq)
 {
     int i;
 
--- a/rigs/yaesu/ftx1/ftx1_ctcss.c
+++ b/rigs/yaesu/ftx1/ftx1_ctcss.c
@@ -101,6 +101,10 @@ int ftx1_set_ctcss_mode(RIG *rig, tone_t mode)
 
     rig_debug(RIG_DEBUG_VERBOSE, "%s: mode=%u\n", __func__, mode);
 
+    /* CT sets are transient overlays in Memory mode — accepted but not
+     * persisted.  Exit memory mode so the change actually sticks. */
+    ftx1_ensure_vfo_mode(rig);
+
     /* CT P1 P2; where P1=VFO (0=Main), P2=mode */
     SNPRINTF(priv->cmd_str, sizeof(priv->cmd_str), "CT0%d;", (int)mode);
     return newcat_set_cmd(rig);
@@ -168,6 +172,9 @@ int ftx1_set_ctcss_tone(RIG *rig, vfo_t vfo, tone_t tone)
     rig_debug(RIG_DEBUG_VERBOSE, "%s: tone=%u tone_num=%d\n", __func__, tone,
               tone_num);
 
+    /* CN on Main is a transient overlay in Memory mode — exit first. */
+    ftx1_ensure_vfo_mode(rig);
+
     /* P1=0 for TX tone, P2P3P4 is 3-digit tone number */
     SNPRINTF(priv->cmd_str, sizeof(priv->cmd_str), "CN00%03d;", tone_num);
     return newcat_set_cmd(rig);
@@ -225,7 +232,9 @@ int ftx1_set_ctcss_sql(RIG *rig, vfo_t vfo, tone_t tone)
     rig_debug(RIG_DEBUG_VERBOSE, "%s: tone=%u tone_num=%d\n", __func__, tone,
               tone_num);
 
-    /* P1=1 for RX tone, P2P3P4 is 3-digit tone number */
+    /* P1=1 for RX tone, P2P3P4 is 3-digit tone number.  CN10 targets
+     * Sub VFO, which is not affected by Main's Memory mode, so no
+     * VM000 injection is needed here. */
     SNPRINTF(priv->cmd_str, sizeof(priv->cmd_str), "CN10%03d;", tone_num);
     return newcat_set_cmd(rig);
 }
@@ -279,6 +288,9 @@ int ftx1_set_dcs_code(RIG *rig, vfo_t vfo, tone_t code)
     (void)vfo;  /* Unused - always Main for now */
 
     rig_debug(RIG_DEBUG_VERBOSE, "%s: code=%u\n", __func__, code);
+
+    /* CN on Main is a transient overlay in Memory mode — exit first. */
+    ftx1_ensure_vfo_mode(rig);
 
     /* CN01XXX: Main VFO (0), DCS (1), code index */
     SNPRINTF(priv->cmd_str, sizeof(priv->cmd_str), "CN01%03u;", code);
--- a/rigs/yaesu/ftx1/ftx1_freq.c
+++ b/rigs/yaesu/ftx1/ftx1_freq.c
@@ -17,6 +17,7 @@
 #include "misc.h"
 #include "yaesu.h"
 #include "newcat.h"
+#include "ftx1.h"
 
 // FTX-1 uses 9-digit Hz format for frequency
 #define FTX1_FREQ_DIGITS 9
@@ -43,6 +44,10 @@ int ftx1_set_freq(RIG *rig, vfo_t vfo, freq_t freq)
     case RIG_VFO_MAIN:
     case RIG_VFO_CURR:
         cmd = 'A';
+        /* FA set is rejected with ?; while the Main side is in Memory
+         * mode. Force an exit before sending the command. FB (sub) is
+         * unaffected and does not need this. */
+        ftx1_ensure_vfo_mode(rig);
         break;
 
     case RIG_VFO_B:
@@ -175,6 +180,9 @@ int ftx1_set_rptr_shift(RIG *rig, vfo_t vfo, rptr_shift_t shift)
 
     rig_debug(RIG_DEBUG_VERBOSE, "%s: vfo=%s shift=%d\n",
               __func__, rig_strvfo(vfo), shift);
+
+    /* OS Main is rejected with ?; in Memory mode. Exit before sending. */
+    ftx1_ensure_vfo_mode(rig);
 
     /* Determine VFO */
     switch (vfo)
--- a/rigs/yaesu/ftx1/ftx1_mem.c
+++ b/rigs/yaesu/ftx1/ftx1_mem.c
@@ -141,21 +141,34 @@ int ftx1_get_mem(RIG *rig, vfo_t vfo, int *ch)
  *                target slot. Ignore '?' if the slot is currently
  *                unprogrammed — the cursor still updates for BM.
  *   2. VM000;    Force VFO mode in case MC entered memory mode.
- *   3. FB / MD1 / CT1 / CN10 / CN11 / OS1   Program VFO-B.
- *                CT/CN/OS only work in FM, so tone/shift are only
- *                emitted for FM-family channel modes.
+ *   3. FB / MD1 / CT1 / CN10 / OS1 / CF1   Program VFO-B.
+ *                CT/CN/OS only work in FM, so tone and shift are only
+ *                emitted for FM-family channel modes. Clarifier (CF)
+ *                is programmed for all modes.
  *   4. BM;       Copy VFO-B to the last-selected memory channel.
  *
  * VFO-B is used rather than VFO-A so the operator's primary RX VFO
  * is not disturbed. VFO-B's frequency and mode are snapshotted at
  * entry and best-effort restored on exit (including error paths).
- * Advanced VFO-B state (CT/CN/OS/CF) may still be clobbered for
- * callers that relied on it.
+ * VFO-B's clarifier and CT state are cleared during cleanup — any
+ * caller that relied on those aspects of VFO-B's state will need to
+ * reprogram them after rig_set_channel.
  *
- * Clarifier/RIT/XIT persistence into the memory slot is not yet
- * supported — BM's handling of clarifier state has not been
- * empirically verified. This is a potential follow-up if users
- * report it missing.
+ * Firmware limitations verified empirically on fw v1.12: the BM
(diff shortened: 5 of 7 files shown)


KJ5HST:
## Verified the three unknowns against a live FTX-1

Ran empirical tests against the same field head + Optima/SPA-1 (fw v1.12). Results:

| Field | BM propagates? | Action |
|-------|---------------|--------|
| freq | ✅ | already in initial commit |
| mode | ✅ | already in initial commit |
| CTCSS (ENC/TSQ + tone) | ✅ | already in initial commit |
| Repeater shift | ✅ | already in initial commit |
| **Clarifier offset** | ✅ | **added** — `CF101<sign><4-digit>;` programs the offset |
| **RIT (RX CLAR flag)** | ✅ | **added** — `CF100<0/1>0000;` toggles RX CLAR |
| **XIT (TX CLAR flag)** | ❌ firmware drops | emit anyway for forward compat, flag in comment |
| **DCS code + mode** | ❌ firmware drops | emit anyway for forward compat, flag in comment |

### Method

For each feature I programmed VFO-B to a known test state, committed with `BM;`, then read the raw 30-char MR response and compared byte positions. DCS and XIT both produced an MR response byte-for-byte identical to the no-tone/no-clarifier baseline, across multiple codes/offsets — conclusive evidence that BM isn't capturing those fields, even though querying VFO-B between the `CT13;`/`CF100.3` set and the `BM;` shows the target state is present.

### Commit

Pushed `c1a2594` on top of the initial `4024cd5`:

- Adds the clarifier `CF` sequence to the rewrite — `rig_set_channel` now persists `chan->rit` correctly
- Documents the two firmware limitations (DCS, XIT) in the function-level comment
- Tightens the cleanup path so VFO-B's clarifier and CT state are explicitly reset on exit, not just FB/MD1

### Verified via rigctl

```
$ rigctl -m 1051 H 1
Frequency: 7050000
mode: LSB
rptr shift: 0
rit: 200
xit: 0
ctcss tone: 0
ctcss sql: 0

# raw MR readback
MR00001007050000+020010110000;
#              ^^^^^^ ^^ — +0200 offset, rx=1, tx=0 ✓
```

Channel 1 round-trips cleanly and both VFOs survive. Full test plan items for freq/mode/clarifier are now checked. DCS and XIT are marked explicitly as firmware limitations rather than untested.

KJ5HST:
I'm going add a few more changes to this PR instead of adding to the merge effort. Looks like the 3/16 firmware update resulted in new behavior, but Yaesu has not updated the CAT documentation since August.

N0NB:
Given that this only affects the memory channels, I'm going to release 4.7.1 today and then this will be included in 4.7.2 when you're satisfied it is ready.

KJ5HST:
## Scope expansion: also fix the get_channel side

Pushed `130d733` as a third commit on this branch — it addresses a closely related bug in `ftx1_get_channel` (same file, same firmware quirks) that showed up while verifying the set path.

### The bug

`ftx1_get_channel` in `ftx1_mem.c` parses the MR response's P8 byte (CTCSS mode) and then immediately discards it with `(void)ctcss_mode;`. As a result, `rig_get_channel` has always returned `chan->ctcss_tone == 0` regardless of what's actually programmed — an FM memory channel with a CTCSS tone reads back as "no tone", silently misleading for callers.

### Why MR alone isn't enough

Byte-level comparison of MR responses across different tone configurations (fw v1.12):

| Scenario | MR byte 23 (P8) |
|----------|-----------------|
| No tone | `0` |
| CTCSS ENC tone 12 | `2` |
| CTCSS TSQ tone 12 | `1` |
| DCS 023 | `0` (firmware doesn't store, see earlier PR comment) |

Only P8 differs — the tone number is nowhere in the MR response. It only lives in `CN`, which is VFO-oriented, not channel-oriented.

### Fix

New static helper `ftx1_read_channel_tone_state()` that briefly selects the memory channel via `MC<ch>;`, queries `CN00;` (CTCSS) or `CN01;` (DCS), runs the result through the existing `ftx1_tone_num_to_freq()` helper, and restores the caller's prior VM/MC state. The intrusion is only triggered when P8 indicates a tone is present — no-tone channels skip the CN query entirely.

Restore logic:
- If the caller was in VFO mode: `VM000;`
- If the caller was in memory mode on some other channel: `MC<original>;`

### Verification

```
# Write + read ENC
$ rigctl -m 1051 H 1 / 146520000 / FM / 0 / 0 / 0 / 1000 / 0
$ rigctl -m 1051 h 1 0
  Freq: 146.5200000 MHz  Mode: FM
  CTCSS: 100.0Hz, CTCSSsql: 0.0Hz, DCS: 0.0, DCSsql: 0.0

# Write + read TSQ
$ rigctl -m 1051 H 1 / 146520000 / FM / 0 / 0 / 0 / 1000 / 1000
$ rigctl -m 1051 h 1 0
  CTCSS: 100.0Hz, CTCSSsql: 100.0Hz, DCS: 0.0, DCSsql: 0.0

# Write + read no-tone baseline
$ rigctl -m 1051 H 1 / 7000000 / LSB / 0 / 0 / 0 / 0 / 0
$ rigctl -m 1051 h 1 0
  CTCSS: 0.0Hz, CTCSSsql: 0.0Hz, DCS: 0.0, DCSsql: 0.0
```

After each test, VFO state and channel 1 MR response are byte-for-byte identical to the pre-test snapshot — the intrusive sequence cleans up correctly.

### DCS readback

The `P8==3` branch is implemented and queries `CN01;` for the DCS code, but I haven't been able to verify it on my hardware. The firmware doesn't appear to persist DCS state into memory channels via the BM write path (see earlier comment), so there's no DCS channel to read back on the test radio. If front-panel-programmed DCS channels do produce `P8=3` in MR, the code will handle them correctly; if not, the branch is dead code that future verification can confirm.

KJ5HST:
## Third scope expansion: auto-exit Memory mode before VFO-state setters

Pushed `370dca2` as a fourth commit on this branch. Same file family, same firmware quirk family — it's the last known memory-mode issue in the FTX-1 driver and ties up a loose end from the set_channel/get_channel work.

### The bug

Commands that target Main-side VFO state behave poorly on fw v1.12 when the radio is in Memory mode (VM011). Empirically verified:

| Command | In Memory mode |
|---------|----------------|
| `FA` (main freq set) | **rejected** with `?;` |
| `OS` (repeater shift) | **rejected** with `?;` |
| `MD` (main mode set) | accepted but **transient** — vanishes on next MC recall |
| `CT` (CTCSS mode main) | transient overlay (same pattern) |
| `CN00/CN01` (main tone/DCS) | transient overlay (same pattern) |
| `FB` / `CN10/CN11` (Sub side) | work fine — Sub is never in Memory mode |

A caller doing `rig_set_mem()` followed by `rig_set_freq()` would silently fail (FA rejected), and `rig_set_mode()` would appear to succeed but the mode change would vanish on the next MC recall. Hamlib had no awareness of the Memory-mode state transition.

### Fix

Add a small memory-mode tracking flag to `struct newcat_priv_data`:

```c
int ftx1_in_memory_mode;  /* set by ftx1_set_mem on successful MC */
```

Introduce a shared helper in `ftx1.c`:

```c
void ftx1_ensure_vfo_mode(RIG *rig) {
    if (flag is set) { send VM000; clear flag; }
}
```

Call it at the top of the VFO-state setters that the firmware actually affects:

- `ftx1_set_freq` (Main-VFO branch only)
- `ftx1_set_rptr_shift` (OS)
- `ftx1_set_ctcss_mode` (CT Main)
- `ftx1_set_ctcss_tone` (CN00 Main)
- `ftx1_set_dcs_code` (CN01 Main)
- new `ftx1_set_mode` wrapper that delegates to `newcat_set_mode`

Sub-targeted setters (`set_ctcss_sql` / `set_dcs_sql` using CN10/CN11) are deliberately left alone — the Sub side is never in Memory mode on this firmware, and an injection there would needlessly clobber a user who has intentionally parked Main in Memory mode while tuning Sub.

The flag-based approach avoids paying for a VM000 round trip on every setter call; it only fires when the driver knows the radio entered Memory mode via a prior `set_mem`.

### Verified end-to-end

```
# set_freq path
$ rigctl -m 1051 -r ... E 1
Rig command: F 7050000
Rig command: f
        7050000
# (radio in VFO mode at 7.050 — without the fix, FA would have been
# rejected and the radio would still be in Memory mode at 7.000)

# set_mode path (baseline: VFO-A at 14.200 USB)
$ rigctl -m 1051 -r ... E 1
Rig command: M USB 2700
Rig command: m
        Mode: USB
        Passband: 2700
# (VFO mode USB — without the fix, MD would have been a transient
# overlay that vanished on next MC recall)
```

Both tests verified against a live FTX-1 field head + Optima/SPA-1 (fw v1.12).

### Also clears the flag in set_channel

`ftx1_set_channel` always ends with the Main side in VFO mode (via the internal `VM000;` after `MC`), regardless of the caller's prior position. Added a `priv->ftx1_in_memory_mode = 0;` at the end of the restore block so a subsequent setter doesn't needlessly send another `VM000;`.

This concludes the FTX-1 memory-mode quirk series for this PR. All four of the main-side memory-mode issues (MW broken, clarifier/CTCSS/DCS readback missing, CT set in memory, MC blocking VFO commands) are now addressed or empirically-documented limitations.

KJ5HST:
@N0NB -- All checks and test passed so this can be merged whenever you want. I won't add anymore to this one. I've fixed all the stuff I'm aware of.

N0NB:
I am seeing the following set of warnings from GCC and MinGW on Debian 13:

``` cc
  CC       ftx1/ftx1_mem.lo
ftx1/ftx1_mem.c: In function 'ftx1_set_channel':
ftx1/ftx1_mem.c:235:9: warning: 'strncpy' output may be truncated copying 31 bytes from a string of length 128 [-Wstringop-truncation]
  235 |         strncpy(saved_fb, priv->ret_data, sizeof(saved_fb) - 1);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ftx1/ftx1_mem.c:240:9: warning: 'strncpy' output may be truncated copying 15 bytes from a string of length 128 [-Wstringop-truncation]
  240 |         strncpy(saved_md1, priv->ret_data, sizeof(saved_md1) - 1);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In function 'ftx1_read_channel_tone_state',
    inlined from 'ftx1_get_channel' at ftx1/ftx1_mem.c:658:9:
ftx1/ftx1_mem.c:444:9: warning: 'strncpy' output may be truncated copying 15 bytes from a string of length 128 [-Wstringop-truncation]
  444 |         strncpy(saved_vm, priv->ret_data, sizeof(saved_vm) - 1);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ftx1/ftx1_mem.c:449:9: warning: 'strncpy' output may be truncated copying 15 bytes from a string of length 128 [-Wstringop-truncation]
  449 |         strncpy(saved_mc, priv->ret_data, sizeof(saved_mc) - 1);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I don't see these warnings from clang on the same OS.

KJ5HST:
Reopening with the four review blockers addressed.

## Changes since the original close

**New commit:** `2b28d2e0a ftx1: silence GCC strncpy warnings, target Sub cursor in set_channel`

Two problems fixed in one commit:

### 1. `-Wstringop-truncation` warnings (per @N0NB's report)

Replaced four `strncpy(dst, src, sizeof(dst) - 1)` calls in `ftx1_mem.c` (lines 245, 250, 459, 464) with `snprintf(dst, sizeof(dst), "%s", src)`. This is the canonical fix GCC accepts — `snprintf` guarantees NUL termination and doesn't trigger the truncation heuristic.

Build is clean on Apple Clang with `-Wall -Wextra` (no warnings on `ftx1_mem.c`).

### 2. MC cursor targeting in `ftx1_set_channel`

Changed the final cursor-step command in `ftx1_set_channel` from `MC%06d;` → `MC1%05d;`. Since the whole write path uses VFO-B (`BM1...`), the cursor step must also target Sub; otherwise the radio's visible Main-VFO memory channel flips every time Hamlib commits a channel write. Verified on hardware 2026-04-18.

## Why `370dca2` (memory-mode auto-exit) is load-bearing

@N0NB asked a fair question in review about whether the memory-mode guard in `370dca2` was defensive-only. It is not. Re-probed on hardware 2026-04-18 with Main in `VM011`:

| SET command | In-memory-mode result | VFO state after `VM000;` | Stored memory |
|-------------|-----------------------|--------------------------|---------------|
| `FA007074000;` | `?` (rejected) | unchanged | unchanged |
| `OS01;`        | `?` (rejected) | unchanged | unchanged |
| `MD01;`        | echoed (accepted) | **reverted to MD04** | unchanged |
| `CT02;`        | echoed (accepted) | **reverted to CT01** | unchanged |
| `CN00005;`     | echoed (accepted) | **reverted to CN00014** | unchanged |

- `FA`/`OS` hard-reject with `?` — the auto-injection on `set_freq` / `set_rptr_shift` is load-bearing.
- `MD`/`CT`/`CN` transient-accept — firmware echoes the SET and read-back returns the new value, but the next `VM000;` silently reverts. That's **worse** than a `?`-reject because it hides the bug from callers. The auto-injection on `set_mode` / `set_ctcss_mode` / `set_ctcss_tone` / `set_dcs_code` converts that silent-revert bug into a working SET.

Removing `370dca2` would re-introduce a class of silent-revert bugs that are invisible to polling-style clients. Keeping it.

## MW status (unchanged)

Re-verified MW with exact-replay of the MR payload post-factory-reset. Still returns `?;`. VFO-B + BM workaround stands.

## Branch

Force-pushed `KJ5HST:ftx1-set-channel-mw-workaround` (`2b28d2e0a`). Based on current `origin/master` (`9de702a45`). Five commits total:

```
2b28d2e0a ftx1: silence GCC strncpy warnings, target Sub cursor in set_channel
370dca238 ftx1: auto-exit Memory mode before VFO-state setters (#2034)
130d733be ftx1: populate CTCSS/DCS fields in get_channel
c1a259406 ftx1: add clarifier support and document firmware limits in set_channel
4024cd5d0 ftx1: rewrite set_channel using VFO-B + BM (firmware MW bug) (#2034)
```

N0NB:
Unfortunately, GCC is still complaining.

``` cc
  CC       ftx1/ftx1_mem.lo
../../../../rigs/yaesu/ftx1/ftx1_mem.c: In function 'ftx1_set_channel':
../../../../rigs/yaesu/ftx1/ftx1_mem.c:245:47: warning: '%s' directive output may be truncated writing up to 128 bytes into a region of size 32 [-Wformat-truncation=]
  245 |         snprintf(saved_fb, sizeof(saved_fb), "%s", priv->ret_data);
      |                                               ^~
../../../../rigs/yaesu/ftx1/ftx1_mem.c:245:9: note: 'snprintf' output between 1 and 129 bytes into a destination of size 32
  245 |         snprintf(saved_fb, sizeof(saved_fb), "%s", priv->ret_data);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../../../rigs/yaesu/ftx1/ftx1_mem.c:250:49: warning: '%s' directive output may be truncated writing up to 128 bytes into a region of size 16 [-Wformat-truncation=]
  250 |         snprintf(saved_md1, sizeof(saved_md1), "%s", priv->ret_data);
      |                                                 ^~
../../../../rigs/yaesu/ftx1/ftx1_mem.c:250:9: note: 'snprintf' output between 1 and 129 bytes into a destination of size 16
  250 |         snprintf(saved_md1, sizeof(saved_md1), "%s", priv->ret_data);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../../../rigs/yaesu/ftx1/ftx1_mem.c: In function 'ftx1_get_channel':
../../../../rigs/yaesu/ftx1/ftx1_mem.c:459:47: warning: '%s' directive output may be truncated writing up to 128 bytes into a region of size 16 [-Wformat-truncation=]
  459 |         snprintf(saved_vm, sizeof(saved_vm), "%s", priv->ret_data);
      |                                               ^~
In function 'ftx1_read_channel_tone_state',
    inlined from 'ftx1_get_channel' at ../../../../rigs/yaesu/ftx1/ftx1_mem.c:673:9:
../../../../rigs/yaesu/ftx1/ftx1_mem.c:459:9: note: 'snprintf' output between 1 and 129 bytes into a destination of size 16
  459 |         snprintf(saved_vm, sizeof(saved_vm), "%s", priv->ret_data);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../../../rigs/yaesu/ftx1/ftx1_mem.c: In function 'ftx1_get_channel':
../../../../rigs/yaesu/ftx1/ftx1_mem.c:464:47: warning: '%s' directive output may be truncated writing up to 128 bytes into a region of size 16 [-Wformat-truncation=]
  464 |         snprintf(saved_mc, sizeof(saved_mc), "%s", priv->ret_data);
      |                                               ^~
In function 'ftx1_read_channel_tone_state',
    inlined from 'ftx1_get_channel' at ../../../../rigs/yaesu/ftx1/ftx1_mem.c:673:9:
../../../../rigs/yaesu/ftx1/ftx1_mem.c:464:9: note: 'snprintf' output between 1 and 129 bytes into a destination of size 16
  464 |         snprintf(saved_mc, sizeof(saved_mc), "%s", priv->ret_data);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

N0NB:
Meanwhile, MinGW has no complaints.  Go figure.  Sigh...

KJ5HST:
GCC 14 has gotten testy, hasn't it? Pushed `8a714a44b` — switched the four snapshot sites from `"%s"` to `"%.*s"` with an explicit precision of `sizeof(dst) - 1`. Hoping the bounded precision is enough to talk `-Wformat-truncation` down.

Builds clean on Apple Clang here, but I don't have a Debian/GCC 14 box handy to reproduce. If it's still grumbling on your end, let me know and I'll try another angle.

N0NB:
GCC is quiet.  Thanks for your patience, Terrell.

I don't have all that many physical machines but I do use QEMU a lot for virtual machines.

PR #2039, opened 2026-04-20
Give newcat.c some needed TLC

GeoBaltz:
Apply some modern techniques to handling of Yaesu rigs.

- Make `ncboolean` a `bool` typedef.  Use real boolean operations, parameters, and values with the compiler knowing the right way to handle them all.
- Remove the static flags for the rig type(i.e., is_ft710, etc.)  Replace them with macros that use the rig_model to give the same results.  Allows more than one Yaesu radio to be controlled by a single computer(think SO2R).
- Move many of the hard-coded constants used in bandwidth control out of the code and into structures attached to the rig; all in one place.  Makes the code much easier to handle, and allows for much more commonality between the many different feature sets.
- Fix some of the missing parts and incorrect behavior of the simulators - still a long way to go.

This PR does not address two issues in the original code:

1. What to do if a width request is larger that the rig can handle?  The current code is divided between returning an error and using the max width the rig is capable of. This now goes for using the max, but can be changed by swapping two lines.
2. Would the real FTDX-3000 please stand up.  The ftdx3000 and ftdx3000dm use the same rig_model, and supposedly differ by the results of the 'ID' command.  But I found that the FTDX-3000 Series CAT manual says that we're using the wrong ID.  I'll send a query to Yaesu tech support to see if they can sort it out.

For master only

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -14,6 +14,8 @@ Version 5.x -- future
         * Change FT1000MP Mark V model names to align with FT1000MP
         * Internal data structures moved out of rig_struct to individual
           heap buffers - Github issues #487 & #1445 (n3gb)
+        * Update bandwidth handling in rigs/yaesu/newcat.c to use data in
+          priv_caps. Moves lots of hard-coded data into tables.  (n3gb)
 
 Version 4.7.0
         * 2026-02-15
--- a/ReleaseNotes_5.0.md
+++ b/ReleaseNotes_5.0.md
@@ -12,6 +12,9 @@ changes to the Application Binary Interface(ABI).
 - shortfreq_t changed to int32? (may not happen)
 - Functions rig_get_conf(), rot_get_conf() and amp_get_conf() have been removed.
   See [issue #924](https://github.com/Hamlib/Hamlib/issues/924).
+- Clean up bandwidth handling in newcat.c.  Move data from code to priv_caps.
+- Convert ncboolean (in newcat.h) to real `bool` typedef. Clean up previous
+  pseudo-boolean code.
 
 ## Build changes
 - C compiler needs at least c11 capability.
--- a/rigs/yaesu/ft1200.c
+++ b/rigs/yaesu/ft1200.c
@@ -33,6 +33,20 @@
 #include "ft1200.h"
 #include "tones.h"
 
+// Shared with FTDX-3000/FTDX-3000dm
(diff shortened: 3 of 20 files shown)


GeoBaltz:
Email sent to Yaesu Amateur Tech support:

> Hi
> 
> I'm trying to verify some legacy code relating to the CAT control of the
> FTDX-3000 Series radios. Currently we have two models defined, the
> FTXD-3000 and FTDX-3000DM. Our code (Hamlib) thinks they are the same in
> basic features, and are distinguished by differing return values from
> the 'ID;' command.
> 
> But now I find that our idea doesn't match the FTDX-3000 Series CAT
> Operation Reference Book, as far as the ID; return value is concerned.
> 
> So, my questions are
> 
> 1. Do these models differ in ID values (0460 & 0462)?
> 
> 2. What is the correct ID value for all rigs in the FTDX-3000 series?
> 
> 3. Should we just get rid of the distinction and treat all -3000 Series
> rigs the same?
> 
> Thanks for any info you can offer
> 
> 73 n3gb
> 

And the response:

>  George:
> #3, Regarding CAT command options, treat all FTDX3000s the same.
> 

So I will generate another PR when I get this applied to newcat.c

N0NB:
You actually got a reply!

My flabber is gasted.

N0NB:
On the width, it seems setting it to the max the radio is documented to support is a reasonable expectation.  Returning an error here would just seem to result in aggravation for downstream applications.

GeoBaltz:
Your first response actually had me LOL :-) 

As for the width, that's sort of what I was thinking - principle of least surprise.  And now it will be consistent across the rigs.

N0NB:
Glad I could brighten your day.

Thanks for all of that work.  It is most appreciated.

ny4i:
I wonder if we should press our luck and write to icom asking them to standardize on 25 and 26 processing along with 07 D2 processing support. Implementing direct support in TR4W for the Icom network protocol was like playing whack-a-mole in the differences between the 7610, 705, 9700 and 7760. And it would be orto much to hope for to ask Icom to honor transceive with 25/26 commands rather than just sending our 03 and 04.

GeoBaltz:
In The Best of All Possible Worlds, they would have already done that. In the real world, I wouldn't hold my breath.

Just the differences between the IC-7300 and the IC-7300MK2 are enough to convince me that consistency is not high in their priorities.

PR #2040, opened 2026-04-22
Finish converting Icom clock routines

GeoBaltz:
Update the last four rigs.

Fix the commands for the IC-7160 - incorrect values.

For both master and 4.7 branches


N0NB:
Thanks, George.

I really appreciate these kinds of macro fixes that gather as much into common code as possible.

GeoBaltz:
I like them too.  I'll take a little extra data for less code any day.

Issue #2038, opened 2026-04-16
Icom: Filter commands (0x1A 0x03) fail in CWR mode (missing RIG_MODE_CWR in bitmasks)


cyfrax:
Tested with: 4.7.0 

ICOM set filter fails in CWR mode

I had a look at the code after github copilot tried to identify the issue . I appologize in advance if the code below isn't the real reason.  
BTW I noticed IC 7760 is still in alpha state. If you need a tester, let me know. I have one.

**Description:**
On Icom rigs, sending filter width commands via `rigctl/rigctld` (e.g., `W 500`) works correctly when the rig is in **CW** mode, but is ignored when the rig is in **CWR** (CW Reverse) mode.

**Technical Analysis:**
The issue lies in `rigs/icom/icom.c`. Several functions that handle DSP filter width calculations specifically omit `RIG_MODE_CWR` from their mode bitmasks.

1.  In `icom_get_dsp_flt()` (around line 2084), the check for CW-like modes excludes `CWR`:
    ```c
    else if (mode & (RIG_MODE_CW | RIG_MODE_USB | RIG_MODE_LSB | ...))
    ```
2.  In `icom_set_dsp_flt()` (around line 2127), the same omission exists:
    ```c
    else if (mode & (RIG_MODE_CW | RIG_MODE_USB | RIG_MODE_LSB | ...))
    ```

Because `CWR` is not matched, the functions fall through to the `else` block, which logs "unknown mode" and returns `RIG_OK` without sending the CI-V command to the radio.

**Suggested Fix:**
Update the mode bitmasks in `icom_get_dsp_flt` and `icom_set_dsp_flt` to include `RIG_MODE_CWR` alongside `RIG_MODE_CW`.

**Note:**
I am unable to compile or test Hamlib from source at this time, so I am reporting this as an issue for a maintainer to implement and verify.

N0NB:
Whenever you get a chance to build from source and test this and there is a problem, feel free to reopen this issue.

PR #2041, opened 2026-04-22
Fix 58

GeoBaltz:
Two minor patches
- Fix issue #2038
- Drop the remnants of FTDX-3000DM

Both to go into master
Patch 1 'Add CWR' also for 4.7.x


N0NB:
Thanks, George.

Nice catch on the FT-818 simulator.

PR #2043, opened 2026-04-23
Add Hamgeek uSDX

pizzolante:
Hello, it seems that getting this rig to work required a direct modification to this library.
It's quite similar to the TS-480, but there are completely random disconnections, and software like WSJT-X also reports errors, making normal use practically impossible

Tested with firmware 4.00d on the device: there's a slight delay in command handling (necessary for proper operation), and no other issues were observed


GeoBaltz:
The obvious first question is "If it is using the TS-480 command set (as far as it goes), why isn't it just another rig in rigs/kenwood?" Sort of like the Elecraft rigs?

pizzolante:
Just tried right now all the Elecraft rigs, all end up with `Communication timed out`
edit: selecting other Kenwood rigs (a few was selected), consistently ran into issues with commands being ignored (frequency changes, PTT) or even random disconnections.

What is the second question?

N0NB:
Hi Leo.

What George means is that as this radio uses Kenwood compatible commands that it probably uses some commands common to all radios in the `kenwood` backend and that in similar fashion to the Elecraft models, it can take advantage of existing code without writing new code.  Then the parts that are unique to the uSDX can be in its own file.  This is the preferred method of adding new models when it is possible.

pizzolante:
Thank you Nate, hope got that right

N0NB:
That looks much better, Leo.

I do get the following warning from GCC on Debian 13:

``` cc
  CC       hamgeek.lo
../../../../rigs/kenwood/hamgeek.c:71:9: warning: "BACKEND_VER" redefined
   71 | #define BACKEND_VER "20260423"
      |         ^~~~~~~~~~~
In file included from ../../../../rigs/kenwood/hamgeek.c:35:
../../../../rigs/kenwood/kenwood.h:33:9: note: this is the location of the previous definition
   33 | #define BACKEND_VER "20250515"
      |         ^~~~~~~~~~~
```

I think the easiest fix is for me to rename the macro to `USDX_BACKEND_VER` to avoid the conflict.

N0NB:
Thanks, Leo.

This will be added to the 4.7 series as well.

N0NB:
Added to the 4.7 branch and will be included in its snapshot tomorrow morning:  https://hamlib.sourceforge.net/snapshots-4.7/

pizzolante:
Thanks, and thanks as well to GeoBaltz for not insulting me yet :pray:
Just for information, a small typo here: https://github.com/Hamlib/Hamlib/blob/837e1ff82e8df30057562341bbae4afef8f31894/NEWS#L31

N0NB:
Fat fingers.  Ughh...  I'll get that fixed in a future push.

George's reviews are valuable as he will catch things I miss.  In this case your driver fits quite well in the Kenwood backend rather than stand-alone.  In any case, we thank you for the contribution.

PR #2046, opened 2026-04-26
Initialize motors of Skywatcher mount

hamarituc:
To use the Skywatcher rotator, both motors have to by initialized. When using the direct mode of the mounts hand control unit, this initialization is performed by the hand control. We just need to set the position.

When using the motor driver directly with a simple USB-RS232-adapter, this init command must be sent before we are able to steer the mount. This allows us to use the mount without cabeling up the whole hand control unit.


N0NB:
Thanks Mario.

I have also applied this to the `Hamlib 4.7` and will appear in 4.72 when it is released.

PR #2048, opened 2026-05-03
Fix errors found by glibc 2.43

GeoBaltz:
glibc implements some new parts of the C23 standard: (Taken from https://download.opensuse.org/tumbleweed/iso/Changes.20260430.txt)

  * For ISO C23, the functions bsearch, memchr, strchr, strpbrk, strrchr, strstr, wcschr, wcspbrk, wcsrchr, wcsstr and wmemchr that return pointers into their input arrays now have definitions as macros that return a pointer to a const-qualified type when the input argument is a pointer to a const-qualified type

and

  * The aforementioned change in ISO C23 of the declaration of bsearch, memchr, strchr, strpbrk, strrchr, strstr, wcschr, wcspbrk, wcsrchr, wcsstr, and wmemchr as const-preserving macros can lead to compilation issues in code not set up for it

This lead to some warnings about assigning returns from strchr(const, target)
  to non-const pointers, and then discovering that Hamlib was modifying
  those const parameters.

This commit aligns the const-ness of the offending parameters. Only affects
  internal functions in ampctl_parse.c, rigctl_parse.c and rotctl_parse.c,
  so API/ABI should not change.

For both master and 4.7.x

### DIFF
--- a/tests/ampctl_parse.c
+++ b/tests/ampctl_parse.c
@@ -124,14 +124,14 @@ struct test_table
                        int,
                        const struct test_table *,
                        const char *,
-                       const char *,
+                       char *,
                        const char *,
                        const char *,
                        const char *,
                        const char *);
     int flags;
     const char *arg1;
-    const char *arg2;
+    char *arg2;
     const char *arg3;
     const char *arg4;
     const char *arg5;
@@ -146,7 +146,7 @@ struct test_table
                                                     int interactive,    \
                                                     const struct test_table *cmd, \
                                                     const char *arg1,   \
-                                                    const char *arg2,   \
+                                                    char *arg2,   \
                                                     const char *arg3,   \
                                                     const char *arg4,   \
                                                     const char *arg5,   \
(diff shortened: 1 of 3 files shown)


N0NB:
Very interesting, George.

I've not forgotten the PRs.  Just things got real busy here on the farm so I'm a bit swamped ATM.  Rain is forecast--whether it falls is another story--this week, so I should get some time to test and merge.

GeoBaltz:
No problem; do what needs to be done.

This one definitely surprised me - I did a test build and couldn't 
figure out when I had changed those routines. Then I remembered I had 
run a system patch. Reading the changelog brought me up to date.

It will probably be a while before this glibc hits any of the normal 
distros - but my Tumbleweed system gets all the bright shiny stuff quickly.

73 n3gb

N0NB:
Clean compilation on Debian 13 and Arch with GCC 16.1.1 dated 30 Apr so that is fairly recent.

Thanks, George.

PR #2051, opened 2026-05-05
More cleanup/consolidation in newcat.c

GeoBaltz:
Move more rig-specific constants to newcat_width_info.
Set width_info pointer in common code before rig selection.
Eliminates some if()s in newcat_[gs]et_rx_bandwidth(), even merges 2 complete stanzas.

For master only.


N0NB:
Thanks, George.

PR #2052, opened 2026-05-06
alinco/dxsr8: fix serial configuration, timing, and control bugs

toxicgumbo:
- serial_stop_bits corrected from 1 to 2 per protocol documentation; incorrect stop bits caused systematic transaction timing errors
- post_write_delay increased from 0 to 100ms to allow radio processing time between commands
- timeout increased from 200ms to 600ms to accommodate slower responses
- retry reduced from 3 to 1 to prevent command flooding on failure
- DXSR8_LEVEL_ALL advertised RIG_LEVEL_RF but implemented ATT and PREAMP; corrected capability flags accordingly
- set_mode NAR command acknowledgment timeout treated as non-fatal; radio executes the command correctly but does not reliably return OK within the timeout window — logs WARN instead of propagating error
- get_mode RR_NAR query removed; radio does not respond reliably causing systematic timeouts — normal passband returned always
- get_level ATT case 2 returned 10 instead of 20 (copy-paste error)
- get_level RFPOWER case 3 never matched radio response of 2, leaving val->f uninitialized in QRP mode

### DIFF
--- a/rigs/alinco/dxsr8.c
+++ b/rigs/alinco/dxsr8.c
@@ -38,7 +38,7 @@
 
 #define DXSR8_FUNC (RIG_FUNC_FAGC|RIG_FUNC_NB)
 
-#define DXSR8_LEVEL_ALL (RIG_LEVEL_RFPOWER|RIG_LEVEL_RF)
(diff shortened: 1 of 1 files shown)


N0NB:
Thanks, Jeff.

Nice to see a contribution from the home of the Wildcats.

PR #2056, opened 2026-05-15
Fix FT-710 VFO swap after cross-band tune

### LINKED ISSUE #2055: Yaesu FT-710:  VFO-A/B swaps when changing frequencies cross-band
**Describe the bug**
A single `F <hz>` command that crosses a ham-band boundary causes an unnecessary, temporary swap to VFO-B and back to VFO-A after tuning. This is especially an issue when the VFO-B is on a different band than the target, as the radio will end up swapping bands _three_ times for 1 cross-band command.

**To Reproduce**
 
- Hamlib 4.7.1 (`SHA d042479a9`, 64-bit Linux). Also present on current `master` (`b43eca4e4`).
- Yaesu FT-710, model 1049, backend revision `20241118.7`.
- Stock `rigctld` invocation: `rigctld -m 1049 -r <device> -s 38400 -T 127.0.0.1 -t 4532`.

1. Start rigctld with `-vvvv` so the trace is captured (see attached)
2. Settle the rig on a known band on VFO-A: `printf 'F 7074000\n' | nc -q1 127.0.0.1 4532`.
3. Cross-band tune: `printf 'F 14233000\n' | nc -q1 127.0.0.1 4532`.
4. Observe the FT710:
   - Switch bands, tuned to the target frequency
   - After ~200ms, swaps to VFO-B
   - After ~50ms swaps back to VFO-A and remains

[rigctld-trace.log](https://github.com/user-attachments/files/27809335/rigctld-trace.log)

**Expected behavior**
The extra swap should not occur, there does not appear to be any reason for it. Sending commands to the FT-710 when using `-m 1035` (FT-991) does not cause the same behavior, due to a guard specifically for the FT-991. 

**Desktop:**
 - OS: Linux
 - Client application: N/A 
 - Client application version: N/A

**Additional context**
The "after band select re-read things" cache-refresh block in `newcat_set_freq`
(`rigs/yaesu/newcat.c`, the live block following the disabled `#if 0` section)
calls:

```c
rig_get_mode(rig, RIG_VFO_MAIN, &tmp_mode, &tmp_width);
rig_get_mode(rig, RIG_VFO_SUB,  &tmp_mode, &tmp_width);
```

The FT-710's `targetable_vfo` is `RIG_TARGETABLE_FREQ` only
(`rigs/yaesu/ft710.c:178`). `rig_get_mode` in `src/rig.c:3124-3169` therefore
takes the non-targetable branch for the second call: `caps->set_vfo(VFO_SUB)`,
`caps->get_mode(VFO_SUB)`, `caps->set_vfo(curr_vfo).

Models with `RIG_TARGETABLE_MODE` (mostIcom backends, some Kenwoods) avoid the swap
and the FT-991/FT-891 are excluded from the surrounding BS block entirely for some
reason (`!is_ft991 && !is_ft891`), so they also do not exhibit this.

The comment two lines above the offending call already concedes the read is
optional: *"reading both VFOs is really only needed for rigs with just one VFO
stack"*.

mdeltito:
Avoid re-reading the sub-VFO mode after band change on the FT-710 unless we are operating in split mode.

---

This might be too specific of a fix, but I cannot verify the change on any other device.

Perhaps gating this behind `RIG_TARGETABLE_MODE` would be more appropriate?

```c
!(caps->targetable_vfo & RIG_TARGETABLE_MODE) || cachep->split
```

trace after fix [rigctld-trace-fix.log](https://github.com/user-attachments/files/27810196/rigctld-trace-fix.log)

Fixes #2055

### DIFF
--- a/rigs/yaesu/newcat.c
+++ b/rigs/yaesu/newcat.c
@@ -1561,7 +1561,14 @@ int newcat_set_freq(RIG *rig, vfo_t vfo, freq_t freq)
         }
 
         rig_get_mode(rig, RIG_VFO_MAIN, &tmp_mode, &tmp_width);
-        rig_get_mode(rig, RIG_VFO_SUB, &tmp_mode, &tmp_width);
+        // FT-710 does not advertise RIG_TARGETABLE_MODE, so reading mode
+        // on the non-current VFO forces a physical VFO swap (rig.c).
+        // Skip the post-BS cache refresh unless we are in split mode, where
+        // VFO_SUB is the TX VFO.
+        if (!is_ft710 || cachep->split)
+        {
+            rig_get_mode(rig, RIG_VFO_SUB, &tmp_mode, &tmp_width);
+        }
 
         if ((target_vfo == 0 && tmp_freqA == freq)
                 || (target_vfo == 1 && tmp_freqB == freq))


N0NB:
Thanks, Mike.

I've also cherry-picked this into the Hamlib-4.7 branch and will be in the forthcoming 4.7.2 release (probably next month).

GeoBaltz:
Looking back at the 710 caps, I wonder why RIG_TARGETABLE_MODE isn't set for the FT-710(and a couple of other Yaesu rigs).  Mike, is there any chance you could try this (with the original code) and add ' | RIG_TARGETABLE_MODE' to the .targetable_vfos parameter in ft710_caps?
73 n3gb

mdeltito:
@GeoBaltz absolutely, I will report back

mdeltito:
@GeoBaltz I did some more testing today of `master` (with the fix from this PR) vs [this branch](https://github.com/Hamlib/Hamlib/compare/master...mdeltito:Hamlib:ft710-target-mode?expand=1) (fix reverted + cap flag added).

I confirmed the originally-reported problem is, in fact, resolved by adding `RIG_TARGETABLE_MODE` for the FT-710.

However, it adds a bunch of latency to the command/response for reasons I have yet to trace, but almost certainly caused by the state cleanup that occurs here. The timings below show the difference which is significant - with `RIG_TARGETABLE_MODE` set, the target frequency on a cross-band tune is not set on the rig until 3s+ after the band selection has changed.

(in both cases, the band modes were manually tuned to 14000/7000 prior to the test) 

### `master` branch

band changes immediately, ~200ms later the frequency adjusts to the target

```
$ time printf 'F 14201000\n' | nc -q1 127.0.0.1 4532
RPRT 0

________________________________________________________
Executed in    1.69 secs      fish           external
   usr time    1.30 millis  468.00 micros    0.83 millis
   sys time    2.02 millis  350.00 micros    1.67 millis

$ time printf 'F 7074000\n' | nc -q1 127.0.0.1 4532
RPRT 0

________________________________________________________
Executed in    1.69 secs      fish           external
   usr time    1.34 millis  440.00 micros    0.90 millis
   sys time    2.40 millis  596.00 micros    1.80 millis

```

### `ft710-target-mode` branch

band changes immediately, frequency does not adjust on the rig until 3+s later

```
$ time printf 'F 14201000\n' | nc -q1 127.0.0.1 4532
RPRT 0

________________________________________________________
Executed in    4.19 secs      fish           external
   usr time    3.29 millis    0.00 micros    3.29 millis
   sys time    2.10 millis  989.00 micros    1.11 millis

$ time printf 'F 7074000\n' | nc -q1 127.0.0.1 4532
RPRT 0

________________________________________________________
Executed in    4.19 secs      fish           external
   usr time    1.39 millis    0.00 micros    1.39 millis
   sys time    2.25 millis  845.00 micros    1.41 millis

```

If you have thoughts about where/what to test or debug in order to narrow it down, I'm happy to do more digging. I can also open a dedicated issue for revisiting the 710's caps reporting if you'd like.

I do wonder if this was noticed previously in #1339 which caused those cap flags to be removed.

PR #2061, opened 2026-05-17
Support building hamlib on Android

goodspeed34:
The old Android.mk was broken and messy, https://github.com/Hamlib/Hamlib/pull/2058 tried to fix it but the diffs aren't sustainable as it's hard to verify and easy to break. This pull request tackles that by building hamlib with GNU build system and NDK toolchains.


goodspeed34:
Added a Github CI workflow file to let us know when it breaks.

N0NB:
I approved the workflows (this seems to be a requirement for all first time contributors).

goodspeed34:
The current branch is okay to merge now. No further commits will make to the branch. The last commit ([bbde232](https://github.com/Hamlib/Hamlib/pull/2061/commits/bbde232a0c86c6b87d728952c0fb6e5a453d45c4)) was cherry-picked from https://github.com/goodspeed34/Hamlib/tree/feature/java-binding branch which introduces java binding. Hopefully it could make its way to another pull request.

Hamlib correctly works on Android but requires a pty helper that fakes a serial port for Hamlib. However, that doesn't require root permission and it's flexible enough to enable access to Bluetooth Serial and USB Serial. Currently hamlib could be built into a ~8MiB aar file with arm64, arm, x86_64, x86 shared library.

I wrote a simple app that polls data from rigs connected via usb otg and works great to me. https://github.com/goodspeed34/HamlibAndroidExample/

Screenshot_20260518-170208_Hamlib Example
Screenshot_20260518-165914_Hamlib Example

configure.ac
```diff
@@ -153,6 +153,46 @@ sys/select.h glob.h ])
...
+    AC_CHECK_HEADER([android/sensor.h],
+    [
+        have_android_sensor=yes
+    ],
+    [
+        have_android_sensor=no
+    ])
+
+    CXXFLAGS="$CXXFLAGS -stdlib=libc++"
+    LDFLAGS="$LDFLAGS -lc++"
```
dg0yt, line 179: The Android NDK offers alternatives for the C++ runtime library. IIUC this hardcodes a particular choice. However, a C++ app might want to use another one, and mixing usually creates problems. (https://developer.android.com/ndk/guides/cpp-support#one_stl_per_app)
Would it be possible to just let the C++ compiler drive the linker?

goodspeed34: That's a very nice point and did some research. In the [Android documentation](https://developer.android.com/ndk/guides/cpp-support), NDK provides theree runtimes: `libc++`, `system` and `none`. However, there is only one functional C++ runtime. `none` provides no STL functions,  the depreicated `system` runtime  only provides `new/delete` functions (those are standard), So I would assume that nobody would intentiionally use those two other alternatives, and even twhen hey do, it would still link fine against libc++.

dg0yt: In terms of the NDK choices, what does `-stdlib=libc++`/`-lc++` select?  
Obviously, it is not `none`.
`system` is `libstdc++.so` (and it is not GNU libstdc++).
`libc++` is either `libc++_shared.so` or `libc++_static.a`.

goodspeed34: `-lc++` refers to `libc++_shared.so` when compiling shared libraries and `libc++_static.a` when compiling statically.

dg0yt: Okay. I guess I will revisit this in the future, but it shouldn't block the merge.

N0NB:
I appreciate the addition of Android CI, but as it seems interest in that platform is limited, I hope I can count on you and @dg0yt to respond if those builds fail at some point.

goodspeed34:
> I appreciate the addition of Android CI, but as it seems interest in that platform is limited, I hope I can count on you and @dg0yt to respond if those builds fail at some point.

No problem, I will try my best maintaining that target.

N0NB:
Thanks, William (Gong).  Android needed an update so this is most welcome.

N0NB:
I've also cherry-picked this into the Hamlib-4.7 branch so it will be included in 4.7.2 in a few weeks.

Issue #2062, opened 2026-05-19
FT-710 should advertise as RIG_TARGETABLE_MODE


mdeltito:
Much like the FTdx10, the FT-710 is a single-receiver, dual-VFO rig. I have confirmed that it supports the `MD0`/`MD1` forms for mode read and write against the non-active VF0. Advertising it as `RIG_TARGETABLE_MODE` will enable codepaths for the rig that handle dual-VFO scenarios more gracefully.

PR #2063, opened 2026-05-19
fix(ft710): add RIG_TARGETABLE_MODE

### LINKED ISSUE #2055: Yaesu FT-710:  VFO-A/B swaps when changing frequencies cross-band

mdeltito:
The FT-710 firmware supports the targeted MD0/MD1 forms for mode read and write. Advertising the capability lets hamlib dispatch mode operations on the non-current VFO directly.

The FT-710 like other similar Yaesu models is a single-receiver with multiple VFO, and thus only has a single bandpass filter slot. So we must apply the same workaround as the FTDX10 when refetching state after a band-select, and avoid issuing the (unsupported) `SH1` command.

This reverts the previous change in #2056 as this is a more direct/complete fix for the physical VFO swap reported in #2055.

Fixes #2055, #2062


rigs/yaesu/newcat.c
```diff
@@ -9487,7 +9480,7 @@ int newcat_get_rx_bandwidth(RIG *rig, vfo_t vfo, rmode_t mode, pbwidth_t *width)
 
     if (sh_command_valid)
     {
-        if (is_ft2000 || is_ftdx10 || is_ftdx3000)
+        if (is_ft2000 || is_ftdx10 || is_ft710 || is_ftdx3000)
```
mdeltito, line 9483: fwiw without the same workaround, we get:

```txt
****4:newcat.c(1861):newcat_get_mode entered
...snip...
*****5:newcat.c(9409):newcat_get_rx_bandwidth entered
newcat.c(8142):newcat_valid_command returning2(1)
******6:newcat.c(8489):newcat_set_vfo_from_alias entered
******6:newcat.c(8547):newcat_set_vfo_from_alias returning(0)
******6:newcat.c(10301):newcat_get_cmd entered
newcat_get_cmd: Rig busy - retrying 1 of 3: 'SH1;'
newcat_get_cmd: Rig busy - retrying 2 of 3: 'SH1;'
newcat_get_cmd: Rig busy - retrying 3 of 3: 'SH1;'
newcat_get_cmd: Rig busy - retrying 4 of 3: 'SH1;'
******6:newcat.c(10534):newcat_get_cmd returning(-9) Command rejected by the rig
*****5:newcat.c(9497):newcat_get_rx_bandwidth returning(-9) Command rejected by the rig
****4:newcat.c(1922):newcat_get_mode returning(-9) Command rejected by the rig
```

and after:

```txt
****4:newcat.c(1861):newcat_get_mode entered
...snip...
*****5:newcat.c(9409):newcat_get_rx_bandwidth entered
newcat.c(8142):newcat_valid_command returning2(1)
******6:newcat.c(8489):newcat_set_vfo_from_alias entered
******6:newcat.c(8547):newcat_set_vfo_from_alias returning(0)
******6:newcat.c(10301):newcat_get_cmd entered
******6:newcat.c(10534):newcat_get_cmd returning(0)
*****5:newcat.c(10089):newcat_get_rx_bandwidth returning(0)
****4:newcat.c(1922):newcat_get_mode returning(0)
```

N0NB:
Thanks for the update, Mike.  Also cherry-picked into 4.7.

PR #2066, opened 2026-05-21
Fix ambiguous syntax of includes

dforsi:
This PR replaces the angled brackets < > with double quotes " " for private includes.

Fixed by hand in gqrx.c and with:
perl -pe 's/<bandplan.h>/"bandplan.h"/' -i rigs/*/*.c perl -pe 's/<serial.h>/"serial.h"/' -i rigs/*/*.c
perl -pe 's,<hamlib/rig.h>,"hamlib/rig.h",' -i rigs/*/*.h

### DIFF
--- a/rigs/dummy/gqrx.c
+++ b/rigs/dummy/gqrx.c
@@ -26,10 +26,10 @@
 #include <math.h>
 #include <stdbool.h>
 
-#include <hamlib/rig.h>
-#include <serial.h>
-#include <misc.h>
-#include <cal.h>
+#include "hamlib/rig.h"
+#include "serial.h"
+#include "misc.h"
+#include "cal.h"
 #include "idx_builtin.h"
 
 #define BACKEND_VER "20250718.2"
--- a/rigs/harris/harris.h
+++ b/rigs/harris/harris.h
@@ -27,7 +27,7 @@
 #ifndef _HARRIS_H
 #define _HARRIS_H
 
-#include <hamlib/rig.h>
+#include "hamlib/rig.h"
 
 struct harris_priv_data {
     freq_t current_freq;
--- a/rigs/harris/prc138.h
+++ b/rigs/harris/prc138.h
@@ -27,7 +27,7 @@
 #ifndef _PRC138_H
 #define _PRC138_H
 
-#include <hamlib/rig.h>
+#include "hamlib/rig.h"
 
 #define PRC138_MODES (RIG_MODE_USB | RIG_MODE_LSB | RIG_MODE_CW | RIG_MODE_AM)
 #define PRC138_VFO   (RIG_VFO_A)
--- a/rigs/icom/ic736.c
+++ b/rigs/icom/ic736.c
@@ -22,7 +22,7 @@
 #include <stdlib.h>
 
 #include "hamlib/rig.h"
-#include <bandplan.h>
+#include "bandplan.h"
 #include "icom.h"
 
 
--- a/rigs/icom/ic737.c
+++ b/rigs/icom/ic737.c
@@ -22,7 +22,7 @@
 #include <stdlib.h>
 
 #include "hamlib/rig.h"
-#include <bandplan.h>
+#include "bandplan.h"
 #include "icom.h"
 
 
--- a/rigs/icom/ic738.c
+++ b/rigs/icom/ic738.c
@@ -22,7 +22,7 @@
 #include <stdlib.h>
 
 #include "hamlib/rig.h"
-#include <bandplan.h>
+#include "bandplan.h"
 #include "icom.h"
 
 
--- a/rigs/kenwood/ts930.c
+++ b/rigs/kenwood/ts930.c
@@ -20,7 +20,7 @@
  */
 
 #include "hamlib/rig.h"
-#include <bandplan.h>
+#include "bandplan.h"
 #include "kenwood.h"
 
 


dforsi:
Why "Workflow runs completed with no jobs" ?

N0NB:
Is it because the workflows only key on the master branch?

rigs/dummy/gqrx.c
```diff
@@ -26,10 +26,10 @@
 #include <math.h>
 #include <stdbool.h>
 
-#include <hamlib/rig.h>
-#include <serial.h>
-#include <misc.h>
-#include <cal.h>
+#include "hamlib/rig.h"
```
dg0yt, line 29: This should be 
```suggestion
#include "../../include/hamlib/rig.h"
```

rigs/dummy/gqrx.c
```diff
@@ -26,10 +26,10 @@
...
+#include "serial.h"
```
dg0yt, line 30: ... and there is no `rigs/dummy/serial.h`.

dg0yt (commented):
I think this should not be applied blindly. It is not "private" vs. "public".

`#include "..."` searchs in an implementation-defined way first. Effectively all implementations search the including file's directory first. This is super useful to avoid system headers when you want headers from the local project. But you have to use a proper relative path.

rigs/dummy/gqrx.c
N0NB: In practice this doesn't seem to be an issue as the build system sets paths to the local directory structure:

``` console
$ make -j4  --no-print-directory V=1 | grep gqrx
depbase=`echo gqrx.lo | sed 's|[^/]*$|.deps/&|;s|\.lo$||'`;\
/bin/bash ../../libtool  --tag=CC   --mode=compile gcc -DHAVE_CONFIG_H -I. -I../../../hamlib/rigs/dummy -I../../include/hamlib  -DIN_HAMLIB -I../../../hamlib/include -I../../../hamlib/src -I../../../hamlib/lib -I../../include  -I../../../hamlib/include/hamlib -D_TIME_BITS=64 -D_FILE_OFFSET_BITS=64  -Wall -g -O2  -MT gqrx.lo -MD -MP -MF $depbase.Tpo -c -o gqrx.lo ../../../hamlib/rigs/dummy/gqrx.c &&\
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I../../../hamlib/rigs/dummy -I../../include/hamlib -DIN_HAMLIB -I../../../hamlib/include -I../../../hamlib/src -I../../../hamlib/lib -I../../include -I../../../hamlib/include/hamlib -D_TIME_BITS=64 -D_FILE_OFFSET_BITS=64 -Wall -g -O2 -MT gqrx.lo -MD -MP -MF .deps/gqrx.Tpo -c ../../../hamlib/rigs/dummy/gqrx.c  -fPIC -DPIC -o .libs/gqrx.o
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I../../../hamlib/rigs/dummy -I../../include/hamlib -DIN_HAMLIB -I../../../hamlib/include -I../../../hamlib/src -I../../../hamlib/lib -I../../include -I../../../hamlib/include/hamlib -D_TIME_BITS=64 -D_FILE_OFFSET_BITS=64 -Wall -g -O2 -MT gqrx.lo -MD -MP -MF .deps/gqrx.Tpo -c ../../../hamlib/rigs/dummy/gqrx.c -o gqrx.o >/dev/null 2>&1
/bin/bash ../../libtool  --tag=CC   --mode=link gcc -Wall -g -O2    -o libhamlib-dummy.la  dummy_common.lo dummy.lo rot_dummy.lo rot_pstrotator.lo netrigctl.lo netrotctl.lo flrig.lo trxmanager.lo amp_dummy.lo netampctl.lo tci1x.lo aclog.lo sdrsharp.lo quisk.lo gqrx.lo  
```

Note that I build out of tree in a sibling directory to the Git checkout.  In practice I'm not sure it matters whether `hamlib/rig.h` is enclosed in angle brackets or double quotes.  It has always seemed to work either way.

N0NB: @dg0yt Is my explanation satisfactory?

dg0yt: TBH I thought it was a final word, not an explanation. There is no dispute that both variants search in the include dirs (`-I` et ak.) So of course headers are still found after changing from `<...>` to `"..."`.

My point is that it doesn't make sense to adopt a style which forces the compiler to look in the wrong place first. It can make sense to switch for local includes, but then it also makes sense to use the proper relative path.

N0NB: We've had discussion around this in the past and never came to a firm conclusion that I recall.  I can see both ways making sense, but I seem to recall that angle brackets imply system installed headers while quotes imply local headers.  Given that Hamlib may be compiled on a system that doesn't have its headers installed in the system `include` directory hierarchy it seems to me that it should err on the side of local headers (there is also the case where system installed headers are well out of date).

@GeoBaltz and @mikaelnousiainen please share your thoughts.

GeoBaltz:
It has been discussed before, but let's go back to the C standard.  From 
the C23 version, slightly paraphrased

2. of the form #include <char-sequence>\n searches a sequence of 
implementation-defined defined places for a header identified uniquely 
by the sequence between < and >, and replaces the directive by the 
contents of that header. How the places are specified or the header 
identified are implementation defined.

3. of the form #include "char-sequence"\n causes the replacement of that 
directive by the entire contents of the source file identified by the 
sequence between the " delimiters. The named source file is searched for 
in an implementation-defined manner. If this search is not supported, or 
if the search fails, the directive is reprocessed as if it read #include 
<char-sequence>\n with the identical contained sequence from the 
original directive.


So the difference between <> and "" really just adds an additional 
search operation for finding the header.  The implementation (i.e., gcc, 
clang, MSVC, etc) is what makes the distinction between "system" headers 
and "local" or "unique" ones.  What is needed is to decide which header 
should take precedence in which type of build.

I think for building Hamlib, we always should prefer the "local" flavor 
- the header changes may be the reason for the build, after all. The 
current source does this in two ways - adding the extra search 
("header.h"), and by modifying the implementation-defined manner (adding 
-I options), which can muddy the distinction quite a bit.

For building apps that use Hamlib, we want to insure consistency between 
the headers and the library, so I think we only want to search the 
"system" path.  That means for includes within installed includes we 
should use <>, and let the Makefiles take care of modifying the search path.

Hope I haven't confused the issue even more.

N0NB:
Thanks, George.

To be clear, I think we also want the utilities in the `tests` directory to use local includes.

dg0yt:
> For building apps that use Hamlib, we want to insure consistency between the headers and the library, so I think we only want to search the "system" path.  That means for includes within installed includes we should use <>, and let the Makefiles take care of modifying the search path.

Installed headers are in a fixed layout.  
In the user code, there will be `#include <...>` because it must use the search path.  
In the installed headers. `#include "..."` is normally a precise way to refer to other headers from the same installation, so users don't actually mix a system-wide installation ('/usr/include/hamlib`) with a local installation ('/home/dg0yt/hamlib/include/hamlib`), subject to order of search path entries.

N0NB:
Thanks, Daniele.

N0NB:
Cherry-picked into master as well.

PR #2067, opened 2026-05-22
Directly open the file descriptor from mkstemp

dg0yt:
Bring back the `s` in `mkstemp`: Don't close the created file descriptor before the changes are actually written.

`fclose` is documented to close the underlying file descriptor, so no additional calls are needed before return.

Amends 56efa155dc0bfdcfa042690ac5780d9c5777f67b.

Disclaimer: I cannot test at runtime at the moment.

### DIFF
--- a/src/settings.c
+++ b/src/settings.c
@@ -1161,9 +1161,14 @@ HAMLIB_EXPORT(int) rig_settings_save(const char *setting, void *value,
     }
 
     int fd = mkstemp(template);
-    close(fd);
-    fptmp = fopen(template, "w");
+    if (fd == -1)
+    {
+        rig_debug(RIG_DEBUG_ERR, "%s: error creating %s: %s\n", __func__,
+                  template, strerror(errno));
+        return -RIG_EIO;
+    }
 
+    fptmp = fdopen(fd, "w");
     if (fptmp == NULL)
     {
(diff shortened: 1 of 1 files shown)


N0NB:
As I see this, the error checking of the `fd` value returned by `mkstemp()` is a good addition, however, while I don't see any practical difference replacing

``` c
fptmp = fopen(template, "w");
```

with

``` c
fptmp = fdopen(fd, "w");
```

I do like the elimination of the possible race condition between `close()` and `fopen()`.  Why not use the new template value to open `fptmp` as before?

dg0yt:
The error checking is just the sugar on top. The `fdopen` is the key point.

> Why not use the new template value to open `fptmp` as before?

After `mkstemp`, the file is already opened for writing. Re-opening again without re-using the file descriptor just doesn't seem right. I assume this caused the "Permission error" described in https://github.com/Hamlib/Hamlib/commit/56efa155dc0bfdcfa042690ac5780d9c5777f67b, at least in combination with the original mingw-w64 behaviour described in https://sourceforge.net/p/mingw-w64/mailman/mingw-w64-public/thread/20221016011230.46939-1-delthas%40dille.cc/#msg37721376.

N0NB:
Ah, thank you very much for that link.

N0NB:
Merged into Hamlib-4.7 branch as well.

PR #2070, opened 2026-05-26
Fix power2mW for Xiegu X6100

ddorian1:
`power2mW` reports wrong values for the Xiegu X6100. This fixes the problem for me. On the way I cleaned up some non-existing modes for this rig.


N0NB:
Thanks, Johannes.

Merged into master and Hamlib-4.7 branches.

PR #2074, opened 2026-06-03
Adjust x6100 S-meter LUT 

Nannk:
This  PR changes LUT for s-meter calculation.

Disclamer, im using cfw that is based on the older version of the official fw.
[gdyuldin v0.33.4 gui](https://github.com/gdyuldin/x6100_gui/releases/tag/v0.33.4)
[gdyuldin r9 base patch](https://github.com/gdyuldin/x6100_base_patch/releases/tag/r9)

Not sure if the reason calculated s-meter values were wrong is cfw, old fw, differences between G108 and X6100 or me misinterpreting what `get_level STRENGTH` returns, so i would like to ask someone with x6100 running normal fw to test this.

Also note that Flrig shows correct s-meter values.


N0NB:
It's hard to know if anyone will come along and test with the stock firmware.  I think this should probably be included in 4.7.2 which will be released on 15 June.

Nannk:
I have flashed stock 1.1.9 firmware with 1.2.0 gui.
Neither my LUT nor G108 LUT returns the correct value. 
G108 LUT is close, but both the slope and offset are wrong.

~~Surprisingly, the value `icom_get_level` spits out, depends on what gui version is running, not  base fw version.~~ 
Dummy me, missed `cat.cpp` in the lvgl_gui repo.

Not sure what to do.

Nannk:
It seems [this equation](https://github.com/gdyuldin/x6100_gui/blob/main/src/cat.cpp#L635) corresponds to what stock gui sends. And that is (supposedly) how icom rigs do it. I guess xiegu introduced this change when adding wfview support.

So, technically, the correct s-meter LUT for stock x6100 gui would be:
https://github.com/Hamlib/Hamlib/blob/3036bd98d30626b4cafb2b95c3bfd648f4522dd2/rigs/icom/ic7200.c#L58-L63

The lvgl_gui uses flrig implementation of the s-meter protocol (older version, i guess?).

Edit. I can confirm the ic7200 LUT works with stock gui.

N0NB:
Thanks!

I did a rebase to simplify the commits into one so just the needed changes are merged.

N0NB:
Does this close #2075?

PR #2076, opened 2026-06-04
Add TCI 2.0 backend for Expert Electronics SunSDR radios

jfrancis42:
Add TCI 2.0 backend for Expert Electronics SunSDR radios

New Hamlib backend (RIG_MODEL_TCI2, dummy family model 12) for TCI
(Transceiver Control Interface) protocol version 2.0.  Tested against
my SunSDR2 Pro running ExpertSDR3 v1.1.7 on Arch Linux.

TCI is a WebSocket push protocol, so this lives in rigs/dummy/.  The
backend does an HTTP/1.1 upgrade handshake over Hamlib's existing TCP
layer (no external WebSocket library needed), drains the server's READY
dump on connect, and keeps a local state cache updated by every incoming
push.  Reads come from the cache or issue explicit queries; writes send
the appropriate TCI command.

What works:
  - Frequency: VFO A and VFO B (set/get), split TX frequency
  - Mode: all modes from the server's MODULATIONS_LIST
    (AM, LSB, USB, CW, NFM, WFM, RTTY, DIGL, DIGU on SunSDR2 Pro)
  - PTT: RIG_PTT_ON, RIG_PTT_ON_MIC, RIG_PTT_ON_DATA (TCI source:
    default, Mic, Vac)
  - Split and VFO switching
  - RIT and XIT (enable/offset)
  - Levels: AF volume, RF drive, squelch, NB threshold, AGC
    (off/fast/normal), CW keyer speed, S-meter, SWR, TX power
  - Functions: NB, NR, ANF, mute, tune (ATU)
  - CW: send_morse (CW_MACROS with TCI escape encoding), stop_morse,
    wait_morse
  - Config params: trx (TRX index, default 0), txsource, digl_offset,
    digu_offset

Protocol quirks worth knowing:
  - The server sends everything lowercase; the backend uppercases the
    keyword on receive so sscanf format strings work.
  - Mode strings are built entirely from the server's MODULATIONS_LIST,
    so only modes the server supports are ever sent.  Requesting a mode
    the server doesn't know returns -RIG_EINVAL.
  - set_mode does NOT write RX_FILTER_BAND: ExpertSDR3 auto-applies a
    sane filter on MODULATION change, and empirical testing shows that
    sending RX_FILTER_BAND right after MODULATION corrupts the server's
    filter state.  get_mode reports the server's actual filter from the
    READY-dump cache.
  - VFO_LOCK (TCI 2.0) is server-to-client only; set_func LOCK returns
    -RIG_ENAVAIL.
  - CW_MACROS_SPEED and CW_KEYER_SPEED are separate; get_level KEYSPD
    returns the macro speed from the READY cache.

Default path: 127.0.0.1:50001


N0NB:
To fix the CI failure, `RIG_MODEL_TCI2` will need to be added to `bindings/python/test_Hamlib_class.py`.

rigs/dummy/tci2.c
```diff
@@ -0,0 +1,2638 @@
...
+    struct tci2_priv *priv = (struct tci2_priv *)STATE(rig)->priv;
+    char buf[TCI2_BUFLEN];
+    int retval;
+    time_t start = time(NULL);
+    int tx_seen = 0;
+
+    ENTERFUNC;
+
+    /* Phase 1: wait up to 2 s for TX to engage */
+    while (!tx_seen && time(NULL) - start < 2)
```
dforsi, line 2476: It seems that there is no sleep in this loop and the in loop at line 2495 below and in `ws_recv_frame()`.
Do they busy loop?

rigs/dummy/tci2.c
```diff
@@ -0,0 +1,2638 @@
...
+/* -------------------------------------------------------------------------
+ * Capabilities
+ * ---------------------------------------------------------------------- */
+
+struct rig_caps tci2_caps =
+{
+    RIG_MODEL(RIG_MODEL_TCI2),
+    .model_name     = "TCI 2.0",
+    .mfg_name       = "Expert Electronics",
+    .version        = "20240112.0",
```
dforsi, line 2554: Is the date 12th January 2024 correct?

rigs/dummy/tci2.c
jfrancis42: Good catch -- looked at it closely.

The two loops in tci2_wait_morse and the loop inside ws_recv_frame
don't busy-spin: they all bottom out in Hamlib's read_block(), which
calls port_wait_for_data() -- a select() with the port's timeout
(default 2 s).  And TCI's ambient traffic (RX_SENSORS at 200 ms when
sensors are enabled, which we do at open) keeps each iteration well
under that timeout in practice, so the loops effectively re-check
state at ~200 ms cadence driven by real I/O, not CPU spin.

Your question made me re-look at tci2_wait_morse, though, and there's
a real thread-safety bug in the updated code (about to replace this
PR) there I'd missed: when the optional audio sidechannel is enabled,
the audio reader thread becomes the sole WebSocket reader, and
tci2_wait_morse calling ws_recv_frame() directly was racing it.  Fixed
in the new PR -- when the reader thread is running, tci2_wait_morse
poll-waits on the cached PTT state (which the reader thread keeps
fresh via tci2_process_message on every text frame).  When the reader
thread isn't running it keeps the original ws_recv_frame loop, since
that path is what drives the cache.

Thanks for the review.

rigs/dummy/tci2.c
jfrancis42: No. That's on me.

jfrancis42:
Closing in favor of a near complete rewrite after more experience with
the driver.

This PR was a first-cut of the TCI 2.0 backend.  In the time since
opening it I built out an optional audio sidechannel (so a separate
process can handle TCI audio while rigctld owns the single allowed TCI
client connection), and the work involved enough rethinking of the
backend's threading and lifecycle that the result is better presented
as a single coherent driver than as a follow-up.

The new code has the same backend, plus the optional sidechannel gated
by a config param.  The driver still works as a normal CAT-only Hamlib
backend if you don't enable the audio sidecar proxy.

Sorry for the churn.

N0NB:
No need to apologize.  It's good to have this sort of interaction and refactoring.  Open the next PR when you're ready, Jeff.

Issue #1940, opened 2025-11-01
Network audio and I/Q data streaming support


mikaelnousiainen:
This is the main issue for ARDC Hamlib improvement project [Milestone 3 - Network audio and I/Q data streaming support](https://github.com/Hamlib/Hamlib/milestone/26) - plans and discussion.

# Project background

* See ARDC grant: https://www.ardc.net/apply/grants/2025-grants/hamlib-stability-enhancements-and-sdr-transceiver-support/
* See Open Collective page: https://opencollective.com/hamlib-stability-enhancements-and-sdr-transceiver-support

The project backlog is available in: https://github.com/orgs/Hamlib/projects/1 (still in progress)

# Description

**Goal:** Expand Hamlib to Support Modern Network-Controlled Radio Transceivers

**Issue:** Modern amateur radio transceivers increasingly rely on network interfaces for interfacing
with computers rather than traditional serial ports or sound cards. Hamlib has no support for
network-based audio or I/Q data transfer, which are necessary to use network-controlled
transceivers.

**Solution:** This milestone implements the infrastructure to support audio and I/Q data transfer (converted to a well-defined format) when operating network-controlled transceivers (e.g. FlexRadio or Icom).

The deliverable of this milestone is to Implement support for two-way audio and I/Q data transfer via UDP network packet based streaming protocols with appropriate data conversion to a selection of well-defined data formats (e.g. PCM audio and integer or float I/Q data)

The design will be based on prior work in Hamlib: there is already support for waterfall spectrum data streaming via UDP packets.

#  Development Process

* GitHub issues in the main Hamlib repository will track the project's progress, milestones, and deliverables transparently.
* All code contributions will be made as GitHub Pull Requests against the main Hamlib repository for final community review and comments

# Design

TBD - Will be available early 2026.

mikaelnousiainen:
Design and development is in progress, will publish first designs soon.

KJ5HST:
@mikaelnousiainen,

A while back, I built a network audio library that features:

- TCP server/client for bidirectional audio streaming
- Thread-safe ring buffer with jitter compensation
- Binary packet protocol with CRC32 validation
- Platform-aware virtual audio device detection
- Configurable audio format (default: 48kHz, 16-bit, mono)

[...]

PR #2078, opened 2026-06-07
TCI 2.0: Add full audio and IQ support via external sidecars

jfrancis42:
# TCI 2.0: Add full audio and IQ support via external sidecars

This PR extends the TCI 2.0 backend (`rigs/dummy/tci2.c`) with complete
audio and IQ streaming support through an external sidecar
architecture, plus a small reusable sidecar library
(`include/hamlib/sidecar.h`, `src/sidecar.c`) that's intended to drive
the same protocol for other backends later.

> ## 📄 Read the protocol spec first
>
> The **complete, canonical, language-agnostic sidecar wire-protocol
> specification** lives in the sidecar repo:
>
> **https://github.com/jfrancis42/hamlib-audio-sidecar/blob/main/PROTOCOL.md**
>
> Anyone implementing a sidecar in any language — or evaluating the
> protocol design that's the heart of this PR — should start there.
> This PR description summarizes the protocol enough to understand the
> Hamlib-side code changes, but `PROTOCOL.md` is the authoritative
> reference (frame format, every stream type, IEEE-754 encoding of
> float-valued control frames, reserved ranges, forward-compat rules,
> reference decoding snippets in C / Python / Rust / Go).

## Status of this PR — please read

This PR is being **published for an overall architecture review**, not
as a final implementation:

- It largely works. JS8Call, fldigi, and WSJT-X all run against
  ExpertSDR3 through this pipeline; on-air RX and TX are both
  functional.
- **There are still minor latent audio issues being worked out.** The
  control flow, framing, and CAT integration are stable; the
  remaining problems are around the edges of the audio path. None of
  them are blockers for the architecture discussion.
- The sidecar **protocol and library** (the `sidecar_*` API in this PR)
  are what we'd most like feedback on. Once the protocol settles,
  expect a follow-up that lands the additional backends below.

**Held back from this PR for now:** complete, end-to-end working
backends for **KiwiSDR (model 102, WebSocket)** and **RTL-SDR (model
202, librtlsdr)** have both been written and verified with live
signals (HF AM/SSB/CW, WFM stereo broadcast, airband AM, NBFM weather,
KiwiSDR IQ mode, etc.). They use exactly the sidecar API in this PR
and are exercised heavily by the same Python sidecar codebase. They're
deliberately kept out of this PR until the sidecar design is finalized
here, so we're not asking reviewers to evaluate three backends in
parallel.

## What This Adds

**Audio support** (bidirectional RX/TX) and **IQ support** (RX-only)
for TCI 2.0, compatible with JS8Call, fldigi, WSJT-X, GNU Radio, and
other ham radio software.

The implementation uses **external sidecar processes** connected via
TCP to rigctld, keeping Hamlib's C codebase free of platform-specific
audio dependencies while enabling full TCI functionality.

## Architecture

```
ExpertSDR3 (TCI server, port 50001)
        |
        | WebSocket (one connection)
        |
    rigctld (:4532 CAT, :4534 audio sidechannel, :4535 IQ sidechannel)
        |
        +--- sidecar (Python, separate repo)
                |
                +---> PulseAudio virtual sinks
                |       hamlib-rx (RX), hamlib-tx (TX)
                |       -> JS8Call / fldigi / WSJT-X / etc.
                |
                +---> GNU Radio ZMQ endpoints (any OS)
                |       audio RX/TX, IQ RX/TX (four sockets)
                |
                +---> SoapySDR direct hardware (bypasses rigctld;
                        for radios without a Hamlib backend yet)
```

**rigctld** remains the sole TCI WebSocket client to ExpertSDR3. Audio
and IQ streams are proxied as **length-framed binary TCI frames** over
independent TCP sidechannels (ports 4534 and 4535) to an external
Python sidecar process. The sidecar handles platform-specific audio
plumbing and the various downstream transports.

## Why External Sidecars?

1. **No new Hamlib dependencies** — rigctld stays pure C with existing
   dependencies; no libpulse / coreaudio / wasapi.
2. **Platform flexibility** — Linux, Windows, and macOS sidecars can
   use completely different audio APIs without touching Hamlib code.
3. **Process isolation** — audio/IQ bugs don't crash rigctld (and the
   CAT connection with it).
4. **Fast iteration** — audio pipeline changes are Python edits +
   restart, not Hamlib rebuild + redeploy.
5. **Generic protocol** — the same wire protocol works for TCI,
   KiwiSDR, RTL-SDR, and (we expect) anything else with a usable
   IQ-or-audio sidechannel. Backends only need to learn the sidecar
   library's `sidecar_send_*` and `sidecar_emit_*` calls.

## Wire Protocol (rigctld ↔ sidecar)

**Canonical spec:** the complete, language-agnostic protocol
specification is in the sidecar repo at
**https://github.com/jfrancis42/hamlib-audio-sidecar/blob/main/PROTOCOL.md**.
The summary below is just enough context for the code review here;
anything that disagrees with `PROTOCOL.md` is a bug in this summary,
not in `PROTOCOL.md`.

Pure binary, length-framed TCI frames in both directions:

```
64-byte header (16 little-endian uint32 words) + 0..N payload bytes
```

| Offset | Field       | Audio/IQ frames                    | Control frames |
|--------|-------------|------------------------------------|----------------|
| 0      | receiver    | trx index                          | trx index      |
| 4      | sample_rate | Hz                                 | 0              |
| 8      | format      | 0=int16 1=int24 2=int32 3=float32  | 0              |
| 20     | length      | samples in payload                 | control value  |
| 24     | stream_type | 0=IQ 1=RX_AUDIO 2=TX_AUDIO         | 3=TX_CHRONO 4=PTT_STATE ... |
| 28     | channels    | 1 (audio) or 2 (IQ)                | 1              |
| 32..63 | reserved    | zero-filled                        | zero-filled    |

Stream types 0–5 are data streams (IQ, RX_AUDIO, TX_AUDIO, TX_CHRONO,
PTT_STATE, TX_IQ). Types 6–19 are control frames (MODE, FREQ, SPLIT,
FILTER, AGC_LEVEL, NR_LEVEL, NB_LEVEL, NOTCH, RF_GAIN, SQUELCH,
PREAMP, ATT, CW_PITCH, APF). Types 20–24 are reserved for future
Hamlib use; 100–999 for future Hamlib growth; 1000–1999 for
vendor-specific extensions. Sidecars MUST silently skip unknown
types — that's how the protocol stays forward-compatible without
version negotiation.

**Float-valued control frames** (NR_LEVEL, NB_LEVEL, RF_GAIN,
SQUELCH, APF) put the IEEE-754 single-precision bit pattern in the
`length` field. See `sidecar.c` and the sidecar repo's
[`PROTOCOL.md`](https://github.com/jfrancis42/hamlib-audio-sidecar/blob/main/PROTOCOL.md)
for the encoding details and decoding snippets in C / Python / Rust /
Go.

**Why binary-only?** Audio and IQ payloads contain arbitrary bytes
(including 0x0A). Early prototypes that mixed text lines
(`TX_CHRONO 0 512\n`) with binary frames on the same socket
corrupted audio whenever a sample value happened to contain a
newline. Going all-binary with TCI's existing header eliminates that
entire class of bug.

### IQ sidechannel

`iq_port` and `iq_rate` config params; `tci2_iq_init / cleanup /
accept` follow the same shape as the audio path. The audio poll thread
is the sole WebSocket reader and dispatches binary frames by
`stream_type` — IQ frames go to the IQ sidecar fd, audio frames + TX
chrono to the audio sidecar fd. The WebSocket receive buffer was
raised to 32 KB to accommodate IQ frames at 384 kHz.

**RX IQ only:** the TCI 2.0 spec defines `IQ_STREAM = 0` as
unidirectional. There is no spec-defined way to push IQ samples back
to the radio for transmission, so the IQ sidechannel and sidecar are
RX-only. TX of baseband audio remains available via the audio path
(`STREAM_TX_AUDIO`).

### Audio sidechannel

- **Single-reader queue** between the WebSocket reader thread and the
  CAT thread. Text frames → 64-slot ring buffer; binary frames →
  sidecar fds. CAT thread pops from the queue.
- **TCP-stream reassembly** of sidecar TX frames. Partial `recv()`s
  used to forward malformed WebSocket frames (ExpertSDR3 silently
  dropped them → 0 W TX). The reader now accumulates a 64-byte header
  before emitting.
- **`tci2_send` is thread-safe**; audio and CAT threads share the
  socket under `priv->ws_mutex`.
- **TX-from-sidecar pumped every iteration**, not only on WebSocket
  poll timeout — the old code mangled chunks when sensors streamed
  continuously and poll() never timed out.
- **TX_CHRONO as binary `STREAM_TX_CHRONO`** (not text) — keeps the
  sidechannel binary-only.
- **PTT edges as `STREAM_PTT_STATE`** inside `tci2_set_ptt()`. rigctld
  is the authoritative PTT source; the sidecar uses these edges to
  flush stale TX capture buffer at the start of each transmission.
- **Idempotent audio listen-socket setup** across TCI READY /
  reconnect cycles.

### Config parameters

| Token | Type | Default | Purpose |
|-------|------|---------|---------|
| `trx` | int | 0 | TRX index (0-based) for multi-RX rigs |
| `txsource` | string | `default` | `default` / `mic` / `vac`; overridden by PTT type |
| `digl_offset`, `digu_offset` | int | 0 | DIGL/DIGU freq offset, Hz (0–4000) |
| `audio_port` | int | 0 | Audio sidechannel TCP port (0 = disabled) |
| `iq_port` | int | 0 | IQ sidechannel TCP port (0 = disabled) |
| `iq_rate` | int | 192000 | TCI IQ stream rate (48k / 96k / 192k / 384k) |

`audio_port` and `iq_port` are independent; either or both can be
enabled.

## Sidecar Library (`include/hamlib/sidecar.h` + `src/sidecar.c`)

The sidecar wire format and the partial-send retry logic were
duplicated across early prototypes. This PR consolidates them into a
small library that any Hamlib backend can use:

- `sidecar_init_port(port)` — listen on `localhost:port`, return fd
- `sidecar_send_rx_audio(fd, ...)`, `sidecar_send_rx_iq(fd, ...)`
- `sidecar_send_tx_chrono(fd, ...)`, `sidecar_send_ptt_state(fd, ...)`
- `sidecar_emit_mode(fd, ...)`, `sidecar_emit_freq(fd, ...)`,
  `sidecar_emit_agc_level(fd, ...)`, etc. — one per control frame type
- `sidecar_close_port(fd)`

The library handles framing, retries `EAGAIN` on partial sends, and
encodes float values into the `length` field correctly. `tci2.c` is
the first user; the upcoming KiwiSDR and RTL-SDR backends use it
identically.

See `docs/sidecar-api.md` for the C API reference and the integration
checklist, and the sidecar repo's `PROTOCOL.md` for the canonical
language-agnostic spec.

## Sidecar Implementations (not part of this PR)

The sidecar processes live in a separate repo
(https://github.com/jfrancis42/hamlib-audio-sidecar). The Python
codebase ships two sidecars plus a runtime control CLI and a
lifecycle script:

1. **`hamlib_sidecar_linux.py`** — Linux. Selectable user-side
   backend: `pulseaudio` (virtual sinks `<prefix>-rx`, `<prefix>-tx`
   for demodulated audio in/out) or `pulseaudio-iq` (raw IQ as a
   stereo soundcard, L=I, R=Q). Bidirectional.
2. **`hamlib_sidecar_portable.py`** — Linux / macOS / Windows. GNU
   Radio ZMQ bridge (four endpoints: audio RX/TX + IQ RX/TX), or
   SoapySDR direct hardware bypassing rigctld entirely.
3. **`hamctl`** — runtime CLI: frequency, mode, AGC, every DSP level,
   one-shot or live-bargraph S-meter, audio routing (PulseAudio
   loopback management), favorites, band-aware `tune <khz>` verb that
   auto-applies mode/width/AGC from a JSON band plan.
4. **`hamlib.sh`** — lifecycle script: `start` / `stop` / `restart` /
   `status` for the Linux sidecar, with `--radio {rtlsdr|kiwisdr|tci}`
   and per-radio defaults. Brings up rigctld + sidecar + PulseAudio
   loopback together.

Both sidecars sit on top of one shared library
(`hamlib_sidecar_common.py`) which contains a built-in
software-defined-radio receiver and transmitter (SSB / AM / FM
mono+stereo / CW) plus the DSP suite. For IQ-only backends that
library is also the canonical Python reference for parsing the wire
protocol.

## Things That Will Bite You

A few non-obvious lessons accumulated during the implementation;
useful to know in advance if you're reviewing or extending this
code:

- **Binary on the sidechannel is non-negotiable.** Any text framing
  (newline-terminated records, `:` separators, etc.) will eventually
  hit an audio sample that contains the framing byte. Don't mix.
- **Send loops, not single sends.** Always loop `send()` until the
  full frame is on the wire. `sidecar_send_frame()` in `src/sidecar.c`
  is the reference; partial sends used to drop bytes mid-frame and
  produce an endless "bogus header" resync on the sidecar.
- **The `channels` field is load-bearing.** Stereo IQ payload size is
  `length × 2 × sample_bytes`, not `length × sample_bytes`. Sidecars
  that ignore `channels` will read half the bytes and either run at
  half rate or scribble over the next frame.
- **Threading discipline.** With both audio and CAT writing to the
  WebSocket, one mutex around the send path is mandatory. Without it
  you get rare interleaved-byte corruption that ExpertSDR3
  silently drops.
- **Single WebSocket reader.** ExpertSDR3 (and TCI in general) does
  not tolerate two clients on the same channel. rigctld must own the
  one WebSocket; sidecars get TCP from rigctld.
- **Skip unknown stream types.** Forward-compat depends on it. The
  sidecar library and every reference implementation already do this.
- **Frame-boundary state in the sidecar.** Decimators, resamplers,
  AGC envelopes, deemphasis — all stateful filters must thread their
  state across frame boundaries. Stateless per-frame DSP produces a
  transient at the frame-rate cadence (15–150 Hz) that sounds like
  pumping or thumping. Bit us several times during sidecar
  development.

## Verified End-to-End

- **JS8Call** (14.079 MHz, 40 m): CQ via this pipeline gets replies.
  On-air decode confirmed by remote receivers. There's a ~2 s
  pre-roll that turned out to be intrinsic to JS8Call's slot
  scheduler, not this pipeline.
- **1 kHz tone TX** (`pacat` via PulseAudio sidecar): clean carrier
  at 14.0790 MHz, -27 to -33 dBm sustained, RFPOWER_METER ~50 mW.
  Zero silence frames mid-transmission.
- **RX audio** (`parec` from `hamlib-rx.monitor`): 8000 samples/sec,
  ~98% nonzero, peaks tracking real signals.
- **IQ stream** (portable sidecar, gnuradio user-side backend):
  effective rate matches configured `iq_rate`.

## Build Notes

The Hamlib build system silently uses stale convenience archives when
only `tci2.c` is touched. If you don't see your changes take effect,
rebuild `libhamlib-dummy.la` and `libhamlib.la` explicitly. There's a
worked example in `rigs/dummy/README-TCI-2.0.md` if you hit this.

## Testing

```bash
# CAT
echo 'f' | nc -w 1 localhost 4532    # dial freq
echo 'm' | nc -w 1 localhost 4532    # mode + width

# Audio sidecar running: RX audio flows into hamlib-rx.monitor
timeout 3 parec --device=hamlib-rx.monitor --rate=8000 --channels=1 \
    --format=s16le > /tmp/rx_check.raw
# Expect 24000 samples, peak well above 100, fraction nonzero > 95%

# Portable sidecar (gnuradio user-side backend): RX IQ on ZMQ
python3 -c "
import zmq, time
s = zmq.Context.instance().socket(zmq.SUB)
s.setsockopt(zmq.SUBSCRIBE, b'')
s.setsockopt(zmq.RCVTIMEO, 2000)
s.connect('tcp://localhost:5552')
end = time.time() + 3
msgs = samples = 0
while time.time() < end:
    try: m = s.recv()
    except zmq.Again: continue
    msgs += 1
    samples += len(m) // 8   # complex64 = 8 bytes
print(f'{msgs} ZMQ messages, {samples} complex samples in 3 s')
print(f'effective rate: {samples/3:.0f} Hz')
"
# Expect effective rate ≈ iq_rate (default 192000)
```

Ultimate test: **call CQ with JS8Call on a band where someone is
listening**. If you get replies, the entire audio pipeline is working
on-air.

## Related Documentation

In this repo:

- **TCI Protocol PDF:** `TCI_Protocol.pdf`
- **TCI backend reference:** `rigs/dummy/README-TCI-2.0.md`
- **Sidecar architecture (Hamlib side):** `README.audio-sidecar.md`,
  `docs/sidecar-api.md`, `docs/EXTENSIBILITY-IMPLEMENTED.md`

In the sidecar repo
(https://github.com/jfrancis42/hamlib-audio-sidecar):

- **[`PROTOCOL.md`](https://github.com/jfrancis42/hamlib-audio-sidecar/blob/main/PROTOCOL.md)**
  — **canonical wire-protocol specification (language-agnostic).
  Anyone implementing a sidecar in any language should start here.**
- **[`README.md`](https://github.com/jfrancis42/hamlib-audio-sidecar/blob/main/README.md)**
  — top-level overview, install, 30-second quick start.
- **[`USERS.md`](https://github.com/jfrancis42/hamlib-audio-sidecar/blob/main/USERS.md)**
  — operator guide: JS8Call / WSJT-X / fldigi / gqrx setup, the
  `hamlib.sh` + `hamctl` workflow, audio routing, troubleshooting.
- **[`DEVELOPERS.md`](https://github.com/jfrancis42/hamlib-audio-sidecar/blob/main/DEVELOPERS.md)**
  — writing a new sidecar in any language, or extending the Python
  codebase.
- **[`linux.md`](https://github.com/jfrancis42/hamlib-audio-sidecar/blob/main/linux.md)**
  — internals of the Linux PulseAudio virtual-device backend.

External references:

- **Working C++ TCI reference:** https://github.com/maksimus1210/TCI
- **Working Python TCI reference:** eesdr-tci library
  (`pip install eesdr-tci`)

## License

Same as Hamlib (LGPL 2.1+).

## Acknowledgements

- Expert Electronics for the TCI 2.0 protocol specification.
- eesdr-tci and the C++ TCI reference for showing how a working TCI
  client behaves.
- The Hamlib project for the foundation this builds on.


mikaelnousiainen:
Just a quick note that I'm in the process of developing audio and I/Q data streaming support in #1940 for Hamlib - the project has been lagging a bit, but I'm determined to publish my plans and an initial "reference implementation" soon! We already got a proof-of-concept implementation of this working with FlexRadio (also to be released later).

N0NB:
Jeff, you might consider adding your excellent PR description as a `README-TCI-2.0.md` file in the dummy directory (it doesn't need to be distributed in the source archive by adding it to the `EXTRA_DIST` in `Makefile.am`).  Once closed the text of a PR is generally hidden away and I think others will benefit by having such a file in the repository to call attention to the extra capability of this new backend.

jfrancis42:
That is an excellent idea (adding a readme with the PR info). I'll do exactly that.

N0NB:
I'm going to hold off releasing 4.7.2 a few days to get this merged in.

GeoBaltz:
I would really like this better with a clearer distinction between the parts - the protocol definition/operation, the device handling, and especially the audio sidecar.  I think some kind of audio support will be needed in a future Hamlib, so separating the audio handling might be useful.  Maybe this is a first step to get something going (i.e., 4.7.x), but it could use some generalization moving forward.

When this is merged into 4.7.x, will it be optional?  Like `./configure -with-tci2=yes`.  Or marked as experimental, and subject to change?

jfrancis42:
I've added the README as N0NB suggested. I'm brand-new to Hamlib, so I don't feel I have an informed opinion on the "separation of audio" point that GeoBaltz brought up, but I'm happy to work with the group to define what that might look like. I will say that it's functionality is almost certain to change as other TCI device support is merged in (FlexRadio being the obvious elephant in that room - but I don't own one to experiment with) and as other requirements and limitations are discovered as we dig into this new space. I also haven't quite figured out the best way to add in the IQ streaming support, so I've held off on that, but it's an obvious feature that will need to be added once the requirements are better understood. It's not hard to get the IQ out of the radio, what's hard is figuring out what to do with it once you have it so that existing tools can make use of it. Or perhaps a whole new standard needs to be defined for that, but I'm loathe to do that without a much better understanding of the existing landscape of IQ-based tools. One obvious requirement would be gnuradio support, but that's likely not the only way forward.

mikaelnousiainen:
@jfrancis42 Thanks for this great contribution, it's a good start for TCI protocol support!

I'm also trying to think about the design and Hamlib future, which leads me to thinking similar thoughts as the comments from @GeoBaltz above: separate protocol handling routines and proper audio handling.

I'll be publishing a proposal (and later an implementation) of Hamlib audio and I/Q streaming routines that will have a generic UDP-based format for transferring these types of data (from any type of rig). I try to get the spec proposal out in a couple of days now in #1940. This will certainly help with TCI protocol audio and I/Q streaming.

Additionally, I see it would benefit Hamlib to start grouping SDR protocols under their respective manufacturers (both directory-wise and adding new Hamlib manufacturer definitions) - and NOT put all of them in the "dummy" directory. I do not see the benefit of grouping unrelated protocols here (even if it has been the practice).

For these reasons, I would not necessarily publish this backend in the Hamlib 4.x series, but maybe leave things to mature for the 5.x releases (current master branch).

For 4.x releases, there could be a "lite-version" of this backend if people see urgent need for it - but right now I see the audio sidecar being very much against what Hamlib tries to be: an access method that is not specific to any rig/protocol.

GeoBaltz:
I think I would like to see this in 4.7.x, if only to gauge user/developer interest in TCI.  I do worry about adding audio handling into Hamlib(especially multi-platform support) as it is **Mission Creep**, not just **Feature Creep** that is normal for any S/W project.  However, the TCI spec does require it, and as mentioned above it has to be via the same route as CAT; I don't think we have much choice.

As far as where to put it, I think the protocol and audio handling could go in /src, if they can be cleanly carved out. I don't know how much commonality the rig handlers would have - they might all just go into rigs/TCI, or some such. This is all something that could worked on for 5.0 and beyond.

Maybe for 4.7.2 we label this as a "Technology Preview" or some such weasel words so we can spend some time rearranging things later.

mikaelnousiainen:
I agree with what @GeoBaltz you've stated here. If we could somehow make at least the audio sidecar separate from the rest of the code, labeling it as "tech preview" (as suggested) and possibly build it only if specifically requested (for the ones knowing how it really works), it might be fine.

Otherwise I'm afraid we'd be introducing an interface Hamlib users start using, but which is implemented in a way that is not future proof or manufacturer/rig-independent - which is definitely something I'd like to avoid.

And what I'd like to stress: 5.0 release will have audio and I/Q streaming in one form or another, so the TCI backend can then implement those interfaces properly. Please hang in there, it won't take much longer for me to post the full design for review.

jfrancis42:
I'm happy to go with the flow and relocate the code to a different place, or whatever. I've already got the audio sidecar hosted elsewhere (in my personal github), just to keep it "mentally separate" from Hamlib, but I'm happy to move it over to Hamlib if that's the decision. I found a race condition in my code and a few other things I'm working on right now, so there will be at least one more update to the PR soon. The update should also remove the periodic "tick" sounds in TX where the buffer runs out of data. I think I've found the cause and I'm trying a fix for it right now. Might be another day or two. Day job gets in the way of real work.

N0NB:
I originally considered including this PR in 4.7.2 but I understand the concerns.  As such I think at this late time I can wait until after the 4.7.2 release to merge this into Hamlib-4.7 for release with 4.7.3.

Calling this a tech preview should be fine and we will stress that this is not the final intended API for handling audio I/O.  I would like to get this in the stable series and solicit feedback from downstream developers which should help development of the 5.0 API.

jfrancis42:
Incoming changes since the last PR: the rigctld<->sidecar socket is now purely binary in both directions. The old version mixed text lines (TX_CHRONO 0 512\n) with binary RX_AUDIO frames on the same socket, which falls apart the moment an audio sample value happens to be 0x0A -- the demuxer eats the newline and shreds the frame. So TX_CHRONO is now a small length-framed binary control frame using the same 64-byte TCI header the audio frames already use, with the stream_type field telling the sidecar what kind of frame it's looking at. One parser, one mental model, no more demux bugs.

While I was in there, I also added a second control frame: PTT_STATE. rigctld emits it from tci2_set_ptt() on every PTT edge so the sidecar knows when a TX cycle starts and can flush whatever stale silence its capture pipeline accumulated while idle. Without that, the first few seconds of every transmission were leftover idle-buffer junk being shipped to the radio. Now the sidecar gets a clean PTT_STATE: ON, dumps its capture buffer, and starts fresh. End-to-end test: real JS8 CQ on 40m through this pipeline got replies from remote stations, so the whole thing decodes properly on the air. Both the in-tree README-TCI-2.0.md and the sidecar repo README are updated to describe the current protocol, the per-OS sidecar plan, and a hook for eventually carrying IQ over the same channel as a new stream_type.

It's way cleaner now, and all of the audio glitches are gone. Also lower latency between PTT and start of audio TX. Next task is to integrate (optional) IQ into the flow.

jfrancis42:
New PR update. Added IQ support. Much to my surprise, the TCI v2.0 specification does NOT including TX IQ, only RX IQ. I had been working under the seemingly reasonable assumption that both were supported. The new PR reflects this unfortunately reality.

I did write new sidecars (not part of this PR, they're hosted separately on my personal github) to support the IQ functionality with GNU Radio, as well as GNU Radio support for the existing non-IQ TX/RX. Both have been extensively tested with GNU Radio and include a couple of simple demo programs.

I do recognize that a new audio and IQ standard is forthcoming, and I'm happy to work to conform to that standard once it appears, but for now, this works extremely well. I intend to purchase an SDRPlay RSPDuo at some point in the next few months which will be useful for testing both IQ RX and TX (though it's not TCI-compliant, they use their own proprietary API, which will require a full new driver).

N0NB:
I've an RSP DX laying around that I got some years back.  It might be useful for testing, etc.

jfrancis42:
I've almost got a new PR ready with a bunch of new stuff. I realized that a lot of the code I've built for TCI is going to have to be repeated. First thing I did was "genericize" all of the sidecar audio, then split it out into sidecar.c that can be shared as a library among various SDR rig drivers. Then I built drivers for rtl-sdr and kiwisdr (mostly because I own both of those). Both produce only IQ output, so I wrote new sidecars that are basically GUI-less SDR radios. Then I cleaned up the sidecar API significantly and got everything mostly working (still working on some audio artifacts). Now I have a full SDR framework that normalizes both baseband and IQ, presents them in a standardized sidecar API, and have sidecars that will TX and RX both baseband and RX audio to and from the three SDR back-ends (SunSDR, RTL-SDR, and KiwiSDR) and present that data to the user in various formats. If the SDR data stream is baseband audio, it can translate it to virtual audio devices, feed it natively to GNU Radio, or present it as SoapySDR. If it's IQ, it can hand the data to GNU Radio as IQ, present it as a virtual stereo audio device with I as left channel and Q as right, present it as SoapySDR, or feed it into the sidecar SDR, which can demodulate it into AM, narrow FM, wide FM (mono or stereo), CW, USB, or LSB. The Hamlib calls for frequency, filters, AGC, mode, etc are all intercepted and sent to the sidecar where they are applied to the sidecar SDR. In theory, this should all work for TX, also, but none of my SDRs with IQ will transmit in IQ mode, so the TX features are all untested (and not quite complete). RX, though, is working quite well. This now provides a complete framework for handling control, baseband audio, and IQ to/from any current or future SDR device. I have a few lingering audio issues to clean up, but I'm actually rather shocked at how well this all works, and how it now presents a fully documented, fully extensible API for baseband/IQ work. Using this framework, it should be relatively simple to extend to pretty much any SDR on the market and handle sound in a way that is extremely minimally invasive on the Hamlib side, but provides developers a nice API to work with. The current sidecars are all Linux (I don't own a Windows machine), and they're all written in Python. Once the research phase is over, we'll obviously need Windows and OSX sidecars (I have a Mac for work, so I can work on that one later), and they'll also need to be simple downloadable/runnable binaries so that users don't have to install a fleet of python pip libraries to get audio, which means probably C or C++. That's fine, the appropriate libraries are available for both. I've also written up a migration plan for that. I have to travel out of state for a couple of days, but will get this posted no later than the end of the week. I may go ahead and drop a PR with the 90% working version sooner, just for review and feedback. It's been a productive few days. BTW, FM stereo over IQ from my cheap RTL-SDR dongle decoded into wideband FM in the sidecar actually sounds beautiful. I'm very pleased.

I suspect I'm starting to tread into mikaelnousiainen's work. Ping me if you want to work together on this. I don't want to step on toes, but it was the obvious next step to build this into a generic audio framework for handling baseband and IQ for SDRs.

mikaelnousiainen:
Thanks @jfrancis42, I will check your latest changes. Please have a look at the design I posted in #1940 - a PR with an initial implementation will follow that (hopefully latest next week). Of course the design and the implementation will likely need adjustments, but I feel I've covered most use cases for a low-level component, such as Hamlib.

N0NB:
Did you rebase and push?  The top of the page says this PR want to merge 9012 commits.  YIKES!

We're going to have to figure out how to get back to just your intended commits or close this an open a new PR.

mikaelnousiainen:
Wow, it really seems something went wrong with the latest changes (1558 changed files?) - not sure if starting with a fresh PR would fix things?

Anyway - I think this PR is trying to tackle a bit too much at once: both TCI 2.0 support _and_ an attempt to introduce a wire protocol for audio and I/Q streaming is a lot. And please don't get me wrong, I still appreciate all of this effort, but we may need to alter the approach a bit to end up with a solution we can maintain and develop further.

I see we could benefit from splitting this PR into at least 2 distinct parts:

1) TCI 2.0 rig control support _without_ any audio or I/Q streaming - this would be part 1

2) Once implementation and docs for #1940 are out, adapt the audio and I/Q streaming for the API. I can assist with implementation (if needed) and do additional testing (I have access to a SunSDR2 QRP here).

Looking at the audio + I/Q support in this PR, I think my proposal in #1940 will definitely cover all the features needed by a TCI 2.0 backend. I'm currently in the phase of cleaning up and documenting the initial implementation of the Hamlib data streaming subsystem and it looks like I can publish a PR some time next week.

I will also provide simple guides for implementing backend support for streaming and a separate guide documenting the network protocol (so it's not only living in a PR comment), since I see many apps would interface with Hamlib streaming directly on the network level (e.g. rigctld remote control).

How do you all feel about this kind of approach?

jfrancis42:
Sigh. I was trying to get this pushed while they were starting boarding for my flight, so I resorted to a python github tool a guy at work wrote and clearly chose the wrong options. I'll try again, I'm back from my trip and will do the good ol'-fashioned (correct) way this time.

I'm open to splitting the PRs into two, though the TCI driver without audio has no real value to the end user, though it would have value in terms of understanding the code in smaller chunks. And I worked really hard to keep it modular enough to swap out for a different transport.

I had not considered encryption. That's doable, but doing it right requires certs, and I can't imagine end-users are going to want to be generating self-signed (or any other kind) of certs as part of the setup. You could do pre-shared keys, but that's just asking for people to set them once and leave them for a lifetime once it works, and will add frustration to the setup (and be of limited security value if they're never rotated).

Again, I'm the new guy here, open to doing it "the hamlib way". I've got my stuff (SunSDR2 Pro, rtl-sdr, kiwisdr and soon-to-be RSPDuo) all working flawlessly now in my own code tree, I'm just trying to "share the wealth".

I'm good either way. Let me know what you guys prefer. I'll wait to fix and resubmit the PR until I hear which way you prefer. The PR (when done correctly) is really only two files: tci2.c and sidecar.c (plus the Makefile.am and some documentation files). It's actually a very understandable amount of code, my git screw-up just made it look far worse than it really is.

mikaelnousiainen:
@jfrancis42 Thanks for your comments and for being flexible to adapt to the on-going development of Hamlib. 🙇 If you take a look at the plan in #1940, it should be actually a superset of the protocol/features you've implemented - to go for a more generic approach that will suit a wider range of traditional ham radio transceivers and some more advanced SDRs. This is why it preserves information such as timing and details of any dropped data (which are important for the receiving application to know). Additionally, the wire format is generic and there is data format conversion built in so that applications don't need to care about the implementation details of transceivers and can simply receive PCM audio (or Opus in the future) or one of the supported I/Q sample formats (e.g. unsigned 8-bit or signed 16-bit which are common).

I think it may be easier to start with a new PR - and I'd still be in favor of doing the split as I suggested in my previous message: it also makes reviewing and merging easier, as the pieces are smaller. The state (step 1) where you have only the TCI 2 backend without streaming won't end up in any release anyway.

How do you feel about this approach @N0NB @GeoBaltz ? (suggestion copied below)

* PR 1: TCI 2.0 rig control support without any audio or I/Q streaming
* PR 2: Once implementation and docs for https://github.com/Hamlib/Hamlib/issues/1940 are out (next week), adapt the audio and I/Q streaming for the API. I can assist with implementation (if needed) and do additional testing. I have access to a SunSDR2 QRP here!

PR #2082, opened 2026-06-21
Make more data constant

GeoBaltz:
Change some more data to const, change more global symbols to locals.

Add rig_stop_voice_mem() to Yaesu rigs.

For master branch


N0NB:
Thanks, George.

PR #2083, opened 2026-06-25
Add TCI 2.0 (Expert Electronics) backend, CAT only

jfrancis42:
# TCI 2.0 (Expert Electronics / ExpertSDR3): CAT-only support

This PR adds a CAT backend for the **TCI 2.0** protocol (Transceiver
Control Interface), used by Expert Electronics' ExpertSDR3 and Apache
Labs' ANAN-with-TCI firmware.  It lives in a new top-level rig
backend, **`rigs/tci/`**, with the generic TCI 2.0 protocol code in
`tci2.c` and a per-radio capabilities table for the SunSDR2 PRO in
`sunsdr2-pro.c` — model **`SUNSDR2_PRO` (43001)**, manufacturer
"Expert Electronics".  Full bidirectional CAT — frequency, mode, PTT,
split, RIT/XIT, levels, functions, and CW macros.

## Scope

This is **CAT only**.  Audio and IQ are out of scope for this PR.
The TCI protocol does define both directions of audio and an RX IQ
stream, but adding that support cleanly involves more design choices
(external audio plumbing vs. in-process, platform-specific sink/source
handling, IQ resampling) than belong in a single PR.  A follow-up PR
will add audio/IQ support; the wire-level interaction with the TCI
server already in this PR will not need to change.

For now, binary WebSocket frames carrying audio or IQ are silently
consumed and discarded.  Users running a digital-mode app on the same
machine as the radio software can use the radio's own audio device or
their own audio plumbing (PulseAudio loopbacks, JACK, etc.) outside
of Hamlib.

## What's in this PR

A new `rigs/tci/` backend, separate from `rigs/dummy/`:

```
rigs/tci/
├── tci2.c          (~2650 lines: WebSocket + TCI 2.0 protocol library)
├── tci2.h          (public API for per-radio caps files)
├── sunsdr2-pro.c   (SunSDR2 PRO rig_caps, wires Hamlib callbacks to tci2_*)
├── tci.c           (DECLARE_INITRIG_BACKEND(tci) → registers sunsdr2_pro_caps)
├── Makefile.am
└── README-TCI-2.0.md
```

The `tci2.c` module is the shared protocol library — WebSocket
handshake, frame send/recv, message parser, push-driven state cache,
and all the get/set callbacks.  It exports every entry point through
`tci2.h` so future TCI-speaking rigs (SunSDR2 DX, ANAN with TCI,
MB1, etc.) can be added by dropping in another caps file alongside
`sunsdr2-pro.c` that points the same callbacks at a different
`rig_caps` table.

`sunsdr2-pro.c` is the only thing that's hardware-specific in this PR
— RF envelope, filter widths, TX power (1.8–54 MHz at 15 W), and the
model name.  Everything else is generic TCI 2.0.

Outside the new directory:

- `include/hamlib/riglist.h` — new backend family `RIG_TCI = 43` /
  `RIG_BACKEND_TCI = "tci"`, and `RIG_MODEL_SUNSDR2_PRO =
  RIG_MAKE_MODEL(RIG_TCI, 1) = 43001`.
- `src/register.c` — adds `DEFINE_INITRIG_BACKEND(tci)` and a
  `{ RIG_TCI, RIG_BACKEND_TCI, RIG_FUNCNAMA(tci) }` table entry.
- `configure.ac` — adds `rigs/tci` to `RIG_BACKEND_LIST` and
  `rigs/tci/Makefile` to `AC_CONFIG_FILES`.
- `bindings/python/test_Hamlib_class.py` — swaps `RIG_MODEL_TCI2` for
  `RIG_MODEL_SUNSDR2_PRO` in the allow-list.

No new dependencies and no new headers outside the new directory.

## Architecture

```
   ┌──────────────────────────┐    TCI 2.0 (WebSocket)    ┌──────────────────────┐
   │  ExpertSDR3 / SunSDR /   │◀─────────────────────────▶│  rigctld -m 43001    │
   │  ANAN-with-TCI server    │   single connection,      │  (this backend)      │
   │  (default :50001)        │   text + binary frames    │                      │
   └──────────────────────────┘                           └─────────┬────────────┘
                                                                    │ NET rigctl
                                                                    ▼ (default :4532)
                                                  JS8Call / WSJT-X / fldigi / ...
```

TCI is a **push-based** protocol over WebSocket.  On connect, the
server streams its full current state and ends with `READY;`.  After
that, any change — by us or by another client (e.g. the radio's own
GUI) — is broadcast to every connected client.

Because TCI's text responses aren't strictly correlated with requests
(an unsolicited `RX_SENSORS:` push can arrive between our
`VFO:0,0;` query and the server's `VFO:0,0,14078000;` answer), this
backend maintains a state cache in `struct tci2_priv` that's updated by
every incoming text frame via `tci2_process_message()`.
`tci2_recv_until()` drives that loop while waiting for a specific reply
prefix.  A handful of read-only values (S-meter, SWR, TX power meter,
CW keyer speed, VFO lock state) are server-push-only — these are
served from the cache rather than queried on demand.

The implementation is also written to behave well as a long-running
process: every `tci2_send()` first drains any stale echoes / telemetry
that have accumulated in the socket buffer, so the next round-trip
doesn't match the prefix of the *previous* command's response.  See
the gotchas section in `README-TCI-2.0.md` for the AGC-round-trip
symptom that drove this.

## Configuration parameters

Set with `-C key=value` on the rigctld command line:

| Token | Type | Default | Meaning |
|-------|------|---------|---------|
| `trx` | int 0..7 | 0 | Which TRX to control on multi-receiver hardware. |
| `txsource` | `default` / `mic` / `vac` | `default` | TCI audio-source field sent in `TRX:trx,true,<source>;`. |
| `digl_offset` | int 0..4000 Hz | 0 | Server-side DIGL mode frequency offset, sent on connect. |
| `digu_offset` | int 0..4000 Hz | 0 | Server-side DIGU mode frequency offset, sent on connect. |

`txsource` is a per-radio TCI-side choice (Mic input vs. the radio's
internal VAC).  No audio bytes touch this backend either way — the
field just tells the radio which input feeds its TX modulator when
keyed.

## Operations implemented

| Hamlib op | TCI command(s) | Notes |
|---|---|---|
| `set_freq` / `get_freq` | `DDS:` + `VFO:` | DDS first — see gotchas. |
| `set_mode` / `get_mode` | `MODULATION:` | Width is read-only from server's auto-applied filter. |
| `set_vfo` / `get_vfo` | client cache only | TCI has no "current VFO." |
| `set_ptt` / `get_ptt` | `TRX:` | `RIG_PTT_ON_MIC` → `Mic`, `RIG_PTT_ON_DATA` → TCI audio source. |
| `set_split_vfo` / `get_split_vfo` | `SPLIT_ENABLE:` | TX VFO is always B. |
| `set_split_freq` / `get_split_freq` | `VFO:trx,1,freq;` | |
| `set_rit` / `get_rit` | `RIT_OFFSET:` + `RIT_ENABLE:` | 0 turns off. |
| `set_xit` / `get_xit` | `XIT_OFFSET:` + `XIT_ENABLE:` | 0 turns off. |
| `set_level` / `get_level` | `VOLUME:` / `DRIVE:` / `SQL_*` / `AGC_MODE:` / `RX_NB_PARAM:` / `CW_KEYER_SPEED:` / cached sensors | `STRENGTH` / `SWR` / `RFPOWER_METER` from sensor streams. |
| `set_func` / `get_func` | `RX_NB_ENABLE:` / `RX_NR_ENABLE:` / `RX_ANF_ENABLE:` / `MUTE:` / `TUNE:` | `LOCK` is read-only (TCI has no client-side lock). |
| `send_morse` / `stop_morse` / `wait_morse` | `CW_MACROS:` / `CW_MACROS_STOP;` | Two-phase wait: TX engage then TX release. |
| `power2mW` / `mW2power` | client-side | SunSDR2 PRO TX cap: 15 W. |

## Testing

Tested against an Expert Electronics **SunSDR2 PRO** running
**ExpertSDR3**.  All 53 CAT round-trips passed, and the baseline state
was preserved exactly across the test:

- 5 × frequency (80m, 40m, 30m, 20m, 15m, 10m)
- 8 × mode (CW, USB, LSB, AM, FMN, RTTY, PKTUSB, PKTLSB)
- 4 × split on/off + readback
- 2 × split TX-freq
- 3 × RIT (+, −, 0)
- 3 × XIT (+, −, 0)
- 3 × AGC (OFF, FAST, SLOW)
- 2 × AF volume
- 2 × SQL
- 2 × KEYSPD
- 4 × NB / NR / ANF / MUTE on/off
- 2 × VFO A/B
- read-only sensors: STRENGTH, SWR, RFPOWER, RFPOWER_METER, KEYSPD
- `get_info` returning `"TCI 2.0 — SunSDR2PRO (trx 0)"`

After the `rigs/tci/` refactor the driver was re-exercised live
against the same SunSDR2 PRO and confirmed working: `rigctld -m 43001
-r <host>:50001 -t 4532` connects, `get_info` returns `"TCI 2.0 —
SunSDR2PRO (trx 0)"`, and freq/mode/level round-trips read clean
values (S-meter -37 dBm, AF 73%, AGC slow) at the test frequency.

PTT-keying was exercised via the CW-macro path during earlier
development; on-air RF was confirmed.

## Non-goals (deferred to future PRs)

- **Audio support** — bidirectional `RX_AUDIO` / `TX_AUDIO` per the TCI
  spec.  Architecture for this is being worked out separately; it
  affects how the backend handles binary WebSocket frames but not its
  CAT surface.
- **RX IQ support** — TCI's `IQ_STREAM` (unidirectional from radio to
  client).  Same comment: future PR, no impact on the CAT surface.
- **Multicast / multi-client coordination** — this backend treats
  itself as the sole client.  TCI does support multiple simultaneous
  clients, and the server broadcasts state changes to all of them; the
  state cache here will pick up changes made from the radio's GUI or
  another connected app, but there's no explicit master/slave
  coordination logic.
- **Other TCI radios.**  Only the SunSDR2 PRO is wired up.  Adding
  another TCI rig is a new caps file alongside `sunsdr2-pro.c` plus a
  `rig_register(&...)` line in `tci.c` — no protocol-level work needed.

## Files in this PR

```
rigs/tci/tci2.c                  new (TCI 2.0 protocol library)
rigs/tci/tci2.h                  new (public API for per-radio caps files)
rigs/tci/sunsdr2-pro.c           new (SunSDR2 PRO rig_caps)
rigs/tci/tci.c                   new (DECLARE_INITRIG_BACKEND, registers SunSDR2 PRO)
rigs/tci/Makefile.am             new
rigs/tci/README-TCI-2.0.md       new (backend documentation, moved from rigs/dummy/)
include/hamlib/riglist.h         +RIG_TCI / RIG_BACKEND_TCI / RIG_MODEL_SUNSDR2_PRO
src/register.c                   +DEFINE_INITRIG_BACKEND(tci) + table entry
configure.ac                     +rigs/tci in RIG_BACKEND_LIST, +rigs/tci/Makefile
rigs/dummy/Makefile.am           -tci2.c from DUMMYSRC
rigs/dummy/dummy.c               -rig_register(&tci2_caps);
rigs/dummy/dummy.h               -extern struct rig_caps tci2_caps;
bindings/python/test_Hamlib_class.py   RIG_MODEL_TCI2 → RIG_MODEL_SUNSDR2_PRO
```

## License

LGPL-2.1-or-later, same as the rest of Hamlib.


mikaelnousiainen:
Thanks a lot! I will do a final review during the weekend and will test it with my SunSDR2 QRP briefly.

The only thing change I'd vote for right now is that we'd add a new directory for TCI, e.g. rigs/tci for anything supporting the protocol - so it would be just moving those files and configuring automake accordingly. The exact models of rigs would be added in that directory as even the SunSDR2 models have differences that need to be specified in the rig caps (but those adjustments can be follow-ups to this PR later)

I think putting SDRs/protocols under the dummy directory has no advantages and simply causes confusion.

Any objections to this @N0NB @GeoBaltz @dforsi ?

jfrancis42:
Before I update the PR, do you anticipate a TCI directory containing multiple drivers, each optimized for a specific TCI device (sunsdr.c, anand.c, etc)? Or one single TCI driver with "device knowledge" inside of that driver? I think the former would be wiser, and I'm pretty sure that's what you said, but want to be sure before I make the change. Assuming that's correct, do you want sunsdr-pro.c, sunsdr-dx.c, sunsdr-qrp.c, etc? Or group all of the sunsdr functionality under a single sunsdr.c driver? Similarly, there will almost certainly be differences between TCI 1.x and TCI 2.x, as well as ExperSDR2 vs ExpertSDR3? I'm assuming those differences would be rolled into a single driver, but I'm just thinking it's easier to make those choices now, before committing to code, rather than trying to retrofit these things later.

N0NB:
I can't speak for Mikael, but I think separate files is more maintainable.

mikaelnousiainen:
@jfrancis42 I think we can have much of the "common TCI driver code" (the devices do share quite a bit of functionality if I've not understood things entirely wrong?) as separate tci2.c and tci2.h (or tci1.c and tci1.h for the 1.x protocl if it's still used?) - and then each rig can have its own .c/.h file with the capabilities and registration + any device-specific functions. This is pretty much how the `icom` backends have been built.

-> I agree with Nate and your example in your comment.

jfrancis42:
Ok, PR has been updated with the new file/directory structure. Builds clean and tests clean against my rig.

include/hamlib/riglist.h
```diff
@@ -110,6 +110,26 @@
...
 ///@}
 
+/*
+ * TCI (Transceiver Control Interface) — Expert Electronics ExpertSDR3
+ * and other TCI 2.0-speaking radios.  Each model is one physical rig
+ * that talks TCI 2.0 over WebSocket; the generic protocol library lives
+ * in rigs/tci/tci2.c and is shared by every rig in this family.
+ */
+/** The `TCI` family. */
+#define RIG_TCI 43
```
mikaelnousiainen: Really small thing, but could we move this new block of code for TCI to the end of the file, so that the "families" are in order (43 is last)?

rigs/tci/sunsdr2-pro.c
```diff
@@ -0,0 +1,120 @@
...
+    .has_get_parm   = RIG_PARM_NONE,
+    .has_set_parm   = RIG_PARM_NONE,
+
+    /* SunSDR2 PRO: HF + 6 m RX, HF + 6 m TX at 15 W out.
+     * Datasheet RX range is 0.09–65 MHz; TX is the amateur bands within
+     * 1.8–54 MHz at the rig's licensed limits.  We publish a generous
+     * envelope and let the server clamp it. */
+    .rx_range_list1 =
+    {
+        { kHz(1), MHz(65), TCI2_MODES, -1, -1, TCI2_VFOS, RIG_ANT_NONE },
```
mikaelnousiainen, line 56: Why is output power range set to -1? 🤔

mikaelnousiainen (commented):
This is looking good - I did a first quick pass with just minor comments. I'm about to test this on a SunSDR2 QRP now briefly.

mikaelnousiainen:
@jfrancis42 This is rather odd. I tested the SunSDR2 QRP with latest ExpertSDR3 version 1.1.7 - and updated it to firmware version 11.0. It works fine with ExpertSDR3 and indicates it's listening in port 50001 in my LAN.

However, when trying to make Hamlib connect to that port, it never answers. I don't really have much experience on the TCI protocol and the SunSDR2 connectivity, not sure if I'm doing something wrong...

mikaelnousiainen:
Ok, I think I got it working now. I just realized an important point: one needs to have ExpertSDR3 software running to have the TCI server available - and that the radio protocol is **completely undocumented** (probably uses UDP, there seems to be some reverse-engineered info somewhere, but probably not worth looking into). Why did they ever make it this way...= Anyway, I'm able to proceed with testing.

I guess there is some way to run ExpertSDR3 in a headless mode in a "server-like" way?

> @jfrancis42 This is rather odd. I tested the SunSDR2 QRP with latest ExpertSDR3 version 1.1.7 - and updated it to firmware version 11.0. It works fine with ExpertSDR3 and indicates it's listening in port 50001 in my LAN.
> 
> However, when trying to make Hamlib connect to that port, it never answers. I don't really have much experience on the TCI protocol and the SunSDR2 connectivity, not sure if I'm doing something wrong...

mikaelnousiainen:
@jfrancis42 I noticed one issue that's rather important to fix: The backend communication seems to get stuck if the connection breaks. For every command run - if the connection fails (or has failed) - there should be a configurable number of retries with a timeout. There are already _global_  conf settings called `timeout` (in milliseconds) and `retry` that could be used here, so no need to add new ones.

jfrancis42:
@mikaelnousiainen Thanks for testing. Here's an update addressing all of the feedback:

## 1. riglist.h ordering

Done — moved the `RIG_TCI 43` block to the end of the file (after HARRIS 42) so families appear in numerical order. Sorry, got a bit carried away and just stuck it right next to Dummy where it was oringinally without thinking about order.

## 2. rx_range_list1 power fields (-1)

If I'm reading the other backends correctly, `-1` is the standard Hamlib convention for `rx_range_list` entries — it means "not applicable, this row describes reception." The struct definition in `rig.h` documents it as:

```c
int low_power;   /*!< Lower RF power in mW, -1 for no power (ie. rx list) */
int high_power;  /*!< Higher RF power in mW, -1 for no power (ie. rx list) */
```

The actual TX power limits are specified in `tx_range_list1` as `mW(1)` to `W(15)`. The IC-7300, IC-705, FLRig, and other backends all use the same `-1, -1` pattern in their RX range tables. I just followed the convention that I saw in other rigs. If this isn't right, I'm happy to change it.

## 3. Connection-break handling (retry/timeout)

Implemented. Added two new functions:

- **`tci2_reconnect()`** — on connection failure, closes the dead socket via `port_close()`, reopens TCP via `port_open()`, performs a fresh WebSocket handshake, drains until READY to re-sync state, rebuilds the runtime mode table, and re-subscribes to RX/TX sensor pushes.

- **`tci2_transaction()`** — wraps every command (query or fire-and-forget) in a retry loop. On I/O errors or recv timeouts, calls `tci2_reconnect()` and retries. Uses the existing `RIGPORT(rig)->retry` count (set to 3 from rig_caps) as the maximum number of attempts.

All `tci2_set_*` / `tci2_get_*` functions now go through `tci2_transaction()` instead of raw `tci2_send()` + `tci2_recv_until()`. Timeout enforcement comes from `read_block()` which uses `select()` with `RIGPORT(rig)->timeout` (2000 ms), so a half-open TCP connection times out after 2 seconds rather than hanging.

## 4. ExpertSDR3 headless mode

TL;DR ExpertSDR3 has to be running with a GUI somewhere on the network, and TCI clients connect to that instance.

Unfortunately ExpertSDR3 doesn't offer a headless mode. The TCI WebSocket server is embedded inside the GUI application, so ExpertSDR3 must be running with a graphical session for TCI to be available. Expert Electronics offers ExpertStarter for headless radio-site deployment, but it uses a proprietary protocol (ERS) and does not expose TCI. I've never tried it, personally.

jfrancis42:
Where are we on the new integrated audio spec? I'm looking forward to getting to work on rolling that in. If you have a preview version, I'm happy to work from that, then update (and provide feedback) as it gets completed.

mikaelnousiainen:
> Where are we on the new integrated audio spec? I'm looking forward to getting to work on rolling that in. If you have a preview version, I'm happy to work from that, then update (and provide feedback) as it gets completed.

I'm doing a final review to publish it in a branch on my Hamlib fork - we can first test integration for TCI 2.0 backend on this "preview version" and then move it on to a proper PR in Hamlib. So it's a matter of days!

I've only done some preliminary testing on Icom (network protocol) and FlexRadio network protocol with it, but neither of those are release-ready and need some more work, so more testing and feedback is all welcome!

rigs/tci/sunsdr2-pro.c
```diff
@@ -0,0 +1,120 @@
...
+
+#include "hamlib/config.h"
+
+#include "hamlib/rig.h"
+#include "tci2.h"
+
+struct rig_caps sunsdr2_pro_caps =
+{
+    RIG_MODEL(RIG_MODEL_SUNSDR2_PRO),
+    .model_name     = "SunSDR2 PRO",
```
mikaelnousiainen: One small change could help here: should we add the protocol here, since we may have a chance later to add support for fully native support (without ExpertSDR software)?

The model_name could be `SunSDR2 PRO (TCI)` - same for model: `RIG_MODEL_SUNSDR2_PRO_TCI` ?

include/hamlib/riglist.h
```diff
@@ -1158,6 +1158,26 @@
...
+/** Used in register.c for the `be_name`. */
+#define RIG_BACKEND_TCI "tci"
+
+/**
+ * \name TCI
+ * TCI 2.0 models.
+ */
+///@{
+/// Model of the `RIG_TCI` backend family.
+#define RIG_MODEL_SUNSDR2_PRO RIG_MAKE_MODEL(RIG_TCI, 1)
```
mikaelnousiainen: One small change could help here: should we add the protocol here, since we may have a chance later to add support for fully native support (without ExpertSDR software)?

The model could be: `RIG_MODEL_SUNSDR2_PRO_TCI` ?

mikaelnousiainen:
Thanks, the changes look good now - I just added a request for a small change to consider adding "TCI" to the rig model constant and the model name string, so it is clear this backend uses TCI and not the native protocol. It looks like the protocol has been reverse-engineered so we may have a chance to add it later. After this change I'm fine merging this for more testing and further development.

Regarding the timeout and reconnect, I would've liked to test it once more, but I don't have the rig with me right now (travelling) - but I trust you tested the scenario where connection gets cut / times out. Hamlib should be able to fully recover from control connection being broken (once it's established).

cc @N0NB

jfrancis42:
Agreed, there are at least two non-ExpertSDR* projects on the INTARWEBZ that we may want to support in the future. PR incoming as soon as I finish the full rebuild from scratch to make sure I didn't blow something up...

jfrancis42:
Guys, my bad. I've been sitting here waiting for an update to this PR and in checking on it realized I had left out one of the requested changes. I just pushed it.

mikaelnousiainen:
Thanks, I think this looks good for merging for further testing, but I'll let @N0NB do a final decision on merge. It's also the holiday season, so people (including me) may not be super responsive all the time.

The audio streaming API will be out very soon - I will share the details immediately when that happens, also directly with you @jfrancis42. I just haven't had the chance to work on it in the past week that much... I did some final reviews using AI tooling as an assistant (just to make sure there's nothing obvious missing that even agents notice :)), so I'll just do one more pass myself and I should be fine to publish it first in my own fork of Hamlib, then do a PR later.

rigs/tci/tci2.c
```diff
@@ -0,0 +1,2737 @@
...
+
+#include "hamlib/config.h"
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <math.h>
+#include <ctype.h>
+#include <time.h>
+#include <unistd.h>
+#include <poll.h>
```
dchristle: Will this limit Windows compatibility e.g. with mingw?

rigs/tci/tci2.c
```diff
@@ -0,0 +1,2737 @@
...
+#define TCI2_WS_BUFLEN     32768  /* Sized to absorb any binary IQ/audio
+                                   * frame the server pushes -- they're
+                                   * discarded but we still have to read
+                                   * the bytes off the socket. */
+#define TCI2_CMDLEN        512
+#define TCI2_INIT_MAX_MSGS 256   /* max frames to drain during open */
+#define TCI2_RECV_MAX      256   /* max frames to scan for a reply */
+#define TCI2_MAX_RT_MODES  64    /* runtime mode table slots */
+
+#define TCI2_MAX_TX_MW 100000u  /* assume 100 W max output */
```
dchristle: sunsdr2-pro.c has a 15W limit, so would full 15 W read back as 0.15 on the meter, and would `mW2power(15000)` thus give 0.15 instead of 1.0?

rigs/tci/tci2.c
N0NB: Good catch, David.

Here is the output from MinGW:

``` console
Making install in rigs/tci
  CC       tci2.lo
  CC       sunsdr2-pro.lo
  CC       tci.lo
tci2.c:48:10: fatal error: poll.h: No such file or directory
   48 | #include <poll.h>
      |          ^~~~~~~~
compilation terminated.
make[1]: *** [Makefile:499: tci2.lo] Error 1
make[1]: *** Waiting for unfinished jobs....
make: *** [Makefile:671: install-recursive] Error 1
```

rigs/tci/tci2.c
```diff
@@ -0,0 +1,2737 @@
...
+    return write_block(RIGPORT(rig), frame, flen);
+}
+
+/*
+ * Receive one WebSocket frame from the server.
+ * Returns payload length on success (buf is NUL-terminated text),
+ * 0 for a binary frame (audio/IQ — drained and discarded; buf is empty),
+ * or a negative Hamlib error code.
+ */
+static int ws_recv_frame(RIG *rig, char *buf, int buflen)
```
dchristle: Part 5.5.2 of the WebSocket RFC ([link](https://datatracker.ietf.org/doc/html/rfc6455#section-5.5.2)) specifies that the Ping op-code 0x09 *must* be handled with a Pong reply. Right now, `ws_recv_frame` only handles opcodes 0x02 and 0x08.

Another required part of the spec is in how to handle fragmentation & re-assembly, defined in section 5.4 ([link](https://datatracker.ietf.org/doc/html/rfc6455#section-5.4)).

`ws_recv_frame` looks like it ignores the `FIN` bit. Fragmented messages are mentioned in the spec & it indicates they "MUST" be re-assembled (FIN + opcode 0x00), and that control frames "MUST" be handled in the middle of a fragmented message.

One more spec issue is that it'll return (for op-code 0x08) a `-RIG_EPROTO` on a Close frame without responding. But 5.5.1 of the spec ([link](https://datatracker.ietf.org/doc/html/rfc6455#section-5.5.1)) says,
> _If an endpoint receives a Close frame and did not previously send a Close frame, the endpoint MUST send a Close frame in response._

So, it should consume the Close payload + send the Close frame. And then also separately we should emit a Close frame during normal shutdown.

jfrancis42:
@mikaelnousiainen @dchristle @N0NB — thanks for the reviews and for reproducing the MinGW failure. Two commits pushed that address all of the feedback.

First, the python bindings test. That's what was breaking the Ubuntu distcheck run. When I renamed the model constant to `RIG_MODEL_SUNSDR2_PRO_TCI` I forgot to regenerate `test_Hamlib_class.py`, so its expected-callables list still had the old name and the `test_callables` assertion failed. Fixed in the first commit.

Second, the MinGW build. Full disclosure: I'm not a Windows guy and I have no way to build MinGW myself, so this is my best shot after some googling and reading how the rest of Hamlib handles the same problem. The `<poll.h>` include was there for a single non-blocking peek in `tci2_drain()`. I swapped that for `select()` with a zero timeout, and used the same include pattern I saw in `src/network.c` (`sys/socket.h`, `sys/select.h`, `sys/time.h` on POSIX; `winsock2.h` on Windows, each behind its `HAVE_*` guard). Compiles clean on Linux; if it doesn't quite land on MinGW, let me know what breaks and I'll iterate.

Third, the power-scale bug @dchristle spotted. You were right — the hardcoded 100 W scale meant `mW2power(15000)` returned 0.15 instead of 1.0 on a 15 W rig. I moved the max into `struct tci2_priv` as `tx_max_mw` and initialize it in `tci2_init` by scanning `caps->tx_range_list1` for the largest `high_power`. Both conversion routines and the `RFPOWER_METER` read now use that. When the SunSDR2 DX or QRP caps land later they'll pick up their own values automatically. Verified against the real rig: `mW2power(15000)` = 1.0, `power2mW(1.0)` = 15000, and the same for the 0.5 and 0.1 points.

Fourth, the WebSocket RFC 6455 issues @dchristle flagged — Ping, fragmentation, and the Close handshake. Split `ws_recv_frame` into a low-level `ws_read_one_frame` that returns one raw frame with its opcode + FIN bit, and a message-assembly loop on top that handles fragmentation (FIN + continuation opcode 0x0), auto-Pongs Pings inline so they don't disturb any reassembly in progress, and echoes a Close back before returning EIO. `tci2_close` now sends its own Close (status 1000) before dropping the TCP. Also split `ws_send_text` into a generic `ws_send_frame(opcode, payload, len)` plus text/close/pong wrappers.

Tested end-to-end against the SunSDR2 PRO running ExpertSDR3: freq/mode/split/RIT/XIT/AGC/AF/KEYSPD/NB/NR/ANF/PTT all round-trip through the rewritten frame layer with no errors or warnings at verbose level, and `ss -tan` shows no stuck server-side sockets after shutdown so ExpertSDR3 is happy with the new Close frame. Rig state restored exactly at the end.

N0NB:
To all involve, I presume that this new model will be 5.0+ only and not back ported to 4.7?  If I understand correctly, 4.7 will lack the necessary support for this model.

mikaelnousiainen:
@jfrancis42 The first draft PR of the Hamlib streaming subsystem is here: https://github.com/Hamlib/Hamlib/pull/2116 - it's likely it will need adjustments (Windows support, GitHub actions checks, ...) but it should be safe to develop a backend against the API already - I don't expect any huge/breaking changes. You can also use the original PR branch to pull the code from here: https://github.com/mikaelnousiainen/Hamlib/tree/streaming-subsystem-pr - I hope to keep the PR open for some weeks maybe to stabilize things and to get further feedback - especially from you and the TCI2 backend development. During that time, I will publish the Icom network protocol backend that uses the streaming API. Please provide your feedback in the PR #2116 where it can be addressed!

EDIT: And let's keep the original plan to merge this CAT-only TCI2 PR first and have a separate PR for the TCI2 streaming implementation!

jfrancis42:
Just FYI, I pulled your 2116 PR into a local copy of hamlib with my 2083 PR and now have it streaming audio over your framework. It required some changes to the tci2 code, but that was expected, since the TCI 2.0 spec mandates a shared socket for control and audio. But it does work. I also had to write a "virtual audio appliance" to actually listen to the audio and provide a virtual soundcard device that other software can use directly (exactly the same way my earlier audio sidecar worked - that code lives outside of hamlib, just like my earlier sidecar did). The good news is that everything works with no changes so far to any PR 2116 code. There's a lot of reading, coding, and testing still to be done, but it's a very good sign that it's going this well this early. I'm keeping the integration code completely separate from PR 2083, I just wanted to report in on the first pass at integrating the two. I woke up at 4am and couldn't sleep and saw your email that the standard was done ("-ish"), so I couldn't resist trying it out. I should probably switch gears to my day job at this point.

N0NB:
@jfrancis42 Now that I merged PR #2116, I see there are three merge conflicts.  It's probably easier for you to fix since you know the additions you need.

N0NB:
Thanks, Jeff.

Looks like I resolved the conflicts correctly and have clean builds.

PR #2085, opened 2026-07-01
Clutter reduction

GeoBaltz:
Reduce the scope or range of many variables

- Scope of iteration variables limited to the controlled loop
- Values of switch variables limited to true/false

General cleanup, removes distractions while reading code.

And then there is the fallout of trying to use a couple of simulators to observe the workings of the code - another session of trying to bring the Real World into simulators/*.c.

For master branch only


N0NB:
Thanks, George.

Issue #2084, opened 2026-06-30
TS790 error in commands for split mode


Mike-hub40638:
When using hamlib with WSJTX in full split operation, there are commands sent to the rig that are not in the Kenwood TS790 CAT manual.

By sniffing the com port it looks like commands FR0 and FT1 are sent. The rig replies with ?
And then when in TX, rig remains on VFOA.

The correct command is SP1;
This puts the rig into split mode.
Hamlib appears to not send this command.

Enviro:
Windows 10
WSJTX v3.1.0 260418
Hamlib 4.7.1 2026-04-15

(I am assuming this is hamlib issue and not WSJTX?)

GeoBaltz:
Definitely a hamlib issue.  I found where the FR0 and FT1 are coming from, but the routine that would send SP0/SP1 looks like it is never called.  Do you have any trace data, so I can see the actual call sequence?

All of those problem may be fixable.  However, I have no way to test, and I have no Windows build capability.  This may  take a while.

73 n3gb

N0NB:
Once the fix is applied, Mike should be able to drop in the DLL from the daily 4.7.3 snapshot to test with WSJT-X.  When all of that is complete I will post a link here.

GeoBaltz:
Thanks, Nate.  I will gen the simple fix right away.  It shouldn't affect anything but the TS-790, which is broken anyway.

N0NB:
Mike, the updated DLL is at: https://n0nb.fbx.one/share/hamlib/libhamlib-4.dll (this is from my home Internet and upload speed is 2.5 Mbps, so will be slow).

Someplace in the WSJST-X directory structure you should find a like named DLL.  with WSJT-X closed, you can copy this one over the old one (perhaps copy the old one to another place to restore, if needed) and restart WSJT-X.  You should be able to verify the version in the radio setup window.  It should show 4.7.3~rc.

Let us know how it goes.  I'll be away most of the day until late evening.

Mike-hub40638:
Hi, progress thanks.

So split command is now working ok. 
When I press the 2m band button for the first time it sends SP1 then SP0 then SP1. So we get where we need to be. (not an issue)

One issue: VFOB is in FM mode. It looks like USB mode is not being written to VFOB. (freq is ok)
From the manual it looks like you have to activate VFOB and then write the mode:

FN1;
MD2;
FN0;


Lastly, FR0 and FR1 are still being sent, but ignored by the rig.

[...]

PR #2086, opened 2026-07-02
Make TS-790 match TS-50 & TS-940

GeoBaltz:
These rigs are of the same vintage, and the command set is very similar.

This is the quick fix - if it doesn't work, the real coding fix is a lot hairier.


N0NB:
DLL updated with the cache fix, Mike.

GeoBaltz:
Closes issue #2084

N0NB:
Thanks, George and Mike.

PR #2087, opened 2026-07-04
Rewrite AT-D578UVIII backend with dual-mode serial protocol support

CowboyPilot:
## Summary

Rewrites the AnyTone AT-D578UVIII backend to support both serial protocols exposed by the radio on its 115200 8N1 port, selectable at runtime via a `-C commode=` config flag.

- **Default mode (`commode=0`)** — Direct serial mic protocol. Sends raw byte commands (0x06 keepalive, 0x41 PTT) with no initialization handshake and no radio display lockout. Ideal for digital mode software like VARA and fldigi that only need PTT control — the operator retains full use of the radio's front panel and the connection is invisible to the user.

- **COM mode (`commode=1`)** — BT-01 ADATA protocol. Performs a COM MODE handshake on open and uses `+ADATA:00,NNN\r\n` framed commands for PTT, get/set frequency, get/set VFO, and set clock. The radio displays "EXTERNAL CABLE MODE" while connected, locking the front panel. Suited for automated station control where frequency and VFO selection need to be driven programmatically.

The single rig model (37001) covers both modes — no changes to `riglist.h` or model registration required.

## Roadmap: raw command passthrough

The keepalive thread checks `transaction_active` before sending, so `rig_send_raw` / rigctld `w` commands pass through without collision. This is intentional — the radio's firmware accepts 34 distinct ADATA command bytes, most of which are undocumented. By exposing the driver's COM MODE session (handshake, keepalive, serial port management) to third-party software via the existing `w` interface, client applications can send arbitrary ADATA commands to explore and use protocol features that the backend doesn't yet implement natively.

This opens the door for a standalone BT-01 client program that connects through `rigctld` rather than managing the serial port directly — it gets session management for free and can coexist with other Hamlib clients sharing the same `rigctld` instance. As commands are identified and stabilized through this passthrough, they can be promoted to proper Hamlib API functions in the backend.

## Changes

All modifications are confined to `rigs/anytone/`:

- **anytone.c** — Keepalive thread branches on `commode`; `set_ptt` branches between raw 0x41 and ADATA 0x56; ADATA-only functions (`get_freq`, `set_freq`, `get_vfo`, `set_vfo`, `set_clock`) return `-RIG_ENAVAIL` when `commode=0`; keepalive yields to `transaction_active` for raw passthrough safety
- **anytone.h** — Added `commode` flag to priv struct, `TOK_COMMODE` config token, clock function declarations
- **d578.c** — `anytone_cfg_params[]` with `RIG_CONF_CHECKBUTTON` for commode; rig caps wired to all function pointers

## Test plan

- [ ] `rigctl -m 37001 -s 115200 -r /dev/ttyUSB0` — PTT on/off with no display lockout
- [ ] `rigctl -m 37001 -C commode=1 -s 115200 -r /dev/ttyUSB0` — PTT, `F`/`f` freq, `V`/`v` VFO, set_clock
- [ ] Verify `commode=0` functions return `-RIG_ENAVAIL` for freq/vfo/clock
- [ ] Verify raw `w` commands through `rigctld` in COM mode don't collide with keepalive

Protocol analysis based on [jrobertfisher/AT-D578UV-software-mic](https://github.com/jrobertfisher/AT-D578UV-software-mic) and firmware reverse engineering of the D578UV v1.21 and BT-01 v1.02 firmware images.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,25 @@ The radio exposes two distinct serial protocols on the same 115200 8N1 port:
 
 Protocol analysis based on [jrobertfisher/AT-D578UV-software-mic](https://github.com/jrobertfisher/AT-D578UV-software-mic) and firmware reverse engineering of the D578UV v1.21 and BT-01 v1.02 firmware images.
 
+## Raw command passthrough
+
+When `rigctld` is running in COM mode (`commode=1`), third-party software can send arbitrary ADATA commands to the radio using Hamlib's `w` (send_cmd) interface — even for functions the driver doesn't explicitly support. The driver handles COM MODE session management and keepalive automatically; the keepalive thread yields to raw commands so they won't collide.
+
+Example from `rigctl`:
+
+```bash
+# Send raw ADATA command 0x57 (1-byte payload) and read 22-byte ACK
+w +ADATA:00,001\r\nW\r\n 22
+```
+
+From `rigctld` over TCP (port 4532 by default):
+
+```
+w +ADATA:00,001\r\nW\r\n 22
+```
+
+This lets any ADATA-aware software leverage the driver's session without needing to manage the serial port, COM MODE handshake, or keepalive directly. The radio's firmware supports 34 ADATA command bytes — see the protocol map spreadsheet for the full list.
+
 ---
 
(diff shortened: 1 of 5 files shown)


CowboyPilot:
Closing in favor of a cleaned-up PR from a squashed branch.

PR #2096, opened 2026-07-10
Update LLM comments in README.developer

N0NB:
Other minor updates.

### DIFF
--- a/README.developer
+++ b/README.developer
@@ -544,14 +544,26 @@ GitHub.
 
 The Hamlib project accepts contributions on the "honor system".  It is implied
 that contributions are the original works of their respective authors and
-offered under the license terms of the GPL 2.0 or LGPL 2.1.  Use of LLM
-generated code breaks this model.  The contributors of such code cannot claim it
-as their own original work.
+offered under the license terms of the GPL 2.0 or LGPL 2.1.  The Project's
+original stance was that use of LLM generated code breaks this model and that
+contributors of such code cannot claim it as their own original work  However,
+recent contributions have clearly disclosed the use of LLM's such as Claude or
+security LLMs to assist with development of code contributed to Hamlib and this
+code has been reviewed, accepted, and committed to the repository.
+
+Consensus from discussions by Hamlib contributors have leaned toward regarding
+LLMs as a tool much like other tools used to aid development.  That said, LLMs
+have the appearance of being able to offer complete code contributions but this
+is often not the case.  Such code needs to be carefully examined and tested.
+All code contributions are reviewed by the core Hamlib team and many other
+contributors.  We strongly recommend that the use of LLM assistance be
+disclosed in commit messages.
 
 Finally, Hamlib is a hobby project for hobbyists and as such all contributors
 should take pride in having their original work included for the benefit of many
-other radio hobbyists.  Against that backdrop, using LLM generated code seems
-like cheating.
+other radio hobbyists.  Against that backdrop, using LLM generated code needs to
+be carefully considered, however, the project understands that these are tools that
+are available and will be used to assist in improving Hamlib.
 
 
 2. How to add a new backend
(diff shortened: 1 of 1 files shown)


N0NB:
Comments on the updated language regarding LLMs are welcome and encouraged.

KJ5HST:
Much better.

README.developer
```diff
@@ -544,14 +544,26 @@ GitHub.
 
 The Hamlib project accepts contributions on the "honor system".  It is implied
 that contributions are the original works of their respective authors and
-offered under the license terms of the GPL 2.0 or LGPL 2.1.  Use of LLM
-generated code breaks this model.  The contributors of such code cannot claim it
-as their own original work.
+offered under the license terms of the GPL 2.0 or LGPL 2.1.  The Project's
+original stance was that use of LLM generated code breaks this model and that
+contributors of such code cannot claim it as their own original work  However,
```
dg0yt:
```suggestion
contributors of such code cannot claim it as their own original work.  However,
```

README.developer
```diff
@@ -544,14 +544,26 @@ GitHub.
...
+recent contributions have clearly disclosed the use of LLM's such as Claude or
```
dg0yt:
```suggestion
recent contributions have clearly disclosed the use of LLMs such as Claude or
```

README.developer
```diff
@@ -544,14 +544,26 @@ GitHub.
...
+security LLMs to assist with development of code contributed to Hamlib and this
+code has been reviewed, accepted, and committed to the repository.
+
+Consensus from discussions by Hamlib contributors have leaned toward regarding
+LLMs as a tool much like other tools used to aid development.  That said, LLMs
+have the appearance of being able to offer complete code contributions but this
+is often not the case.  Such code needs to be carefully examined and tested.
+All code contributions are reviewed by the core Hamlib team and many other
+contributors.  We strongly recommend that the use of LLM assistance be
```
mikaelnousiainen, line 559: Instead of "recommend", should we say "require"? I think it's very much in the spirit of open-source development to be completely transparent about the origins (attribution) and techniques/tools used to produce a particular contribution.

mikaelnousiainen (commented):
Thank you! 🙇 This is a very welcome change.

Reviewing and understanding contributions (before merging any code) is still as important as it has always been - this part of the process does not disappear and still needs the Hamlib community/maintainers as "filters". At least for now, I see this kind of model as being a good way forward.

I added one small suggestion to *require* disclosing use of LLMs. I think it's the right way forward in any open-source project.

README.developer
N0NB: I agree completely, Mikael.

I went back and forth on that a few times between "strongly recommend", "strongly request", and "require".  The first two are similar though the second isn't a common idiom in American English, which, like it or not, is the primary dialect of the project.  That leaves "require" which leads to asking, "What are the consequences of noncompliance?"

The answers I come up with aren't very good.  Historically, there has been a high level of respect between the project and its contributors.  I certainly am not in favor of destroying that respect so I came back to "strongly recommended" as the best wording of the three choices.

KJ5HST:
@mikaelnousiainen, It cannot be "required" if it cannot be enforced. It's the honor system after all.

README.developer
N0NB: I thought I'd replied to your suggestion and others had as well, but I don't see those comments.

I think that trying to require LLM disclosure is doomed especially since this project does not require copyright assignment.  As it is, all authors retain their copyright over their contributions that are submitted under the terms of the GPL/LGPL.  As such I think the best we can do is strongly recommend disclosure.

This text is always open to amendment in the future if the need arises.

N0NB: I reopened the PR and there are the previous comments.  I've only lost half of my mind!  🙄

mikaelnousiainen: 👍 Agreed, last time I just left "thumbs up" reactions here in GH so it may be those did not reach you.

So feel free to merge this, don't see much activity here requiring any other chnages.

I'm assuming these went to the mailing list too so everyone following the project had a chance to see this.

N0NB:
Changes pushed.

PR #2098, opened 2026-07-20
IC-7610, IC-7760: Implement SYNC

df7cb:
\set_func SYNC is called "Tracking" in the IC-7610 main menu.

Tested with my IC-7610 and rigctld here.


GeoBaltz:
It is also available in the IC-7760, so probably should be enabled there.

df7cb:
> It is also available in the IC-7760, so probably should be enabled there.

Added, thanks! (The IC7760_FUNCS is much shorter than the 7610's, but I don't think we should fix that as part of this PR.)

N0NB:
Thanks, Christoph.  Also cherry-picked into 4.7.

PR #2105, opened 2026-07-23
fix: remove debug history race

dchristle:
I found this while running WSJT-X under ThreadSanitizer. Hamlib's debug-history code used one shared temporary buffer, so two threads calling `rig_debug()` could write to it at the same time.

The fix is fairly small: each call now formats into its own buffer, and the shared history is updated under the existing lock. Normal messages use a small stack buffer; longer ones allocate only what they need. The existing rolling-history behavior is unchanged.

Tests cover several threads logging at once, along with the normal trimming and ordering of the rolling history.

This is intentionally limited to the race TSan found in `rig_debug()`; it does not try to redesign Hamlib's older public debug-history API.

I ran the Hamlib test suite and the focused test under TSan and ASan/UBSan on macOS.

### DIFF
--- a/include/hamlib/rig.h
+++ b/include/hamlib/rig.h
@@ -3167,9 +3167,19 @@ rig_set_debug_time_stamp(int flag);
 extern HAMLIB_EXPORT(int)
 rig_need_debug(enum rig_debug_level_e debug_level);
 
+#if defined(__MINGW_PRINTF_FORMAT)
+#define HAMLIB_ATTRIBUTE_FORMAT_PRINTF \
+    __attribute__((__format__(__MINGW_PRINTF_FORMAT, 2, 3)))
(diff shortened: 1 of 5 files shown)


N0NB:
Thanks, David.

It applied clean to `master` but in the `Hamlib-4.7` branch I get a merge conflict in `include/hamlib/rig.h`.  Feel free to open a new PR with this code ported into the Stable branch.  I think it would be a good idea to include this in the forthcoming 4.7.3.

dchristle:
Thanks Nate - I'll look into a backport & open a PR for that. Thanks for reviewing & merging this so quickly.

dchristle:
@N0NB The backport PR is here: https://github.com/Hamlib/Hamlib/pull/2107

PR #2106, opened 2026-07-24
fix: prevent network AGC parser overflow

dchristle:
I found that the NET rigctl and Quisk backends could write past their fixed AGC arrays while parsing `agc_levels` from a peer. The parser reused an index from earlier capability processing and did not enforce the eight-entry limit.

The fix uses a shared bounded parser for both backends. An over-length list is truncated to the eight-entry array capacity — the first eight valid levels are kept and any extras are ignored. This matches how the neighboring `ctcss_list`/`dcs_list` parsing caps at its array size. This choice keeps a client forward-compatible: if a future rigctld advertises a ninth AGC level, older clients still get the eight they support instead of losing AGC control entirely. A malformed field is logged and ignored as a whole, and the connection is not failed when one is encountered.

Tests cover short and maximum-sized lists, a malformed field, and a nine-entry list that is truncated to eight.

I ran the tests under ASan/UBSan on macOS. The full suite ran with the new test passing. There is an unrelated existing `test rig.sh` ASan failure in `get_dummy_level()`, but I left that untouched here & put out a fix for it in #2108

### DIFF
--- a/rigs/dummy/dummy_common.c
+++ b/rigs/dummy/dummy_common.c
@@ -20,10 +20,12 @@
  *
  */
 
-// cppcheck-suppress *
+#include <errno.h>
+#include <limits.h>
 #include <stdlib.h>
+#include <string.h>
 
-#include "hamlib/rig.h"
+#include "dummy_common.h"
 
 struct ext_list *alloc_init_ext(const struct confparams *cfp)
 {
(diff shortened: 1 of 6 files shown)


N0NB:
It looks like pushing the debug race fix has caused a conflict in this PR.

dchristle:
@N0NB Yep. I rebased & pushed, so it should be mergable now.

N0NB:
Thanks, David.

Like the previous PR, this one won't merge cleanly into 4.7 so please create another PR for this fix against `Hamlib-4.7` branch.

PR #2107, opened 2026-07-25
fix: remove debug history race in 4.7

dchristle:
This backports #2105 to `Hamlib-4.7` for the forthcoming 4.7.3 release.

The only porting adjustment preserves the branch's existing `HAMLIB_PARAMS` declaration style while retaining printf-format checking.

The focused test passes under ASan/UBSan on macOS. The full suite reaches the same unrelated existing `testrig.sh` ASan failure in `dummy_get_level()`.


N0NB:
Thanks David.

PR #2108, opened 2026-07-25
fix: prevent dummy level buffer overflow

dchristle:
I found this while running Hamlib’s `testrig` under AddressSanitizer. The dummy backend copied a full `value_t` into the result buffer even when `rig_get_strength()` supplied an `int`, causing a stack buffer overflow.

The fix keeps the existing alignment-safe copy but limits it to the selected integer or float member.

I ran the focused `testrig` test and the full Hamlib test suite under ASan/UBSan on macOS.


N0NB:
Thanks, David.

Also cherry-picked into 4.7.

PR #2109, opened 2026-07-25
fix: remove Morse busy hint race

dchristle:
I confirmed this while exercising Morse sending and getter calls under ThreadSanitizer. The asynchronous Morse worker writes `rig_state.morse_busy` while getter APIs read it without a common lock, creating a data race on the plain `bool`.

The fix is small: the existing field remains unchanged, but its reads and writes now use relaxed atomic operations. Relaxed ordering is sufficient because `morse_busy` is only an advisory cache hint and does not publish any other state.

This is intentionally limited to the `morse_busy` race; it does not change Morse handling, cache behavior, or the public rig state layout.

I reproduced the race before the fix and confirmed that the same focused TSan workload no longer reports it. The full build and all 12 tests pass on macOS.

### DIFF
--- a/src/rig.c
+++ b/src/rig.c
@@ -188,6 +188,17 @@ const char hamlib_copyright[231] = /* hamlib 1.2 ABI specifies 231 bytes */
 // Rig lock for all front side thread control
 #define LOCK(n) rig_lock(rig,n)
 
+static bool morse_busy_load(const struct rig_state *rs)
+{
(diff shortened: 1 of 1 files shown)


N0NB:
Thanks, David.

Cherry pinking this into 4.7 resulted in a number of merge conflicts.  This would probably be a good one to backport if possible.

PR #2110, opened 2026-07-26
FTX-1: fix get/set_rptr_offs to use correct unit

shankerwangmiao:
The raw value of frequency offset for 144 MHz and 430 MHz band is in 50 kHz unit, and the available range is from 0 to 100 MHz.

Tested on real hardware.

Fixes: 2c0ba29d53fd8a67 ("FTX-1: Fix capability gaps - DCD, CWPITCH, RPTR_OFFS, BAND_UP/DOWN")


KJ5HST:
Merge recommended. 

However, one other non-blocking problem that I also noticed in the existing code while reviewing this PR: val.f = (float)(offs / multiplier) is integer division, so offsets that aren’t a multiple of the unit truncate silently. With the corrected 50 kHz granularity on 2 m that’s easier to hit — set_rptr_offs 625000 returns 600000 on my radio with no indication anything was rounded. Might be worth rounding to nearest and/or returning -RIG_EINVAL on a non-representable value. If you want to also fix this, that would be helpful, otherwise, I'll do it and push a separate PR.

shankerwangmiao:
I believe rounding would make sense. It will be nice if the change on the multiplier can be included in a separate PR, since it is not so tightly related to this change.

KJ5HST:
I'll work on that.

N0NB:
Thanks, Miao.

This has also been cherry picked into the Hamlib-4.7 branch and will be included in 4.7.3.

shankerwangmiao:
> Thanks, Miao.
> 
> This has also been cherry picked into the Hamlib-4.7 branch and will be included in 4.7.3.

Many thanks. Will you consider the other two patches that I have proposed to be also included in 4.7.3?

N0NB:
Yes, I'm working through the PRs as time permits and I'll cherry pick those I think will fit well in the 4.7 series.

PR #2111, opened 2026-07-27
Validate Elecraft K3/K4 SWR responses

dchristle:
## What changed

This is a small defensive fix for Elecraft SWR reads.

The [K3S/K3 Programmer's Reference](https://ftp.elecraft.com/KX2/Manuals%20Downloads/K3S%26K3%26KX3%26KX2%20Pgmrs%20Ref%2C%20G5.pdf) documents `SWnnn` replies, while the [K4 Programmer's Reference](https://ftp.elecraft.com/K4/Manuals%20Downloads/K4%20Programmer%27s%20Reference%2C%20rev.%20D12.pdf) documents SWR as the final field in `TMaaabbbcccddd`.

The existing code already converts the documented tenths-of-an-SWR-unit values correctly. This change simply checks that the numeric fields were parsed before using them and lets the K4 path return transaction or parsing errors instead of reporting a successful reading.

Normal, valid radio replies are handled exactly as before. An unexpected reply now produces a regular protocol error rather than a potentially bogus SWR value.

## Testing

- Full build completed successfully.
- All 13 tests passed.
- Checked valid and malformed K3 and K4 responses with a simulated serial endpoint.

### DIFF
--- a/rigs/kenwood/k3.c
+++ b/rigs/kenwood/k3.c
@@ -2347,8 +2347,7 @@ static int k3_get_level(RIG *rig, vfo_t vfo, setting_t level, value_t *val)
     case RIG_LEVEL_SWR:
         if (RIG_IS_K4)
         {
-            retval = k4_get_bar_graph_level(rig, &val->f, NULL, NULL, NULL);
-            return RIG_OK;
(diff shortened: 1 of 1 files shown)


N0NB:
Thanks, David.

Cherry picked into 4.7 as well.

PR #2113, opened 2026-07-28
fix(guohetec): initialize fallback filter width

dchristle:
Initialize the PMR-171 and Q900 cached filter width to `RIG_PASSBAND_NORMAL` during backend setup. This keeps fallback mode queries from returning an undefined width.

Tested with `make check` and hardware-independent PMR-171/Q900 checks under Clang ASan/UBSan.


N0NB:
Thanks, David.

Also cherry picked into 4.7.

PR #2114, opened 2026-07-28
fix(core): terminate empty hexadecimal output

dchristle:
Make `to_hex()` return a terminated empty string when there is no source data and the destination has space. This prevents callers such as the spectrum snapshot serializer from treating an untouched destination buffer as a C string. Add a small regression for the empty conversion.

Tested with the focused `testbcd` test, the zero-length spectrum snapshot loopback harness, and the full `make check` suite (20 tests).


N0NB:
Thanks, David.

Cherry picked into 4.7.

PR #2115, opened 2026-07-28
fix(dummy): copy string parameter values

dchristle:
The Dummy backend was keeping caller-owned pointers for string parameters, which left `BANDSELECT` and `KEYERTYPE` dangling after the command parser returned. A later get could then read from an expired stack buffer and trigger a stack-use-after-return under ASan.

This change:

- Copies string values into backend-owned storage.
- Frees them when replaced or during cleanup.
- Leaves the existing numeric-parameter behavior unchanged.
- Adds focused set/get regression coverage for `BANDSELECT` and `KEYERTYPE`.

### DIFF
--- a/rigs/dummy/dummy.c
+++ b/rigs/dummy/dummy.c
@@ -353,6 +353,14 @@ static int dummy_cleanup(RIG *rig)
     free(priv->ext_parms);
     free(priv->magic_conf);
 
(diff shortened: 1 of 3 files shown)


N0NB:
Thanks, David.

Merge conflicts prevented me from cherry picking this into 4.7.

PR #2117, opened 2026-07-28
ftx1: Fix CTCSS/DCS tone handling and add CSQL support

shankerwangmiao:
Correct several issues in the FTX-1 CTCSS/DCS implementation:

* Add DCS code table (ftx1_dcs_codes[]) and index<->code conversion helpers (ftx1_code_to_dcs_num/ftx1_dcs_num_to_code). Previously set_dcs_code/get_dcs_code passed the raw DCS code value (e.g. 023) directly to the CN command, which expects a 0-based table index (000-103). The code value is now converted to/from the index on set/get respectively.

* Fix CN command parameter documentation. P1 is the VFO selector (0=Main/1=Sub), not a TX/RX selector as previously stated.

* Simplify set/get_ctcss_sql to delegate to set/get_ctcss_tone, and set/get_dcs_sql to delegate to set/get_dcs_code. The FTX-1 does not support separate TX/RX tones or DCS codes, so the previous CN10/CN11 (Sub VFO) usage was incorrect.

* Add RIG_FUNC_CSQL support in set_func/get_func and rig caps, mapping it to the CT3 (DCS) mode via FTX1_CTCSS_MODE_DCS.


KJ5HST:
Confirmed. Merge recommended.

N0NB:
Thanks, Miao.

Also Cherry picked into 4.7.

PR #2118, opened 2026-07-28
ftx1: handle WFM mode quirk (FM-in-WFM-band returns mode '0')

shankerwangmiao:
On the FTX-1, WFM is a bit special: Inside the 76-108 MHz WFM band the radio is set to WFM by sending the normal FM mode code (MD...4), and when read back the MD reply carries P2='0', which newcat_get_mode cannot map and so returns -RIG_EPROTO. Override get_mode/set_mode with ftx1 wrappers:

- ftx1_get_mode: on -RIG_EPROTO with mode '0', verify the current VFO frequency falls in a WFM rx range (rx_range_list1), and if so report RIG_MODE_WFM with a 200 kHz default width instead of the protocol error.
- ftx1_set_mode: when asked for RIG_MODE_WFM, reject it (-RIG_EINVAL) outside the WFM range, and inside the range translate it to RIG_MODE_FM with RIG_PASSBAND_NOCHANGE (let the radio apply its WFM bandwidth).
- ftx1_check_freq_wfm(): helper scanning rx_range_list1 for a WFM-typed entry covering the given frequency.


KJ5HST:
**One thing I'd change before merge.** `ftx1_check_freq_wfm()` matches with exact equality:

```c
if (rig->caps->rx_range_list1[i].modes == RIG_MODE_WFM)
```

That works today only because the caps row happens to be exactly `RIG_MODE_WFM` — but that row understates the hardware. My radio accepts non-FM modes in the broadcast band:

```
(at 98.500 MHz)
MD02;   -> accepted, MD0; reads back MD02;    (USB in the FM broadcast band)
```

So `{MHz(76), MHz(108), RIG_MODE_WFM, ...}` is arguably wrong as-is, and the moment anyone corrects it to include the other modes the rig actually receives there, this equality test silently stops matching and the WFM handling breaks with no compile error and no obvious symptom. `& RIG_MODE_WFM` would be robust against that.

**Two smaller points:**

- `ftx1_check_freq_wfm()` reads `rig->caps->rx_range_list1` directly. The FTX-1 declares identical WFM entries in `rx_range_list1` and `rx_range_list2`, so it's correct today, but the frontend selects the range list by region — using the resolved list would avoid a latent region-dependent bug.

- `*width = s_kHz(200)` is a reasonable value (I checked: `SH0;` returns `SH0000;` in WFM, so the filter query gives nothing useful and a constant really is the right call here). But `ftx1_caps.filters` currently has no WFM entry at all, so `rig_passband_normal(rig, RIG_MODE_WFM)` returns 0 and the reported width is disconnected from the caps that clients read. Adding `{RIG_MODE_WFM, kHz(200)}` to `.filters` and calling `rig_passband_normal()` would keep the two in sync. Note `.tuning_steps` already lists WFM at 25 kHz, so WFM is partially declared in caps already.

Minor: the `extern int ftx1_get_mode(...)` declaration lands in `ftx1.c` under the `/* Wrappers from ftx1_func.c for rig caps */` comment, but the function is in `ftx1_mode.c`. The existing `ftx1_set_mode` extern has the same problem, so this is pre-existing — `ftx1.h` would be the natural home for both. My bad on the original code there, too.

shankerwangmiao:
Hi, thanks for your comment. By only searching the code, I failed to understand the usage of `filters` and `rx_range_list`, since when setting frequencies and modes, the lists are not actually referred to. I first thought that I could have use `& RIG_MODE_WFM`. However, other bands declares `RIG_MODE_ALL`, which includes `RIG_MODE_WFM`, and thus  `& RIG_MODE_WFM` cannot work.

shankerwangmiao:
I have corrected the frequency table and amended `ftx1_check_freq_wfm` accordingly.

N0NB:
Thanks, Miao.

Also cherry picked into 4.7.

Issue #1979, opened 2026-01-02
Xiegu G90 Hamlib 4.6.5 Hamlib error with D19 CAT


marcoleptic123:
Hello, i cannot understant how resolve this:

**Describe the bug**
This message appears in WSJT-X. G90 START in TX when i start the TEST CAT in WSJT-X.
 I am not sure if this is a bug.

**Desktop**
 - OS: Linux Mint 22.2 Cinnamon 
 - WSJT-X V2.7.0
 - G90 firmwares: Base V1.81 and APP v1.81 
 - D19 Xiegu

**Try this with rigctl**
`~$ rigctl -m G90 -r /dev/USB0 -s 19200 -C stop_bits=2 -vvvvv`

The return:
`rigctl.c(475) Startup: rigctl -m G90 -r /dev/USB0 -s 19200 -C stop_bits=2 -vvvvv`
`rigctl Hamlib 5.0.0~git from indeterminate source revision. 64-bit`
`Report bugs to <hamlib-developer@lists.sourceforge.net>`

`rig_test_2038: enter _TIME_BITS=64, __TIMESIZE=64 testing enabled for GLIBC 2.39`
`rig_init: 2038 time test passed`
`rig_check_rig_caps: p1=0x7d042cdcd380, p2=0x7d042cdd3908, rig_model=0x7d042cdcd380, macro_name=0x7d042cdd3908`
`initrigs5_dummy: _init called`
`Unknown rig num 0, or initialization error.`
`Please check with --list option.`

**The error in WSJT-X:**

Hamlib error: write_block(): TX 7 bytes
0000    fe fe a4 e0 19 00 fd                                .......         
read_string_generic called, rxmax=200 direct=1, expected_len=1
read_string_generic(): RX 41 characters, direct=1
0000    ce ff f5 c3 ef 3f fe e9 29 7f ff f5 e7 7a f5 92     .....?..)....z..
0010    ed bf ea e5 c9 f1 ff 1f ef f7 83 6f c7 f0 ef ef     ...........o....
0020    f7 ed f5 ef 7e eb 67 da fd                          ....~.g..       
**2:frame.c(330):icom_one_transaction returning(-8) Protocol error

Protocol error

Protocol error

*1:frame.c(556):icom_transaction returning(-8) Protocol error

ser_close: restoring options
rig.c(1479):rig_open returning2(-8) Protocol error

Protocol error
Protocol error
 while opening connection to rig

**Screenshots**
[...]

PR #2121, opened 2026-07-31
G90: fix ATT ACK, get_info, AGC map, default CI-V address, and split get

cjlawson02:
Stock Xiegu G90 (model 3088) CAT on the DispUnit serial port does not fully
match the generic Icom paths Hamlib was using. This updates `rigs/icom/xiegu.c`
only. Backend version `.11` → `.12`.

Tested on a stock G90 (base/APP v1.81), front User jack, 19200 8N1, RX-only.
Verbose `rigctl` traces (tip baselines + patched backend `.12`) are attached
as [hamlib-3088-g90-evidence.zip](https://github.com/user-attachments/files/30567684/hamlib-3088-g90-evidence.zip). Related report: #1979 (open probe TX to
address `0xa4`).

## What changed

- **ATT (`0x11`):** Stock firmware answers with `11 <state>` (`00` off /
  `0c` on) instead of `FB`, and the old Icom BCD `11 12` set path timed out
  (`-5`) even when the radio toggled. G90 now uses a small set/get path:
  read state, write `11 00` / `11 01` only when needed, accept the state
  echo (same idea as `x108g_set_ptt()` accepting a non-`FB` reply). Get maps
  any non-zero state to the 12 dB attenuator value in caps.

- **Radio ID / `get_info`:** Open preferred the classic Icom `0x19/0x00`
  probe and never wired `.get_info`, so `\get_info` returned `None`. G90 now
  probes `0x1D/0x19` first (`00 90` → G90 on this firmware), keeps
  `0x19/0x00` as fallback, and returns the probed model string.

- **AGC levels:** Numeric/enum control already talked on the wire; caps did
  not list OFF/FAST/MEDIUM/SLOW, and the default Icom AGC byte map does not
  match Xiegu (`00/01/02/03`). Added the usual `agc_levels_present` map
  (same pattern as IC-7300 and friends). `rigctl` still wants Hamlib AGC
  enum values (`0=OFF 2=FAST 5=MEDIUM 3=SLOW` per dumpcaps), not free-text
  names.

- **Default CI-V address:** `0xa4` → `0x88`. Vendor / flrig-style default.
  This bench unit also answers `0xa4` and `0x70`, so `-C civaddr=0xa4`
  still works here. X6100 stays `0xa4`.

- **Split get:** `.get_split_vfo` was `NULL`, so Hamlib never polled and
  always showed split off after a successful set. Wired to
  `icom_get_split_vfo` (`0x0F`).

## Testing

- Full rebuild; `rigctl -m 3088 -u` shows backend `20250517.12` and
  `AGC levels: 0=OFF 2=FAST 5=MEDIUM 3=SLOW`.
- Default open (no `-C`): TX uses `fe fe 88 e0 …`; `f` / `m` OK.
- `-C civaddr=0xa4 f` still OK on this multi-address unit.
- `\get_info` → `G90`; open log shows `1d 19 00 90`.
- ATT: tip `L ATT 12` was `-5` with TX `11 12` / RX `11 0c`. Patched set
  uses read-then-`11 01`, get returns `12`, second identical set is
  idempotent, `L ATT 0` restores off.
- AGC: `L AGC 0/2/5/3` → TX `16 12 00/01/02/03`; get returns matching
  enums.
- Split: `S 1 VFOB` then `s` → `1`; `S 0 VFOA` then `s` → `0` (polls
  `0x0F`).
- PTT (`0x1C`) not exercised (RX-only lab policy).

## Notes for reviewers

- Default address change can surprise users who relied on Hamlib’s old
  `0xa4` default with no override; override remains supported.
- `xiegu_info_buf` is a small process-static string filled at open (same
  pattern class as other simple info helpers). Fine for normal single-rig
  use; multi-instance edge case not hardened here.
- Happy to split this into smaller PRs (ATT / ID / AGC / civaddr / split)
  if that is easier to review.

### DIFF
--- a/rigs/icom/xiegu.c
+++ b/rigs/icom/xiegu.c
@@ -119,11 +119,17 @@
  * Prototypes
 */
 static int x108g_set_ptt(RIG *rig, vfo_t vfo, ptt_t ptt);
+static int g90_set_level(RIG *rig, vfo_t vfo, setting_t level, value_t val);
+static int g90_get_level(RIG *rig, vfo_t vfo, setting_t level, value_t *val);
+static const char *g90_get_info(RIG *rig);
 static int x108g_set_split_vfo(RIG *rig, vfo_t vfo, split_t split,
                                vfo_t tx_vfo);
 static int x108g_set_split_freq(RIG *rig, vfo_t vfo, freq_t tx_freq);
 static int x108g_set_split_mode(RIG *rig, vfo_t vfo, rmode_t tx_mode,
                                 pbwidth_t tx_width);
+
+/* Filled by xiegu_rig_open so \get_info can return the probed model. */
+static char xiegu_info_buf[64];
 
 static int x108g_rig_open(RIG *rig)
 {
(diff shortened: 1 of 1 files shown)


jbdubbs:
Testing on my hardware G90 yielded the following AGC mapping:

0x00=OFF 0x01=Fast 0x02=Slow 0x03=Auto

Different firmware versions?  My unit was on 1.79.  Did they change this in later revisions?  I don't have a real MEDIUM mapping at all, even though the manual shows it.

cjlawson02:
@jbdubbs Thanks for checking this!

My unit was on FW 1.81, and I mostly verified that `00/01/02/03`worked under the Hamlib AGC names. I mapped 00/01/02/03 to Hamlib’s OFF/FAST/MEDIUM/SLOW names from the round-trip, unfortunately not from matching the radio’s own AGC-S / AGC-F / AGC-A labels. Let me try that out, your result lines up better with what the stock head UI seems to use. So I suspect the PR’s `MEDIUM` label is wrong, not that our firmwares disagree on the wire bytes.

Quick question: when you labeled `01/02/03`, was that from the G90 screen itself (AGC-F / AGC-S / AGC-A), or from somewhere else?

jbdubbs:
> @jbdubbs Thanks for checking this!
> 
> My unit was on FW 1.81, and I mostly verified that `00/01/02/03`worked under the Hamlib AGC names. I mapped 00/01/02/03 to Hamlib’s OFF/FAST/MEDIUM/SLOW names from the round-trip, unfortunately not from matching the radio’s own AGC-S / AGC-F / AGC-A labels. Let me try that out, your result lines up better with what the stock head UI seems to use. So I suspect the PR’s `MEDIUM` label is wrong, not that our firmwares disagree on the wire bytes.
> 
> Quick question: when you labeled `01/02/03`, was that from the G90 screen itself (AGC-F / AGC-S / AGC-A), or from somewhere else?

Thanks for your hard work @cjlawson02 

I had verified this from the hardware G90 screen itself.  But being on different firmware version, and my G90 being borrowed and since returned, I cannot verify at this time if your PR matches the newer firmware version.  If it's not obvious to anyone else here, I mayyyyy be able to figure out how to reborrow the unit and upgrade the firmware.

dforsi:
This is a case when putting unrelated changes in a single commit, stalls the whole PR.
@cjlawson02 , @jbdubbs do you think that we can commit everything even if we are unsure about the AGC mapping, then close this PR and reopen it or a new one if we find that different firmware revisions need different code? Or it's better to move the AGC part to a separate PR and commit the rest?

cjlawson02:
Apologies I can split if needed. Let me validate the AGC tonight, I'm inclined that @jbdubbs is correct

jbdubbs:
> This is a case when putting unrelated changes in a single commit, stalls the whole PR. @cjlawson02 , @jbdubbs do you think that we can commit everything even if we are unsure about the AGC mapping, then close this PR and reopen it or a new one if we find that different firmware revisions need different code? Or it's better to move the AGC part to a separate PR and commit the rest?

I was using an older firmware version and I no longer have access to the radio.  I'll bow out on this one and defer to @cjlawson02 since he can test on the latest firmware.

cjlawson02:
@dforsi Sorry for the delay here - verified and updated the AGC accordingly

- 0x00 OFF
- 0x01 Fast (AGC-F)
- 0x02 Slow (AGC-S)
- 0x03 Auto (AGC-A)

N0NB:
Thanks, Chris.

Also cherry picked into 4.7.

PR #2123, opened 2026-07-31
build(tests): compile AGC parser helper directly

dchristle:
PR #2106 added `testnetrigctl` and linked it against the dummy backend archive. When building from the tests directory, Automake could reuse a stale archive that did not contain the AGC parser helpers, causing undefined-symbol link errors.

Compile `dummy_common.c` directly into `testnetrigctl` so the code under test is rebuilt when needed.

Tested with the normal and ASan/UBSan configurations.


dchristle:

N0NB:
Thanks, David.

Feel free to port to 4.7.  Git/vimdiff are showing a large block of conflicting text and I can't see why.  I think you ought to do this!

PR #2124, opened 2026-07-31
fix(netrigctl): make protocol output locale-independent

### LINKED ISSUE #2054: Incorrect output from netrigctl depending on the language environment
I am using the wfview program on my Linux computer to connect to my ICOM transceiver.
Digi-mode programs such as FreeDV, FlDigi, and QSSTV connect to wfview’s rigctl server via Hamlib.NET (m=2). This works to some extent. The programs can also read the transceiver’s frequency. However, the digi-mode programs are unable to change the transceiver’s frequency.

**Error and Expected behavior**
An analysis of the data traffic has revealed that, for example, the command `F 7177000,000000` is issued to change the frequency. However, the correct command here would be `F 7177000.000000`.

The command is likely issued from the Hamlib module `netrigctl.c`. Line 968:
``` c
SNPRINTF(cmd, sizeof(cmd), “F%s %‘FREQFMT’\n”, vfostr, freq);
```
where `FREQFMT = “lf”`     from `hamlib/rig.h`

However, the floating-point number generated is adapted to the current locale.
Since my `LC_NUMERIC` is set to `de_DE.UTF-8`, the number is generated with a comma.

I found a workaround where I change the Linux environment variable to `LC_NUMERIC="C"` before starting a Digi-Mode program, but this may also have its disadvantages elsewhere.

Hamlib should ensure that all inputs and outputs in communicating data, involving floating-point numbers, always use a decimal point, regardless of the language environment. There is reason to fear that this error occurs not only here, but also in other places and modules.

**Software**
`rigctl Hamlib 4.7.1 2026-04-15T20:20:01Z SHA=d042479a9 64-bit`
[wfview v2.22rc1](https://github.com/eliggett/wfview/releases/tag/v2.22rc1)
Linux Manjaro & Cinnamon

73 de Dieter, DK6OV

dchristle:
Format floating-point NET rigctl commands with a decimal point regardless of the process locale, without changing process-global locale state.

This covers frequency, level, parameter, and power-conversion commands. The tests verify exact wire output under a comma-decimal locale, including the previously undersized power-conversion command buffer.

Tested with the focused `testnetrigctl` test and the full `make check` suite.

Fixes #2054


dchristle:

N0NB:
@dchristle There is a merge conflict that must be resolved.  I'm unsure which way to resolve it.

dchristle:
Hey Nate, this is updated now—the changes had already landed via #2131, so I rebased the branch and the conflict is resolved.

PR #2125, opened 2026-08-01
fix(ft991): extend ID meter range to 25.5 A

### LINKED ISSUE #2073: FT991A: ID_CAL table capped at 10A, should go higher
FT991_ID_CAL in rigs/yaesu/ft991.h stops at raw 107 = 10.0A.

The meter is linear: raw value / 10 = Ampere. The correct table is:
{0, 0.0f}, {255, 25.5f}

The existing FT991-specific override should be removed or replaced with the
above so that yaesu_default_id_meter_cal (or equivalent) covers the full range.

dchristle:
The FT-991/FT-991A ID meter calibration ended at raw 107, causing higher current readings to clamp at 10 A. Map the full 8-bit meter range linearly from 0 to 25.5 A.

Advertise the matching 0–25.5 A range with 0.1 A granularity instead of the normalized 0–1 range. Other Yaesu backends retain their existing capability metadata.

Add hardware-independent regression coverage for conversion values across the full range and the advertised meter capabilities.

Tested with the focused `testft991idmeter` test and the full `make check` suite.

Fixes #2073


dforsi:
You are right, the man page says that ID_METER retunrs Amps

dforsi:
The MacOS (features) workflow fails on:
> Test rx_sequence_continuous...                  [ FAILED ]+

but MacOS (no features) doesn't fail.

dforsi:
I used the web UI to resolve the merge conflict because there were just 2 more lines added in a `Makefile.am` but it made a merge from `master` to `dchristle/fix/2073-ft991a-id-meter-range` with
> 137 files changed, 56033 insertions(+), 1453 deletions(-)

I didn't see a preview of that, I'm not sure there was one available :-(

then it made a merge in the opposite direction, with

>  5 files changed, 80 insertions(+), 7 deletions(-)

which looks good, but it took me much longer to understand what happened, than editing those 2 lines in my working copy by hand...

dforsi:
cherry-picked to the `Hamlib-4.7` branch with PR #2185

PR #2112, opened 2026-07-27
Harden command and backend parser bounds

dchristle:
## Summary

- Bound command fields using their actual destination capacities, reject empty command input before indexing it, and safely decode raw hexadecimal commands in rigctl, ampctl, and rotctl.
- Stop tuning-step, status-token, and metadata parsing at explicit protocol boundaries, returning the printable `BANDGEN` fallback when band metadata is absent.
- Correct GS100 response framing and reject empty or oversized extracted values.
- Validate FTX-1 clarifier, extended-menu, and S-meter replies against their expected command, receiver, framing, and numeric ranges.

## Testing

- Plain focused parser tests: 7 passed
- Plain `make check`: 20 passed
- Clang ASan+UBSan focused parser tests: 7 passed
- Clang ASan+UBSan `make check`: 20 passed
- Clang ASan+UBSan C++ suite: 1 passed
- Sanitizer testing repeated at current HEAD on macOS with `--without-readline`
- Structured command-loop replay for empty commands, raw hexadecimal commands, band lookup, channel descriptions, and unsupported Icom tuning steps

These cases were identified through fuzzing and sanitizer-backed robustness testing.

PR #2127, opened 2026-08-01
docs(news): correct my surname

dchristle:
A few NEWS entries list my surname as Christie instead of Christle. Fix the acknowledgements for PRs #2105, #2106, and #2112 :)


N0NB:
Sorry about that.  Evidently I need better fonts/eyeballs!

I do wear reading glasses...

N0NB:
Thanks, David.

Sometimes something gets stuck in the mind and it becomes impervious to being corrected unless pointed out.  Sigh...

Corrected in the 4.7 branch as well.

dchristle:
No problem at all :) This exact change of `l` -> `i` is one I've encountered my whole life, so you're not the first and definitely won't be the last! 😆 It's all good.

PR #2129, opened 2026-08-01
fix(icom): correct ID-5100 VFO switching

dchristle:
The ID-5100 frequency callbacks were switching VFOs internally while the Hamlib core also handled selection for non-targetable VFOs. In dualwatch mode, the getter additionally sent `16 59 00`, which the radio manual defines as enabling single-watch mode. This caused repeated D0/D1 switching, disabled dualwatch, and could leave B selected.

Keep the frequency callbacks on the current operating side and let the core perform one target, read, and restore sequence. Preserve current and explicit A/B requests during dualwatch, reject targeted access when the initial side is unknown and cannot be restored, and correct frequency response and error handling.

The ID-5100 simulator now models D0/D1, dualwatch, frequency replies, and CI-V response addresses consistently with the documented protocol. A strict CI-V transcript test covers current-frequency reads, A/B targeting and restoration, rejected commands, and unchanged dualwatch state.

Tested with the `testid5100` test and the full `make check` suite (21 tests).

Fixes #2103

Issue #2103, opened 2026-07-22
ID-5100A switching active VFO regardless of which one was requested


adamculbertson:
**Describe the bug**
Using the command line `rigctl -m 3071 -r /dev/ttyUSB0 -s 19200  -c 0x8C --vfo -vvvvv f VFOA` to get the output of `VFOA` causes a weird VFO switching behavior on the radio. Regardless of whether channel `A` or `B` are selected, the active VFO *always* gets switched to `B`. In addition, it will switch rapidly between `A` and `B` two to three times before stopping on `B`. This means that the frequency returned from `rig_get_freq` could be incorrect if VFO `A` was the currently active VFO. It also will turn off dual-watch on the radio most of the time. The same behavior is observed regardless of whether `--vfo` is passed.

In case it's relevant, I'm using Icom's `OPC-478UD` cable to communicate with the radio.
In this output, VFO `A` should be `146.760` MHz and VFO `B` should be `145.490` MHz. 

**To Reproduce**
Steps to reproduce the behavior (please use `rigctl` with the `-vvvvv` option if at all possible and post the output under number 4 below):
1. `rigctl` model: `3071` (ID-5100)
2. The function that fails, i.e. `rig_open`, `rig_set_freq`, `rig_get_freq`, etc.: `rig_get_freq`
3. Hamlib version: `4.7.1`
4. Post `rigctl` output:
```
rig_test_2038: enter _TIME_BITS=64, __TIMESIZE=64 testing enabled for GLIBC 2.43
rig_init: 2038 time test passed
rig_check_rig_caps: p1=0x7f4c609ef100, p2=0x7f4c609f5688, rig_model=0x7f4c609ef100, macro_name=0x7f4c609f5688
initrigs4_icom: _init called
rig_init: rig_model=Icom ID-5100 20250517.9
Requesting 47928 bytes for rig_struct
Requesting 728 bytes for rig_cache
rig_init: rig has VFO_A
rig_init: rig has VFO_B
rig_init: rig has VFO_SUB
rig_init: rig has VFO_MAIN
rig.c(995) trace
*1:icom.c(707):icom_init entered
icom_set_x25x26_ability: Hamlib thinks rig does not support x25/x26 command
*1:icom.c(777):icom_init returning(0) 
rig_set_conf called
rig_confparam_lookup called for rig_pathname
rig_set_conf: rig_pathname='/dev/ttyUSB0'
rig_token_lookup called for civaddr
[...]

PR #2130, opened 2026-08-01
fix(icom): backport ID-5100 VFO switching to 4.7

dchristle:
Backport of #2129 to `Hamlib-4.7`.

This carries the ID-5100 dualwatch and VFO targeting fix for #2103, with the regression test registration adapted to the 4.7 test layout.

Tested with `testid5100` and the full `make check` suite (13 tests).


N0NB:
Thanks, David.

PR #1787, opened 2025-06-28
Improve error handling for GUOHETEC drivers to enhance WSJT-X compatibility

FVsonar:
This PR addresses compatibility issues between WSJT-X and GUOHETEC radio drivers (PMR-171 and Q900) by implementing more robust error handling mechanisms.

### Problem Statement
WSJT-X and similar applications were experiencing communication failures when temporary issues occurred with GUOHETEC radios, causing the applications to crash or become unresponsive instead of gracefully handling the errors.

### Solution
- **Graceful Error Degradation**: Instead of returning error codes that cause application failures, the drivers now return cached values when communication issues occur
- **Unified Response Validation**: Implemented common validation functions to ensure consistent error handling across all GUOHETEC drivers
- **Cache Return Macros**: Added standardized macros for returning cached values, ensuring consistent behavior
- **Backward Compatibility**: All changes maintain full backward compatibility with existing applications

### Technical Changes
- Modified `rigs/guohetec/guohetec.c` and `rigs/guohetec/guohetec.h` to add common error handling functions
- Updated `rigs/guohetec/pmr171.c` to use the new error handling mechanisms
- Enhanced response validation to prevent buffer overflows and invalid data access
- Added proper fallback mechanisms for frequency, mode, VFO, and PTT operations

### Benefits
- **Improved Stability**: WSJT-X and similar applications continue working even during temporary communication issues
- **Better User Experience**: Users no longer experience crashes when radio communication is temporarily interrupted
- **Maintained Functionality**: All existing features continue to work as expected
- **Future-Proof**: The new error handling framework can be easily extended to other GUOHETEC models

### Testing
The changes have been tested with:
- WSJT-X application
- PMR-171 radio
- Q900 radio
- Various communication scenarios including temporary disconnections

### Files Changed
- `rigs/guohetec/guohetec.c` - Added common error handling functions
- `rigs/guohetec/guohetec.h` - Added function declarations and macros
- `rigs/guohetec/pmr171.c` - Updated to use new error handling

This PR ensures that Hamlib provides a more robust and user-friendly experience when working with GUOHETEC radios, particularly in scenarios where communication may be temporarily interrupted.

PR #1781, opened 2025-06-25
Add support for Guohetec PMR171/Q900 radios with Yaesu config cleanup

FVsonar:
## New Features
- Added support for Guohetec devices:
- PMR171
- Q900 
- Created dedicated vendor directory: `rigs/guohetec/`
- Follows Hamlib's vendor classification convention
﻿
## Technical Changes
1. **New files**:
rigs/guohetec/guohetec.c
rigs/guohetec/guohetec.h
rigs/guohetec/pmr171.c
rigs/guohetec/q900.c
2. **Yaesu directory cleanup** (commit e9f5022):
- Removed residual PMR171/Q900 definitions from:
- `yaesu/ft817.c`
- `yaesu/yaesu.h`
- `yaesu/yaesu.c`
- Deleted incorrectly classified legacy code
﻿
3. **Build system updates**:
- Added `rigs/guohetec/Makefile.am`
- Updated top-level `configure.ac`
﻿
## Testing Verification
- **Tested devices**:
- Guohetec PMR171
- Guohetec Q900 
﻿
- **Verified functionality**:
- [x] Frequency set/get
- [x] Mode switching (FM/DMR)
- [x] Signal strength readout
- [x] PTT control

PR #2128, opened 2026-08-01
refactor(tests): make test registration less conflict-prone

dchristle:
This tidies up the test registration in tests/Makefile.am so adding a test does not mean editing the same giant lines every time.

- Keeps test entries sorted and one per line.
- Shares direct-test and generated-wrapper lists where possible.
- Removes a duplicate CLEANFILES entry.
- Does not change which tests are built or run.

Main goal: fewer annoying merge conflicts when multiple PRs add tests.

Validated with autoreconf -fi and a fresh configure.

PR #2133, opened 2026-08-01
fix(guohetec): parse status frames by advertised length

### LINKED ISSUE #2059: Guohetec PMR-171 / Q900 (v2 hardware, firmware 3.7.6): get_freq fails due to status-frame length-byte interpretation
## Summary

The Guohetec PMR-171 and Q900 ship as v2 hardware running firmware **3.7.6**, which changes the framing convention for CAT status replies versus earlier firmware. The current `pmr171` / `q900` backends fail on this firmware because `validate_freq_response()` in `rigs/guohetec/guohetec.c` interprets the length byte differently than the rig now emits.

I have the hardware (PMR-171, v2, firmware 3.7.6) and a partial working patch — happy to iterate or test maintainer-proposed fixes.

## Hardware / environment

- Rig: Guohetec PMR-171, hardware revision v2, firmware 3.7.6
- OS: Windows 10/11, also reproduced via mingw-w64 cross-build from macOS
- Hamlib: master (`b43eca4e4` at the time of testing) — backend version 20250611.0
- Connection: USB CDC (COM3), 115200 baud, 8N1

## Symptom

Every `rig_get_freq` call fails and returns cached (zero) values:

```
pmr171_get_freq: Response too large for buffer: 34 > 32, using cached values
```

WSJT-X / SatPC32 / any host software sees the frequency stuck at 0 and cannot Doppler-track or follow band changes.

## Root cause

Firmware 3.7.6 emits the 32-byte status frame documented in the PMR-171 User Manual V2.1, Appendix 2, command `0x0B`:

```
0xA5 0xA5 0xA5 0xA5 | len | 0x0B | state | VFOA mode | VFOB mode |
VFOA freq (4) | VFOB freq (4) | A/B | NR/NB | RIT | XIT |
filter BW | spectrum BW | voltage | UTC time (3) |
status bar | S/PO | SWR/AUD/ALC | CRC high | CRC low
```

Per the manual: *"Package length: One byte, which represents the length of bytes from the next byte in the packet length to the end of the packet."* — i.e. the length byte's value counts **command echo + payload + CRC**.

For this status frame that gives `len = 27`, total frame = `5 + 27 = 32 bytes`, with CRC at `reply[30..31]` computed over `reply[4..29]` (26 bytes).

The current driver in `validate_freq_response()` computes:

```c
int expected_total_length = 5 + reply[4] + 2;   // assumes len = payload only
uint16_t recv_crc = (reply[31] << 8) | reply[32];
uint16_t calc_crc = CRC16Check(&reply[4], 27);
```

That `+ 2` and the hardcoded CRC offsets assume the earlier firmware's framing (len = data bytes only, CRC outside the count). For 3.7.6 (`reply[4] = 27`), this computes `expected_total_length = 34`, fails the buffer-size check, and bails before parsing.

It looks like firmware 3.7.6 now matches the published manual literally — so arguably the earlier firmware was the deviation and the driver was relying on that deviation.

## Reproducing

```
rigctld -m 40001 -r COM3 -s 115200 -vvvv
```

Then from another terminal:

```
rigctl -m 2 -r 127.0.0.1:4532 f
```

Returns `0`. The daemon log shows `Response too large for buffer: 34 > 32, using cached values` on every `get_freq` call.

## Other commands affected

The same length-byte issue appears in other paths I haven't fully diagnosed yet — verbose logs show:

- `get_ptt`: `Invalid data length 71, using cached values`
- `set_split_vfo` and other `_send_cmd2` callers: `Failed to read header` (215ms timeout)
- `get_mode`: not yet exercised but uses the same hardcoded-offset `validate_mode_response()` in `guohetec.c` — likely affected.

I suspect a single fix to the shared validators (make them length-byte-aware rather than hardcoded for the earlier frame size) would address all of these, plus protect against future firmware tweaks.

## My local workaround

I have a patched build where `pmr171_get_freq` does length-aware CRC validation inline (treating `reply[4]` as "all bytes after itself, incl. CRC"). Verified working on the hardware. I haven't submitted a PR because:

1. Only `get_freq` is fixed; PTT/mode/set_* are still broken.
2. My local approach duplicated the file with a new model ID — almost certainly *not* the right design for upstream. A shared length-aware validator (and possibly a model variant flag, or runtime detection of firmware version via the `0x27` device-type command) would be cleaner.

Happy to:

- Provide the working patch as a starting point if useful.
- Test any approach you'd prefer against the v2 hardware.
- Gather verbose logs for specific commands (set_freq ACK, get_ptt, get_mode, etc.) on request.

Just let me know how you'd like to proceed.

dchristle:
## Summary

Fix PMR-171 and Q900 status-response framing.

Guohetec's length byte includes the command, payload, and trailing CRC. For status command `0x0B`, `length = 27`, making the complete response 32 bytes:

```text
A5 A5 A5 A5 | 1B | command + payload + CRC
5-byte header + 27 bytes = 32 bytes
```

The driver instead read a fixed 40 bytes, expected the CRC outside the advertised length, and accessed nonexistent byte 32. Valid responses therefore timed out or failed validation and returned cached values.

I have a branch prepared with some followup bug fixes to the Guotec  backend, but those changes are localized enough to the ones in this  PR that it makes sense to wait till this one merges & then I'll open the followup PR. These other fixes are for issues that aren't reported in #2059.

## History

The earlier PMR-171 implementation correctly read the five-byte header followed by the advertised length.

During #1781, this was replaced with fixed-size reads and hardcoded CRC offsets. A bounds warning was addressed by enlarging the buffers, which made the access memory-safe without correcting the protocol arithmetic. PR #1787 then moved the incorrect calculation into the shared validator.

Captured firmware 3.5 traffic and the firmware 3.7.6 documentation use the same 32-byte layout, so no firmware-specific framing is needed.

## Fix

- Read the five-byte header, then exactly the advertised number of bytes.
- Locate and validate the CRC from the resulting frame length.
- Share status decoding across PMR-171, Q900, and probing.
- Preserve existing cached-value fallback on invalid responses.

## Tests

Added hardware-free coverage using captured wire data and socketpair transactions for both models, including malformed lengths, headers, commands, CRCs, stream synchronization, and cached fallback.

```text
Focused test: 1/1 passed
Full C suite: 21/21 passed
C++ suite: 1/1 passed
```

No physical radio was used.

Fixes #2059.

### DIFF
--- a/rigs/guohetec/guohetec.c
+++ b/rigs/guohetec/guohetec.c
@@ -19,7 +19,6 @@
  *
  */
 
-#include <string.h>
 #include "hamlib/rig.h"
 #include "hamlib/port.h"
 #include "iofunc.h"
@@ -27,24 +26,23 @@
 #include "riglist.h"
 #include "guohetec.h"
 
-// Common response validation function implementations
-int validate_packet_header(const unsigned char *reply, const char *func_name)
+enum
+{
+    GUOHE_FRAME_HEADER_LENGTH = 5,
+    GUOHE_STATUS_MIN_PACKET_LENGTH = 15,
+    GUOHE_STATUS_COMMAND = 0x0B
+};
+
+static int validate_packet_header(const unsigned char *reply,
+                                  const char *func_name)
 {
     if (reply[0] != 0xA5 || reply[1] != 0xA5 || 
         reply[2] != 0xA5 || reply[3] != 0xA5) {
         rig_debug(RIG_DEBUG_ERR, "%s: Invalid packet header, using cached values\n", func_name);
-        return -1;
+        return -RIG_EPROTO;
     }
-    return 0;
-}
 
-int validate_data_length(const unsigned char *reply, int reply_size, const char *func_name)
-{
-    if (reply[4] == 0 || reply[4] > reply_size - 5) {
-        rig_debug(RIG_DEBUG_ERR, "%s: Invalid data length %d, using cached values\n", func_name, reply[4]);
-        return -1;
-    }
-    return 0;
+    return RIG_OK;
 }
 
 // CRC16/CCITT-FALSE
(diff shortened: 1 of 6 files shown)


dchristle:

N0NB:
Thanks, David.

Of course the cherry picking generates a merge conflict, but then as I'm working with `tests/Makefile.am` I realize that the ID5100 changes aren't in 4.7 yet, so this would need to wait for that, I'd think otherwise there will be a merge conflict on the ID5100 stuff.  This is getting complicated...

PR #2134, opened 2026-08-01
fix(core): preserve DATA submodes for generic packet requests

### LINKED ISSUE #1990: Regression in hamlib 4.6: KENWOOD TS-990S forces USB-DATA1 on every TX/RX and QSY
**Describe the bug**

When using hamlib 4.6 or later with KENWOOD TS-990S, the radio mode is forcibly reset to USB-DATA1 on every TX/RX transition and on every QSY, even if the radio is manually set to USB-DATA2 or USB-DATA3.

This behavior occurs when the client application (WSJT-X or JTDX) is configured with Mode = "Data/Pkt".

The forced re-selection of USB-DATA1 can be confirmed by the audible Morse code notification "UDA", which indicates that USB-DATA1 has been set. The repeated occurrence of "UDA" shows that USB-DATA1 is being re-set each time.

This behavior was not present in hamlib 4.5.5 and earlier.

**To Reproduce**
1. rigctl model: KENWOOD TS-990S

2. The function that fails: rig_set_mode (triggered during TX/RX transitions and QSY)

3. Hamlib version: 4.6 (and later)

4. Post rigctl output: Not yet captured. The issue is reproducible via client applications using hamlib (WSJT-X or JTDX).

**Expected behavior**
If the radio is already set to any USB-DATA mode (USB-DATA1, USB-DATA2, or USB-DATA3), hamlib should not re-set the mode during TX/RX transitions or QSY.

This matches the behavior observed with hamlib 4.5.5 and earlier.

**Screenshots**
Not applicable.

**Desktop (please complete the following information):**
OS: Windows 10 / Windows 11

Client application: WSJT-X or JTDX

Client application version: WSJT-X v2.7.0 (also reproducible with JTDX)

**Additional context**
This behavior appears to be specific to the KENWOOD TS-990S.
The behavior change is not documented in the hamlib 4.6 release notes and may be an unintended side effect of internal changes to mode or state handling.

dchristle:
## What changed

Hamlib 4.6 added distinct `USBD1`/`USBD2`/`USBD3` and `LSBD1`/`LSBD2`/`LSBD3` modes. On the TS-990S, that exposed a compatibility gap in `rig_set_mode()`: the current exact mode no longer compared equal to a generic `PKTUSB` request, so the Kenwood backend mapped the request to `USBD1` and sent `OM0D`. WSJT-X's `Data/Pkt` setting could therefore replace a manually selected USB-DATA2 or USB-DATA3 during transmit/receive transitions and QSY.

Kenwood's [PC command reference](https://www.kenwood.com/i/products/info/amateur/pdf/ts990_pc_command_en_rev2.pdf) defines `D`, `H`, and `L` as USB-DATA1, USB-DATA2, and USB-DATA3, while its [FT8 setup guide](https://www.kenwood.com/i/products/info/amateur/ts_990/pdf/ts990_ft8_settings_en.pdf) configures WSJT-X for `Data/Pkt` and supports Data Mode 1 through 3.

Treat an active USB-DATA1/2/3 mode as satisfying a generic `PKTUSB` request, and apply the same rule to LSB-DATA1/2/3 and `PKTLSB`. Exact submode requests remain directional, and a generic packet request still changes the mode when the radio is not already in the matching DATA family.

This preserves exact DATA submode reporting while restoring the no-op behavior that clients relied on before Hamlib 4.6.

I also made a corresponding PR in WSJTX to make the mode detection there have a better experience w.r.t. these various data submodes. It should be included in the 3.2.0 release.

## Testing

Added a hardware-independent public-API regression test covering:

- `PKTUSB` preserving `USBD2`
- `PKTLSB` preserving `LSBD3`
- an explicit `USBD1` request replacing `USBD2`
- `PKTUSB` still changing plain `USB`

```text
Focused test: 1/1 passed
Full C suite: 21/21 passed
C++ suite: 1/1 passed
```

No physical TS-990S was used.

Fixes #1990

### DIFF
--- a/rigs/kenwood/kenwood.c
+++ b/rigs/kenwood/kenwood.c
@@ -2592,12 +2592,7 @@ int kenwood_set_mode(RIG *rig, vfo_t vfo, rmode_t mode, pbwidth_t width)
     }
     else if (RIG_IS_TS990S)
     {
-        /* The TS990s has targetable read mode but can only set the mode
-           of the current VFO :( So we need to toggle the operating VFO
-           to set the "back" VFO mode. This is done here rather than not
-           setting caps.targetable_vfo to not include
-           RIG_TARGETABLE_MODE since the toggle is not required for
-           reading the mode. */
+        /* OM reads address either band, but writes affect the operating band. */
         vfo_t curr_vfo;
 
         err = kenwood_get_vfo_main_sub(rig, &curr_vfo);
@@ -2614,16 +2609,16 @@ int kenwood_set_mode(RIG *rig, vfo_t vfo, rmode_t mode, pbwidth_t width)
         SNPRINTF(buf, sizeof(buf), "OM0%c", c);  /* target vfo is ignored */
         err = kenwood_transaction(rig, buf, NULL, 0);
 
-        if (err == RIG_OK && vfo != RIG_VFO_CURR && vfo != curr_vfo)
+        if (vfo != RIG_VFO_CURR && vfo != curr_vfo)
         {
             int err2;
 
             err2 = kenwood_set_vfo_main_sub(rig, curr_vfo);
 
-            if (err2 != RIG_OK) { RETURNFUNC2(err2); }
+            if (err == RIG_OK) { err = err2; }
         }
 
-        return RIG_OK;
+        RETURNFUNC2(err);
     }
     else
     {
--- a/rigs/kenwood/ts990s.c
+++ b/rigs/kenwood/ts990s.c
@@ -31,7 +31,8 @@
 
 #define TS990S_AM_MODES RIG_MODE_AM
 #define TS990S_FM_MODES (RIG_MODE_FM|RIG_MODE_FMN)
-#define TS990S_OTHER_MODES (RIG_MODE_CW|RIG_MODE_CWR|RIG_MODE_SSB|RIG_MODE_FM|RIG_MODE_RTTY|RIG_MODE_RTTYR|RIG_MODE_PKTUSB|RIG_MODE_PKTLSB|RIG_MODE_USBD1|RIG_MODE_USBD2|RIG_MODE_USBD3|RIG_MODE_LSBD1|RIG_MODE_LSBD2|RIG_MODE_LSBD3)
+#define TS990S_PACKET_MODES (RIG_MODE_PKTUSB|RIG_MODE_PKTLSB|RIG_MODE_USBD1|RIG_MODE_USBD2|RIG_MODE_USBD3|RIG_MODE_LSBD1|RIG_MODE_LSBD2|RIG_MODE_LSBD3)
+#define TS990S_OTHER_MODES (RIG_MODE_CW|RIG_MODE_CWR|RIG_MODE_SSB|RIG_MODE_FM|RIG_MODE_RTTY|RIG_MODE_RTTYR|TS990S_PACKET_MODES)
 #define TS990S_HP_MODES (TS990S_OTHER_MODES|TS990S_FM_MODES)
 #define TS990S_ALL_MODES (TS990S_OTHER_MODES|TS990S_AM_MODES|TS990S_FM_MODES)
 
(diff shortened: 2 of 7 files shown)


dforsi:
This PR may be need to consider more cases.

One case is because `mode_satisfies_request()` is called only if `width == RIG_PASSBAND_NOCHANGE` (which is `-1`) so when using `simulators/simts990`  I get a mode change if I ask `PKTUSB` and a bandwidth when the current mode is one of USBD*

```
$ tests/rigctl -m 2039 -r /dev/pts/3 M USBD2 2700 m M PKTUSB -1 m
USBD2
2700
USBD2
2700
$ tests/rigctl -m 2039 -r /dev/pts/3 M USBD2 2700 m M PKTUSB 2700 m
USBD2
2700
PKTUSB
2700
```

The other case is that it doesn't work for LSB:
```
$ tests/rigctl -m 2039 -r /dev/pts/3 M LSBD2 2700 m M PKTUSB -1 m
LSBD2
2700
PKTUSB
2700
$ tests/rigctl -m 2039 -r /dev/pts/3 M LSBD2 2700 m M PKTUSB 2700 m
LSBD2
2700
PKTUSB
2700
```

dchristle:
@dforsi Thanks for checking these! The explicit-bandwidth case was a gap in my original fix. It only preserved the DATA profile when the width was `-1`, so specifying a bandwidth still hit the existing behavior that selected DATA1. I’ve pushed an update that preserves the matching DATA profile in both cases, while still passing an explicit bandwidth request through to the backend.

For the LSB example, I think the mode change is expected: `LSBD2` to `PKTUSB` requests a change from lower to upper sideband. Preserving `LSBD2` there would ignore the requested sideband. The corresponding preservation case is `LSBD2` to `PKTLSB`, which now keeps DATA2 with either `-1` or an explicit bandwidth.

I also tightened up the VFO/cache handling and added regression coverage. I'd like to do a separate followup PR after this one merges for actual TS-990S bandwidth programming. This PR just addresses preserving the DATA profile.

N0NB:
@dchristle It looks like something has crept in that is causing a failure since the CI runs:

```
Making all in src
make  all-am
  CC       rig.lo
../../hamlib/src/rig.c: In function 'rig_set_mode':
../../hamlib/src/rig.c:2950:27: error: 'rs' undeclared (first use in this function)
 2950 |                 || vfo == rs->current_vfo))
      |                           ^~
../../hamlib/src/rig.c:2950:27: note: each undeclared identifier is reported only once for each function it appears in
make[2]: *** [Makefile:677: rig.lo] Error 1
make[1]: *** [Makefile:532: all] Error 2
make: *** [Makefile:674: all-recursive] Error 1
```

N0NB:
@dchristle The following will quell the error:

``` diff
diff --git a/src/rig.c b/src/rig.c
index 84de8273e..30e3a0339 100644
--- a/src/rig.c
+++ b/src/rig.c
@@ -2882,6 +2882,7 @@ static int mode_satisfies_request(rmode_t requested, rmode_t current)
 int HAMLIB_API rig_set_mode(RIG *rig, vfo_t vfo, rmode_t mode, pbwidth_t width)
 {
     const struct rig_caps *caps;
+    struct rig_state *rs;
     struct rig_cache_routing_snapshot routing;
     int retcode;
     int locked_mode;
@@ -2921,6 +2922,7 @@ int HAMLIB_API rig_set_mode(RIG *rig, vfo_t vfo, rmode_t mode, pbwidth_t width)
     }
 
     caps = rig->caps;
+    rs = STATE(rig);
 
     if (caps->set_mode == NULL)
     {
```

If you agree, I'll apply this patch and get this PR merged.

N0NB:
Now the tests fail:

Screenshot From 2026-09-19 06-16-41

Attaching the `test-suite.log` file.

[test-suite.log](https://github.com/user-attachments/files/32414015/test-suite.log)

dchristle:
@N0NB Thanks—I tracked this down. The PR’s Actions matrix did pass on September 10, but it was tested against the older cb275c6b base. The cache/routing synchronization changes merged on September 17 removed direct current_vfo access and added a test that rejects direct routing-state reads and writes.

Adding rs makes the compilation error go away, but it then causes testroutingaccess.sh to fail; the TS-990S test fixture also needs to use the new routing setters.

I don’t think we should apply the proposed rs patch as-is. I’ll rebase onto current master, change the core check to rig_get_current_vfo_state(), update the fixture to use rig_set_vfo_state() and rig_set_rx_vfo_state(), and rerun the full suite. This looks like base drift rather than a problem with the DATA-mode behavior itself.

N0NB:
Thanks you for fixing things.  The intricacies of all of this has moved far beyond my understanding.  I know I'm slow and trying to dig out from the PR list.  Don't get me wrong, I am grateful for all of the contributions!

N0NB:
@dchristle 
Thanks, David.

I've given up trying to cherry pick most of these PRs as the divergence between master and 4.7 is that large now.  I will leave it to your discretion if you think a set of patches would be a good fit for 4.7 and submit a new PR for that.

dchristle:
@N0NB yep, makes sense. I just opened #2228 👍

Issue #1963, opened 2025-12-11
IC-7100 rejecting the 'm' command


rgmiller:
**Describe the bug**
My Icom IC-7100 radio rejects the `rigctl` "m" command with an error.

**To Reproduce**
Steps to reproduce the behavior (please use `rigctl` with the `-vvvvv` option if at all possible and post the output under number 4 below):
1. `rigctl` model: 3070
2. The function that fails, i.e. `rig_open`, `rig_set_freq`, `rig_get_freq`, etc.: `rig_get_mode`
3. Hamlib version:
4. Post `rigctl` output:
(`rigctl` with 1 v posted below.  Txt file attached with the same command with 5 v's.)
xxxxxx@xxxx:/opt/hamlib-4.6.3/bin$ ./rigctl -v -m 3070 -r /dev/ttyUSB0 m
Opened rig model 3070, 'IC-7100'
error = icom.c(2951):icom_get_mode returning2(-9) Command rejected by the rig

rig_get_mode: retcode after get_mode=-9
rig_get_mode(3063): freqMainA=7073000, modeMainA=, widthMainA=0
rig_get_mode(3063): freqMainB=0, modeMainB=, widthMainB=0
**2:cache.c(40):rig_set_cache_mode entered
rig_set_cache_mode(42): freqMainA=7073000, modeMainA=, widthMainA=0
rig_set_cache_mode(42): freqMainB=0, modeMainB=, widthMainB=0
rig_set_cache_mode(155): freqMainA=7073000, modeMainA=, widthMainA=0
rig_set_cache_mode(155): freqMainB=0, modeMainB=, widthMainB=0
**2:cache.c(156):rig_set_cache_mode returning(0) 
rig_get_mode(3118): freqMainA=7073000, modeMainA=, widthMainA=0
rig_get_mode(3118): freqMainB=0, modeMainB=, widthMainB=0
rig_lock: client lock disengaged
*1:rig_get_mode: elapsed=18ms
*1:rig.c(3122):rig_get_mode returning(-9) Command rejected by the rig

rigctl_parse.c(2344):rigctl_get_mode returning2(-9) Command rejected by the rig

Command rejected by the rig

xxxxxx@xxxx:/opt/hamlib-4.6.3/bin$ 

[max_verbosity.txt](https://github.com/user-attachments/files/24091826/max_verbosity.txt)

**Expected behavior**
The same `rigctl` command on v4.6.2 returns:
[...]

PR #2136, opened 2026-08-04
fix(icom): fall back from rejected optional CI-V commands

dchristle:
## What changed

Older IC-7100 firmware rejects the optional CI-V `0x25` and `0x26` commands. Hamlib recorded the rejection but returned `RIG_ERJCTED` from the first API call, so clients such as WSJT-X failed before the existing legacy fallback could run.

When either optional command is rejected during initial capability probing, retry immediately with legacy `0x03` or `0x04`. Mode fallback also queries the legacy `0x1A 0x06` data-mode state, preserving USB-DATA reporting as `PKTUSB`.

Only clean rejections of an unknown optional capability trigger fallback. Required-command rejections, later rejections, malformed replies, timeouts, and transport errors remain visible.

This fix was motivated by a reported issue on the WSJTX mailing list: https://wsjtx.groups.io/g/main/topic/mac_and_ic7100_won_t_work/120590167

## Testing

Added public-API regression coverage for:

- old-firmware fallback from `0x25` and `0x26`
- successful operation when the optional commands are supported
- legacy data-mode detection returning `PKTUSB`
- required-command rejections and malformed replies remaining errors

Full C test suite: 22/22 passed.

Physical IC-7100 validation: the original reporter tested this runtime change cherry-picked onto Hamlib 4.7.2 in an Apple Silicon WSJT-X 3.0.2 test build. On the affected radio, the original `rig_get_mode` `RIG_ERJCTED` startup failure was eliminated, and he completed an on-air 15 m QSO. The tested Hamlib commit was `175b751`.

Related to #1963.

Backport: #2138.


N0NB:
@dchristle It looks like this needs the `socket.h` remedy as well.

``` console
  CC       testicomfallback-testicomfallback.o
../../../tests/testicomfallback.c:14:10: fatal error: sys/socket.h: No such file or directory
   14 | #include <sys/socket.h>
      |          ^~~~~~~~~~~~~~
```

N0NB:
Also three merge conflicts in `tests/Makefile.am`.

dchristle:
@N0NB Yep you're right -- just pushed an update. Thanks for the review.

dforsi:
Than you David.
This was already merged in 4.7 with commit dbcb17c8e8be174c7f8d730ce37c33a3a8aceaeb

PR #2138, opened 2026-08-05
fix(icom): backport optional command fallback to 4.7

dchristle:
## Summary

- Backport #2136 for the forthcoming 4.7.3 release.
- Fall back to legacy frequency and mode commands when optional CI-V 0x25 and 0x26 capability probes are rejected.
- Preserve required-command rejection and malformed-response errors.
- Add focused simulated CI-V regression coverage for old and current IC-7100 behavior.

## Backport notes

- Based on the current Hamlib-4.7 branch, not the 4.7.2 tag.
- Adapted only the test registration to the 4.7 layout; no unrelated master tests are included.

## Testing

- make -j4
- ./tests/testicomfallback
- make check: C tests 13/13 passed; C++ test 1/1 passed

The equivalent 4.7.2 backport was subsequently validated by the original reporter on the affected IC-7100 through a WSJT-X 3.0.2 test build; the original startup failure was eliminated and an on-air 15 m QSO completed successfully.


N0NB:
Thanks, David.

PR #2140, opened 2026-08-06
QMX: add SWR and RF power meter readback

jasiek:
QMX: add SWR and RF power meter readback

The QMX backend previously defined no `.get_level` at all, so
`rig_get_level(RIG_LEVEL_SWR, ...)` returned `-RIG_ENAVAIL`. QRP Labs
firmware does provide the readings as extensions on top of the TS-480
command set:

* `SW;` returns SWR in hundredths (`SW121;` = 1.21:1)
* `PC;` returns measured output power in tenths of a watt (`PC45;` = 4.5 W)

Both are transmit-only measurements. In receive the QMX answers with a
bare `SW;` / `PC;`, which the backend reports as 0 rather than passing
back a stale or garbage value.

`RIG_LEVEL_RFPOWER_METER` is scaled against 5 W for the 0.0..1.0 form.

`level_gran_kenwood.h` gains `NO_LVL_SWR` / `NO_LVL_RFPOWER_METER_WATTS`
guards, following the existing `NO_LVL_KEYSPD` pattern, so the QMX can
declare its own resolution (0.01 SWR, 0.1 W, 5 W full scale).

## Testing

Tested against a real QMX, firmware `1_04_003QMX`, high-band unit
(20/17/15/12/11/10 m), macOS, `/dev/cu.usbmodem101`.

Receive — both meters correctly report no reading:

```
$ rigctl -m 2057 -r /dev/cu.usbmodem101 l SWR l RFPOWER_METER_WATTS
0
0
```

Transmitting a carrier — live values:

```
$ rigctl -m 2057 -r /dev/cu.usbmodem101 T 1 w 'TA1500;' \
      l SWR l RFPOWER_METER_WATTS l RFPOWER_METER w 'TA0;' T 0
1.79
3.6
0.72
```

Capability advertisement:

```
$ rigctl -m 2057 -r /dev/cu.usbmodem101 -u | grep '^Get level'
Get level: SWR(0.000000..99.989998/0.010000) \
    RFPOWER_METER(0.000000..1.000000/0.003922) \
    RFPOWER_METER_WATTS(0.000000..5.000000/0.100000)
```

### Band sweep

Readings across the unit's configured bands into a real antenna, to
exercise the parsing over a range of values rather than a single point.
Each row is a short keyed burst; SWR is averaged over the samples taken
while the PA was actually producing output, and the spread column shows
the min/max of those samples.

```
band  freq MHz   SWR       spread      power   valid samples
20m   14.020     --        --          --      no RF
20m   14.175     1.79      1.69-1.91   3.9 W   6/8
20m   14.330     1.90      1.89-1.91   3.8 W   6/8
17m   18.075     --        --          --      no RF
17m   18.118     --        --          --      no RF
17m   18.160     --        --          --      no RF
15m   21.020     1.68      1.67-1.70   6.6 W   6/8
15m   21.225     1.64      1.64-1.64   6.6 W   6/8
15m   21.430     1.65      1.64-1.66   6.6 W   6/8
12m   24.895     --        --          --      no RF
12m   24.940     --        --          --      no RF
12m   24.985     1.69      1.67-1.75   4.5 W   6/8
10m   28.020     1.76      1.76-1.77   5.2 W   6/8
10m   28.500     1.79      1.76-1.83   5.2 W   6/8
10m   29.000     1.76      1.76-1.77   5.2 W   6/8
10m   29.600     1.76      1.74-1.77   5.1 W   6/8
```

The "no RF" rows are the rig declining to transmit, not a readback
failure — most likely its SWR/PA protection. They are included because
they exercise the receive-path branch: with the PA off, `PC;` reports
0.0 W and `SW;` returns either noise or a bare `SW;`, and the backend
handles both without a protocol error.

Two behaviours worth knowing for anyone consuming these levels:

* An SWR reading is only meaningful when `RFPOWER_METER_WATTS` is
  non-zero. With no RF the bridge returns noise (values wandering
  between 1.0 and 9.98 were observed).
* `set_ptt`/`TX;` alone does not key the QMX — it only sets transmit
  state. RF starts on the first `TA` command (digi) or on key-down, and
  the rig returns to receive if the tone is not refreshed. A client that
  keys and immediately reads SWR will see 0.


jasiek:
In case this is relevant - this change was created with the aid of Claude, and tested against real hardware.

jasiek:
Hey. I wanted to get this reviewed and maybe merge in - is there a contribution step I missed to have this reviewed?

N0NB:
Hi Jan.

I just approved the work flows.  I've been very very busy this week and my air conditioner has not been keeping this end of the house cool enough to spend time in here the past several days.  Today is a slight exception.  Hopefully @dforsi  Daniele can take a look.

dforsi:
@jasiek Thanks for this PR and your patience.

PR #2141, opened 2026-08-07
fix (ats-mini): replace strncpy() with memcpy()

jonsnow1357:


dforsi:
Thanks Jon.

rigs/ats-mini/ats.c
```diff
@@ -98,7 +98,12 @@ static int p_ats_update_state(RIG *rig) {
                 break;
             }
             case 4: {
-                strncpy(tmp_mon_data->bandname, token, (MINI_BANDNAME_LEN - 1));
+                if (strlen(token) >= MINI_BANDNAME_LEN) {
+                    rig_debug(RIG_DEBUG_ERR, "UNEXPECTED bandname");
+                    free(tmp_mon_data);
+                    return RIG_EPROTO;
```
dforsi, line 104: I merged this PR as is, because this `return` is consistent with other places in `rigs/ats-mini/ats.c`, however the convention of Hamlib is returning error codes negated with a minus sign. I think that it works here if they are used inside `ats.c` because they are checked against `!= RIG_OK` but it could fail if they are returned to a client application.

It would be great if you could look into adding minus signs to returns.

jonsnow1357: will try  :+1:

PR #2146, opened 2026-08-12
fix(cache): synchronize cache and routing state

dchristle:
Hamlib starts a poll worker for every open rig. That worker checks cached frequency, mode, width, split, PTT, and VFO routing state while normal CAT calls and backend handlers update the same fields. Those accesses did not have common synchronization.

I found the races while running WSJT-X under ThreadSanitizer. They reproduced during both startup and live FT8 operation.

Most racing reads probably return either the previous or next value on common 64-bit systems, which helps explain why this has not produced an obvious widespread failure. Multicast publication is also disabled by default, so many inconsistent poll observations have no visible output. Other cases can associate a cache update with the wrong VFO, publish an inconsistent routing snapshot, or treat stale state as fresh. A user could therefore see an occasional stale or incorrect VFO, frequency, mode, width, or split result that corrects itself on the next CAT update or cache expiration and looks like an ordinary radio or serial synchronization problem.

## What changed

- Make the rig cache private to `cache.c` and protect its state with one mutex.
- Route compiled cache access through synchronized getters, setters, and coherent snapshots.
- Publish current, RX, TX, and split routing state through a single synchronized path.
- Separate the assumed current VFO from an observed VFO reported by the radio, so unrelated operations do not make stale VFO state appear fresh.
- Invalidate VFO freshness after operations such as toggle or exchange when the resulting selection is not known.
- Read one coherent cache snapshot from the poll worker and synchronize its startup, scheduling, and shutdown.
- Keep multicast publication cadence independent from cache expiration while preserving the historical 1000 ms cache default.
- Convert affected frontend and backend call sites rather than leaving part of the repository on unsynchronized direct access.
- Add a source guard that rejects new direct cache or routing-state access outside the cache implementation.

The unused `src/multicast.c` implementation is not built and remains unchanged.

## Testing

- Normal build
- ThreadSanitizer build
- Full C test suite: 27/27 passed
- C++ test suite: 1/1 passed
- Focused cache-routing test under ThreadSanitizer: passed without a cache or routing race
- Source guard verified against direct access, arbitrary `rig_state` aliases, `HAMLIB_STATE()`, and `state_addr`

No physical radio was used.


src/cache.h
```diff
@@ -36,32 +36,21 @@ __BEGIN_DECLS
...
-    freq_t freqCurr; // Other VFO
-    freq_t freqOther; // Other VFO
-    freq_t freqMainA; // VFO_A, VFO_MAIN, and VFO_MAINA
-    freq_t freqMainB; // VFO_B, VFO_SUB, and VFO_MAINB
-    freq_t freqMainC; // VFO_C, VFO_MAINC
-    freq_t freqSubA;  // VFO_SUBA -- only for rigs with dual Sub VFOs
-    freq_t freqSubB;  // VFO_SUBB -- only for rigs with dual Sub VFOs
-    freq_t freqSubC;  // VFO_SUBC -- only for rigs with 3 Sub VFOs
-    freq_t freqMem;   // VFO_MEM -- last MEM channel
+struct rig_cache_snapshot
```
mikaelnousiainen, line 49: Should we still document here the purpose of this struct and - in some detail - what all the fields are used for? It's not always obvious...

src/cache.h
```diff
@@ -71,60 +60,59 @@ struct rig_cache {
...
-    struct timespec time_split;
-    int satmode; // if rig is in satellite mode
+    vfo_t split_vfo;
+    int satmode;
 };
 
-/* Access macros */
-#define CACHE(r) ((r)->cache_addr)
-//#define HAMLIB_CACHE(r) ((struct rig_cache *)rig_data_pointer(r, RIG_PTRX_CACHE))
+struct rig_cache_routing_snapshot
```
mikaelnousiainen, line 93: Should we add a comment of the purpose of this struct?

mikaelnousiainen:
👍 Thanks for adding the comments, this definitely cleans up cache access. Nate will likely do a final review of this.

N0NB:
To be honest, my "review" is primarily just testing compilation and reporting warnings/errors back to the respective PR.  I lost track of the internals long, long ago!  I do look things over, but I don't work through the logic of all changes.  I just don't have that bandwidth.

mikaelnousiainen:
> To be honest, my "review" is primarily just testing compilation and reporting warnings/errors back to the respective PR. I lost track of the internals long, long ago! I do look things over, but I don't work through the logic of all changes. I just don't have that bandwidth.

@N0NB 👍 Totally understandable. I did a brief review on what the changes look like here and I think it looks good, leaving the code and logic cleaner than what it was before. Tests are always a welcome addition (like in this PR!).

P.S: Also, while I may not be able to look at every PR (there's been quite a lot of activity lately!), I try to review the major ones to some extent and flag if something needs to be reconsidered.

N0NB:
@dforsi @mikaelnousiainen @GeoBaltz 

CI is failing in the `testguohetec` run.  You can drill down through the CI run to the section titled "Dump test logs on failure" and see that it ends with the following:

```
pmr171_get_freq: CRC check failed (received: 14B1, calculated: 14B0), using cached values
```

I tried the same `configure` and `make` command lines locally and the `distcheck` does fail at the same place, but the referenced `test-suite.log` is not written locally.

As the CI runs do not preserve the colored output, here is a screenshot of the failure:

Screenshot From 2026-09-14 06-07-23

N0NB:
Ahh, found the `test-suite.log` file and have attached it.
[test-suite.log](https://github.com/user-attachments/files/32260758/test-suite.log)

N0NB:
@dchristle David, can you help resolve the `testguohetec` error?  I know I'm way behind on merges so all help is appreciated.

N0NB:
Thanks, David.

PR #2154, opened 2026-08-16
fix(tests): clear rigctld PID after startup timeout

dchristle:
When rigctld failed to start, the test code stopped and reaped the process but kept its old PID. Retry cleanup could then try to stop that same PID again, which might belong to another process if the operating system had already reused it. Clear the PID after reaping the failed child so retries cannot signal a stale process.


dforsi:
Thanks David

N0NB:
Thanks, David.

PR #2155, opened 2026-08-17
fix(rigctld): stop stream feeders before closing rig

dchristle:
`rigctld` stream feeders retain backend stream handles owned by the rig's streaming state. Several rig-close paths could destroy that state before associated feeders stopped, leaving a feeder able to perform another read or write through an invalid handle.

The most direct case occurs in `--rigctld-idle` mode when the last client disconnects with an open stream. The same teardown ordering also existed during hard-error recovery and daemon shutdown.

This change centralizes rig shutdown around the required lifetime order:

1. Remove the applicable streams from the `rigctld` registry.
2. Signal all affected feeders to stop.
3. Join feeder threads outside the registry lock.
4. Release their backend streams and UDP resources.
5. Close the rig.

Normal disconnects close the departing client's streams before the last-client idle transition. Operations that close the shared rig first close every registered stream, because those handles cannot survive a rig close. The registry remains reusable after a close/reopen cycle.

The same close helper is used at idle startup and clears `rig_opened`, ensuring the first client reopens the rig instead of issuing commands against a closed session. Stream pause, resume, and drain handlers also retain the registry lock until the public API has acquired and released its internal lifetime guard, preventing feeder auto-close from invalidating a borrowed backend handle between lookup and use.

Client thread launch cleanup now destroys initialized thread attributes and releases the accepted socket and connection data when setup or `pthread_create()` fails.

Regression coverage includes:

- closing streams across multiple clients and stream types;
- stopping and joining an active feeder;
- repeated close-all calls and registry reuse;
- disconnecting an idle-mode client without `stream_close`;
- confirming idle startup and later client connections reopen the rig successfully;
- verifying pause, resume, and drain enter their backend hooks while the stream handle remains registry-protected.

Validated with the full test suite and focused ASan/UBSan runs. No public Hamlib API or stream wire-format changes.

### DIFF
--- a/tests/rigctl_parse.c
+++ b/tests/rigctl_parse.c
@@ -7043,13 +7043,10 @@ declare_proto_rig(stream_pause)
 
     if (!stream) { RETURNFUNC2(err); }
 
-    /* Release the registry lock before the (possibly blocking) backend call so
-     * one client's pause cannot stall every other client's stream command. The
-     * backend stream cannot be freed underneath us: rig_stream_close() drains
-     * its in-use count, which this call holds for its duration. */
-    rig_stream_t *backend = stream->backend_stream;
+    /* Keep the registry-owned handle pinned until the public API has entered
+     * and released its internal stream-lifetime guard. */
+    retval = rig_stream_pause(rig, stream->backend_stream);
     rigctld_stream_registry_unlock(&g_stream_registry);
-    retval = rig_stream_pause(rig, backend);
     RETURNFUNC2(retval);
 }
 
@@ -7067,10 +7064,8 @@ declare_proto_rig(stream_resume)
 
     if (!stream) { RETURNFUNC2(err); }
 
-    /* See stream_pause: drop the registry lock before the backend call. */
-    rig_stream_t *backend = stream->backend_stream;
+    retval = rig_stream_resume(rig, stream->backend_stream);
     rigctld_stream_registry_unlock(&g_stream_registry);
-    retval = rig_stream_resume(rig, backend);
     RETURNFUNC2(retval);
 }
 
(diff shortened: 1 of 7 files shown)


tests/rigctl_parse.c
```diff
@@ -7043,13 +7043,10 @@ declare_proto_rig(stream_pause)
...
     if (!stream) { RETURNFUNC2(err); }
 
-    /* Release the registry lock before the (possibly blocking) backend call so
-     * one client's pause cannot stall every other client's stream command. The
-     * backend stream cannot be freed underneath us: rig_stream_close() drains
-     * its in-use count, which this call holds for its duration. */
-    rig_stream_t *backend = stream->backend_stream;
+    /* Keep the registry-owned handle pinned until the public API has entered
+     * and released its internal stream-lifetime guard. */
+    retval = rig_stream_pause(rig, stream->backend_stream);
```
mikaelnousiainen, line 7048: Just my 2 cents about potential effects of these changes: I understand the idea behind this change, but it'd be great to be able to test this with real network-connected rigs. Basically, my original idea was to avoid doing I/O that communicates with the rig when the lock is being held. I guess we will find out soon once we get the Icom and FlexRadio backends out in "heavier" use. Some of this logic may need revisiting.

mikaelnousiainen (approved):
👍 Looks good, but I left a comment about the lock being held during pause/resume/drain - for us as a reminder if we start seeing some odd behavior once we get network rig backends out for real testing.

N0NB:
@dchristle @mikaelnousiainen 

Is this ready to merge or should this be converted to a draft for the time being?

mikaelnousiainen:
@N0NB With the latest changes here, I think we're good to merge this. It definitely fixes some edge cases.

dchristle:
@N0NB @mikaelnousiainen I rebased it & it's ready to merge.

N0NB:
Thanks, David and Mikael.

PR #2156, opened 2026-08-17
fix(ftx1): correct DCS parser format type

dchristle:
The FTX-1 DCS getter parsed a signed integer into an unsigned variable and then logged it using a signed format. This caused a format type mismatch and undefined behavior.

Use a signed integer for the DCS table index, matching the parser, debug format, and range-check helper.

Tested with a warning-enabled full build and the FTX-1 parser test.


N0NB:
@dchristle 

Thanks, David.

Also cherry picked into 4.7.

PR #2157, opened 2026-08-17
fix(rigctld): enforce password authentication consistently

### LINKED ISSUE #2065: rigctld --help text says "-A, --password ... (NOT IMPLEMENTED)" but password IS partially wired
## Summary

`tests/rigctld.c:1395` prints help text claiming `-A, --password=PASSWORD` is `(NOT IMPLEMENTED)`. The password feature is, in fact, partially implemented (lines 110, 124, 149, 324–327, 1031, 1222, 1226). Operators reading `--help` will either:

1. Believe the password is non-functional and not set one (leaving the daemon unauthenticated even when they intended otherwise), or
2. Be uncertain about whether the option actually works and either disable it or use a weak password assuming it's a stub.

Either failure mode is bad for any defense-relevant deployment (MARS, SHARES, satellite GS, etc.) where operators read `--help` first and trust the labels.

This is a follow-up to private security advisory [GHSA-f72v-7gmh-m9mj](https://github.com/Hamlib/Hamlib/security/advisories/GHSA-f72v-7gmh-m9mj) filed earlier today. The advisory covers the underlying password-implementation weaknesses (global `is_passwordOK`, `strcmp` MD5 compare, on-disk cleartext-derived secret). This issue is the user-facing piece.

## Proposed fix

Replace line 1395 with one of:

```c
\"  -A, --password=PASSWORD       set password for rigctld access (single-thread; see SECURITY)\n\"
```

…OR remove the option from the help text and `getopt_long` table entirely until the underlying behavior in `rigctl_parse.c` matches reasonable expectations.

The least-disruptive immediate fix is replacing `(NOT IMPLEMENTED)` with `(best-effort, single-thread only)` and adding one line to `SECURITY.md` pointing at the known limitations.

## Source

- `tests/rigctld.c:1395` — current help text
- `tests/rigctld.c:110, 124, 149, 324–327, 1031` — partial implementation
- `tests/rigctl_parse.c:107, 1820, 5541, 5548` — actual password-check sites (covered in advisory)

## Disclosure / reporter

Reported by **Calm** (Anthropic Claude Opus 4.7 1M ctx, AI agent operated by John Bradley, Credex Research). Contact: `Calm@rainbowsix.dev`. Static source review only.

Cross-ref: GHSA-f72v-7gmh-m9mj, and related deployment-guidance issue.

dchristle:
The password feature was re-enabled with authorization inferred from command argument labels. Commands with no first argument, including `halt`, and VFO-labelled commands could therefore run before authentication. The 32-character secret printed at startup also did not match what the verifier accepted, so the displayed credential could not authenticate.

Use an explicit pre-authentication flag for `password`, `chk_vfo`, and `dump_state`, which existing netrigctl clients need during the opening handshake. Every other command is rejected before it can open, recover, or operate the rig.

Derive one stable 32-character bearer secret at startup, retain that exact value, require an exact-length submission, and compare it without content-dependent early exit. Authentication remains isolated per connection, and the generated credential is no longer written to settings automatically.

Unauthenticated connections now have an absolute 30-second authentication deadline and are closed after five failed password or unauthorized command attempts. Client slots are bounded, socket ownership is coordinated with the authentication watchdog, and daemon shutdown waits for connection workers before releasing shared rig state.

Credential-bearing arguments are redacted from normal diagnostics and startup logging. `rigctltcp` no longer exposes its incomplete `-A` path. The rigctld documentation now describes the supported handshake and makes clear that the reusable bearer secret is sent over plaintext TCP and should be used only on a trusted network or through a secure tunnel.

Validated with a full build, ten repeated focused parser and lifecycle test runs, and `make check` (1/1 C++ tests, 22/22 main tests, and 11/11 streaming tests passed).

Closes #2065.

### DIFF
--- a/doc/man1/rigctl.1
+++ b/doc/man1/rigctl.1
@@ -1635,8 +1635,26 @@ before sending the next command to the radio.
 .
 .TP
 .BR password " \(aq" \fIPassword\fP \(aq
-Sends password to rigctld when rigctld has been secured with -A.  Must use the 32-char shared secret from rigctld.
-(NOT IMPLEMENTED)
+Authenticate to a
+.B rigctld
+server started with
+.BR \-A .
+.B rigctld
+accepts a configuration password containing 1 to 64 bytes and prints a
+32-character bearer secret at startup.
+.I Password
+must be that exact bearer secret.
+Before authentication, the only rig commands accepted by the server are
+.BR \\password ,
+.BR \\chk_vfo ,
+and
+.B \\dump_state
+so clients can complete the opening handshake.  An unauthenticated connection
+must authenticate within 30 seconds of connection establishment and is closed
+after five failed password or unauthorized command attempts.  A failed
+password attempt after successful authentication revokes access and starts a
+new 30-second authentication deadline.  The secret is sent over plaintext TCP;
+use a trusted network or secure tunnel.
 .
 .TP
 .BR set_lock_mode " \(aq" \fILocked\fP \(aq
--- a/doc/man1/rigctld.1
+++ b/doc/man1/rigctld.1
@@ -30,6 +30,7 @@ rigctld \- TCP radio control daemon
 .OP \-t number
 .OP \-C parm=val
 .OP \-X seconds
+.OP \-A password
 .RB [ \-v [ \-Z ]]
 .YS
 .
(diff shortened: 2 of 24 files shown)


N0NB:
@dchristle 

David, there is a merge conflict in `tests/testctlparser.c` that I would rather you examine and fix as needed.

PR #2158, opened 2026-08-17
fix(yaesu): preserve legacy NewCAT bandwidth handling

dchristle:
FT-450, FT-450D, and FTDX9000 do not provide private width tables. NewCAT mode set/get operations dereferenced those missing tables before reaching the existing legacy bandwidth mapping, causing a crash.

Use a shared null-safe selector to preserve the legacy mapping and add socket-backed USB/CW set/get coverage for FT-450 and FTDX9000.

Testing:

- Full build
- `make check TESTS=testnewcatwidth`
- ASan/UBSan focused test

### DIFF
--- a/rigs/yaesu/newcat.c
+++ b/rigs/yaesu/newcat.c
@@ -717,6 +717,41 @@ static int newcat_get_index_from_width(pbwidth_t width, const struct newcat_widt
     return info->count - 1; // Not found, use the maximum
 }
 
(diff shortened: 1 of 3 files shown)


N0NB:
@dchristle 

Hi David.

This PR also has merge conflicts in `rigs/yaesu/newcat.c` and `tests/Makefile.am`.

N0NB:
Thanks, David.

PR #2160, opened 2026-08-18
fix(stream): bound resampler ratios and memory use

dchristle:
Very low requested stream rates can create a large upsampling ratio when Hamlib converts a stream to a backend's native rate. The persistent converter sized its scratch buffers from that ratio, allowing disproportionately large memory allocations.

This change rejects rate pairs outside libsamplerate's supported ratio range, caps each resampler scratch buffer at 16 MiB, omits unsupported effective rates from advertised stream capabilities, and validates the limits before allocating conversion state. It also reports overwrite-ring delivery as fully consumed by the producer. The one-shot resampling API remains unchanged.

The affected streaming allocation path was introduced with the initial streaming subsystem in [#2116](https://github.com/Hamlib/Hamlib/pull/2116). Without these bounds, a client able to request a sufficiently low rate could cause excessive memory use, creating a potential resource-exhaustion or denial-of-service risk for an exposed `rigctld` service.

Tested with the complete test suite both with and without libsamplerate support.

### DIFF
--- a/HAMLIB_STREAMING.md
+++ b/HAMLIB_STREAMING.md
@@ -303,10 +303,12 @@ Limits: `HAMLIB_MAX_STREAM_RATES = 32`,
   effective rates equal the native rates): the native rates, plus the
   curated standard rates {8000, 11025, 16000, 22050, 24000, 44100,
   48000, 96000, 192000}, plus every exact integer division (factors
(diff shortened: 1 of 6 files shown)


mikaelnousiainen:
Will review this tomorrow :)

PR #2161, opened 2026-08-18
fix(usrp): initialize backend state safely

dchristle:
`usrp_init()` left `if_mix_freq` uninitialized and discarded the receiver smart pointer returned by `make()`.

Initialize the backend state with the 45 MHz default, retain receiver ownership through cleanup, and add focused regression coverage.

Tested with a non-USRP build and the existing C/C++ test suites.

### DIFF
--- a/rigs/kit/usrp_impl.cc
+++ b/rigs/kit/usrp_impl.cc
@@ -27,63 +27,55 @@
 #if defined(HAVE_USRP)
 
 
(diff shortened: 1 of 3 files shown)


dforsi:
Thanks David.

I compile-tested this with libusrp 3.4.10 from https://github.com/osmocom/libusrp

Also cherry-picked to Hamlib-4.7

PR #2172, opened 2026-08-24
Enable choosing which stream conversions to allow

mikaelnousiainen:
This PR allows choosing which stream conversions are allowed/disallowed (require_native).
The work is based on feedback from @KJ5HST in https://github.com/Hamlib/Hamlib/pull/2116#issuecomment-5376715109

LLM summary of the changes:

## Summary

Two related changes to the streaming subsystem, both driven by feedback, 
using the format-conversion work in production:

1. `rig_stream_config.require_native` becomes a **per-stage bitmask** instead of
   an all-or-nothing flag, so an application can demand a hardware-native sample
   rate while still accepting the frontend's format conversion.
2. `NONE` becomes the **single empty-list token** across the rigctld protocol —
   conversion-stage lists and capability lines alike — so no key is ever written
   with an empty value.

Nothing has shipped with `require_native` yet, so no compatibility shims are
carried for it.

### 1. `require_native` as a conversion-stage mask

The old boolean covered sample format, rate and channel count together, but those
stages have very different costs to a relay. Resampling is the latency stage, and
demanding a native rate is cheap: read `native_sample_rates` and open at one of
them. Format is the opposite — demanding a native format means taking whatever
each backend happens to offer or `-RIG_ENAVAIL`, and for a relay it is
self-defeating, since doing the conversion itself moves quantization out of the
frontend's routines and into worse ones. The guarantee was only available
all-or-nothing, so consumers negotiated the rate and read
`rig_stream_get_conversions()` afterwards — which lands in the right place but
enforces nothing.

`require_native` now takes the same `RIG_STREAM_CONV_*` constants that
`rig_stream_get_conversions()` reports. Any stage that is both required and
needed refuses the open with `-RIG_ENAVAIL`; stages left out still convert:

```c
cfg->require_native = RIG_STREAM_CONV_RATE;   /* resample = fail, format = fine */
cfg->require_native = RIG_STREAM_CONV_ALL;    /* fully native stream */
cfg->require_native = RIG_STREAM_CONV_NONE;   /* default: convert as needed */
```

`RIG_STREAM_CONV_ALL` is deliberately every bit (`0xffffffffU`), not the OR of
the three defined stages: a stage added in a later release is then forbidden for
a caller that asked for `ALL` without recompiling. The field changed from
`int32_t` to `uint32_t` (layout-neutral) so the value assigns cleanly.

On the wire and command line, `require_native=` takes stage names from the same
vocabulary the `conversions:` response line already speaks:

```
\stream_open AUDIO_RX PCM_S16 48000 require_native=RATE
\stream_open AUDIO_RX PCM_S16 48000 require_native=FORMAT,RATE
\stream_open AUDIO_RX PCM_F32 48000 require_native=ALL
```

Parsing is **strict** here — an unknown stage name fails the open with
`-RIG_EINVAL` rather than being skipped. This is intentionally asymmetric with
the reported-conversions parser, which keeps skipping unknown names for forward
compatibility: a reported stage a client does not know is informational, but a
required stage a server does not know is a guarantee it cannot honour, and
silently dropping it recreates exactly the false assurance this change fixes.

netrigctl forwards the mask by name so the far side enforces the same stages.
`rigstreamtest` gained `--require-native[=LIST]` (bare flag = `ALL`) and now
prints stage names in its conversion report.

### 2. `NONE` as the empty-list token

`stream_conversions_str()` wrote an empty string for a native stream, and the
capability lines wrote empty values for empty lists (`flags=` on an entry with no
capability flags). Both now write `NONE`, from one shared definition:

```
conversions: NONE                                  (was: conversions: )
type=AUDIO_RX ... max_streams=4 flags=NONE ...     (was: flags= )
```

This covers every list-valued key in both list syntaxes — `formats`, `flags`,
`rates`, `channels` and the `native_*` set — so a reader never has to tell "the
empty list" apart from "the field is missing". Applied to the integer lists too:
a rule with an exception for `rates=`/`channels=` would have kept the
inconsistency this change removes. Parsers still accept an empty value as the
empty list (as an absent key already means), so tolerance is read-only.

Wire-visible: `\stream_open`, `\stream_status` and `\stream_list` now report
`conversions: NONE` for native streams, and `\stream_caps`, `dump_caps` and
`dumpstate` write `flags=NONE` where `flags= ` used to appear.

### Documentation

`HAMLIB_STREAMING.md` (conversion model, config struct, open semantics,
`\stream_open` key list, value-syntax grammar), `HAMLIB_STREAMING_BACKEND_GUIDE.md`
(what a mis-declared capability costs a user demanding a native rate),
`README.developer`, and the `rigctld(1)` `require_native`, `conversions` and
`stream_caps` grammar paragraphs.

### Testing

`make check` passes; all 11 streaming suites green. New coverage:

- `test_rigctld_stream`: mask↔text round trip (`ALL` surviving as `ALL`, spaces,
  repeats, short-buffer refusal), strict unknown-name rejection with the mask
  left untouched, `NONE` on the reported side, and a caps line whose every empty
  list renders `NONE` and parses back — including the legacy empty-value form.
- `test_dummy_stream`: per-stage refusals against the F32-native dummy — a format
  demand refuses a format-converted open, a rate+channels demand serves it, and
  with libsamplerate a rate demand refuses a resampled open while a channels
  demand lets it through reporting `FORMAT|RATE`.
- `test_stream_api`: the channel-map stage isolated against a mono-only backend.
- `test_rigctld_commands`: stage names over TCP, `NONE`, and unknown names
  (`CODEC`, `RATE,CODEC`) rejected with the registry left untouched.
- `test_netrigctl_stream`: a per-stage demand crossing the wire end to end —
  `FORMAT` refused, `RATE|CHANNELS` served with `conversions: FORMAT`.

Also verified live against `rigctld -m 1` for every combination above, and the
man page renders clean under groff.

### DIFF
--- a/HAMLIB_STREAMING.md
+++ b/HAMLIB_STREAMING.md
@@ -359,9 +359,15 @@ installed between that source and the stream. Requests beyond the rules
 more channels than the mapping allows) fail with `-RIG_EINVAL`.
 
 `rig_stream_get_conversions()` reports the installed stages
-(`RIG_STREAM_CONV_FORMAT/RATE/CHANNELS`, 0 = native stream). A config
-with `require_native = 1` refuses conversion with `-RIG_ENAVAIL`
-(distinct from `-RIG_EINVAL` for the impossible). One exception to the
+(`RIG_STREAM_CONV_FORMAT/RATE/CHANNELS`, 0 = native stream). The same
+constants go the other way in `require_native`, a bitmask of the stages
+the open must NOT install: any stage both required and needed refuses the
+open with `-RIG_ENAVAIL` (distinct from `-RIG_EINVAL` for the
+impossible), while stages left out of the mask still convert. So
+`require_native = RIG_STREAM_CONV_RATE` demands the hardware's own sample
+rate — the stage that costs latency — and accepts the frontend's format
+conversion, and `RIG_STREAM_CONV_ALL` demands a fully native stream.
+One exception to the
 authoring rule: a relaying backend (netrigctl) fills the `native_*`
 fields too, declaring both views pre-derived; the frontend then serves
 the entry verbatim and delegates conversion to the far side (see §6.1).
@@ -383,8 +389,12 @@ struct rig_stream_config {
     unsigned int        mtu;                /* sender path MTU, 0 = default 1500 (see 6.8) */
     unsigned int        transport_buffer_ms;         /* UDP socket buffer as ms of stream data; 0 = rig token / built-in 250 ms */
     unsigned int        transport_buffer_bytes;      /* explicit UDP socket buffer bytes; 0 = derive from transport_buffer_ms/rate */
-    int                 require_native;     /* 1 = open only as a native stream:
-                                             * -RIG_ENAVAIL rather than convert */
+    uint32_t            require_native;     /* RIG_STREAM_CONV_* stages that must
+                                             * be native: -RIG_ENAVAIL rather
+                                             * than install a listed stage.
+                                             * 0 = convert as needed,
+                                             * RIG_STREAM_CONV_ALL = fully
+                                             * native (see 3.3) */
 };
 ```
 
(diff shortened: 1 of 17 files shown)


mikaelnousiainen:
@KJ5HST Please test this approach - if it's suitable for your use, I'll mark it ready for review.

KJ5HST:
Nailed it. I see no issues at this time. I will incorporate into our pipeline accordingly.

mikaelnousiainen:
@KJ5HST Thanks, this is then ready for merging!

mikaelnousiainen:
@N0NB @dforsi This should be good to merge. Although I think I can merge my own PRs, I'd like someone else do the merges on my PRs :)

mikaelnousiainen:
@N0NB @dforsi Any chance you could check and merge this? It's blocking us moving forward with the Icom and FlexRadio PRs... 🙏

N0NB:
@mikaelnousiainen Merged as requested.  Please let us know the next PR to merge.  I've lost track of the order of everything!

mikaelnousiainen:
@N0NB Thanks! 🙇 I will try to tag you and Daniele once the PRs are ready for merging - and indicating clearly when they've been reviewed (like here). I'll have to do some conflict merges and fixing of build issues on the Icom PR, but that one will be the next - so no immediate action needed.

PR #2177, opened 2026-08-25
ats-mini: fix crashes, BFO offset, FM units; add bandwidth/AF/TS support

red8o:
- Fix NULL pointer dereference when a monitor line has fewer fields than expected (crash)
- Add bounds check when reading a monitor line (buffer overrun)
- Auto-retry once on read timeout / short line to recover from a desynced monitor toggle
- Fix missing BFO offset in SSB frequency calculation
- Fix FM frequency units (10 kHz steps, was treated as 1 kHz)
- Fix bandwidth/step units missing k/M suffix scaling
- Reject set_freq() calls that would require an implicit band change instead of sending them to the radio (was corrupting monitor state)
- Add bandwidth get/set (mode-dependent filter list)
- Add AF (volume) get/set via RIG_LEVEL_AF
- Add tuning step get/set via set_ts/get_ts

Tested with genuine ATS Mini V4, FW 2.35, using rigctld + fldigi.

### DIFF
--- a/rigs/ats-mini/ats.c
+++ b/rigs/ats-mini/ats.c
@@ -1,6 +1,6 @@
 /*
  * Hamlib Mini backend
- * Copyright (c) 2026 by Hamlib Team
+ * Copyright (c) 2026 by Hamlib Team & DG3BP
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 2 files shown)


jonsnow1357:
thanks for the bug fix .... also first time I actually see goto used in real life ...

Issue #1942, opened 2025-11-01
Icom network CAT protocol support


mikaelnousiainen:
This is the main issue for ARDC Hamlib improvement project [Milestone 44 - Icom network CAT protocol support](https://github.com/Hamlib/Hamlib/milestone/27) - plans and discussion. More granular issues will be created as we progress through the milestone.

# Project background

* See ARDC grant: https://www.ardc.net/apply/grants/2025-grants/hamlib-stability-enhancements-and-sdr-transceiver-support/
* See Open Collective page: https://opencollective.com/hamlib-stability-enhancements-and-sdr-transceiver-support

The project backlog is available in: https://github.com/orgs/Hamlib/projects/1 (still in progress)

#  Development Process

* GitHub issues in the main Hamlib repository will track the project's progress, milestones, and deliverables transparently.
* All code contributions will be made as GitHub Pull Requests against the main Hamlib repository for final community review and comments
* All code will be licensed under the existing GNU LGPL 2.1 (or later) license

# Description

To be written after [Milestone 3 - Network audio and I/Q data streaming support](https://github.com/Hamlib/Hamlib/milestone/26) is complete.

mikaelnousiainen:
Icom network protocol backend implementation is now very close to being ready for PR/review. We've tested it on IC-7610, IC-9700 and IC-705 - also with stereo/dual RX and it seems to work very well.

mikaelnousiainen:
Since audio and I/Q streaming is now merged in #2116, we will publish the Icom network backend PR very soon, likely next week.

mikaelnousiainen:
PR open in #2178

PR #2178, opened 2026-08-26
Initial Icom network protocol and network backends implementation based on the Hamlib data streaming subsystem

mikaelnousiainen:
TL;DR: This is (again) a large PR for the initial implementation of the Icom network protocol that the newer Icom rigs use. The implementation uses the Hamlib data streaming subsystem for delivering audio (and I/Q) data over network, merged in #2116. This PR implements #1942. Going for a large and complete PR was a (again) good way to produce something that's a cohesive and a working solution. We've tested this on IC-7610, IC-9700 and briefly on an IC-705 - all working well. What we have not yet done is some extensive, multi-hour streaming tests and it may be some failure scenarios may still need better handling. These can be fixed in follow-up PRs. Anyway, I think this should be ready to merge as it does not really change any existing behaviors - the network-connected rig models use new model numbers!

As a first step, please read the docs: `rigs/icom/ICOM.md` and the updates done to `HAMLIB_STREAMING.md` (design and normative wire protocol) and `HAMLIB_STREAMING_BACKEND_GUIDE.md` (how to add streaming to a backend).

LLM-generated summary of the PR details:

# Icom network backends: CI-V, audio and I/Q over Icom's LAN protocol

This adds **network backends for Icom radios with a built-in LAN server**, the protocol RS-BA1 speaks. CI-V control is tunnelled inside Icom's UDP protocol, and the radio's audio and I/Q streams are carried through the data streaming subsystem merged earlier.

Until now a networked Icom could be reached from Hamlib only by running RS-BA1 (Windows only) or a third-party bridge alongside it, and the audio never arrived through Hamlib at all. With this, `rigctl -m <icom_net_model> -r <radio-ip>` controls the radio directly, and `rig_stream_open()` returns its receive audio — the same API and the same `rigctld` connection as any other backend.

**Target:** 5.0 development. Built on the streaming subsystem already merged; this PR adds no new public streaming API.

## What is in this PR

- **Six network models** — `IC-7610`, `IC-9700`, `IC-705`, `IC-905`, `IC-7760` and `IC-7300MK2`, each a clone of its serial counterpart with the LAN transport and streaming wired in. Capabilities, levels and modes are inherited from the serial model, so they stay in step by construction.
- **The protocol, in three layers** — `network_proto` (pure packet parse/build, no I/O), `network_session` (sockets, handshake, retransmit, liveness) and `icom_network` (the Hamlib backend). CI-V reaches the existing `icom.c` command layer through a seam in `frame.c`, so every CI-V command already implemented works unchanged over the network.
- **Streaming** — `AUDIO_RX`, `AUDIO_TX` and `IQ_RX`, with the wire codecs the protocol defines (LPCM 8/16-bit, mono and stereo, µ-law) decoded to a 16-bit PCM pivot — except the 8-bit LPCM codecs, whose unsigned wire samples *are* `PCM_U8`, so a session that negotiated one serves it with no conversion at all: half the network traffic and no work at either end. Anything else an application asks for is the frontend's conversion.
- **Session-loss reporting** — a small addition to the core: `rig_comm_reason_t` and `rig_strcommreason()` say *why* a link dropped (peer disconnect, link timeout, socket error, auth failure), and a stream whose session has died now returns `-RIG_EIO` rather than looking like an ordinary timeout.
- **A use-after-free fix in the core's teardown order.** `rig_close()` called the backend's `rig_close` and only cleaned up streaming state afterwards, so a backend that releases a per-rig resource there was pulling it out from under its own still-running stream threads. Streams are now torn down first. This is not specific to Icom — it applies to any backend whose stream threads use something the backend owns — and it is one moved block in `src/rig.c`.
- **Configuration** — 15 `net_*` tokens covering credentials, radio selection, codecs, sample rate, jitter buffers, TX framing, liveness and opt-in auto-reconnect.
- **`rigs/icom/ICOM.md`** — documentation for the whole directory: the three transports Icom backends use, every token, the wire-format essentials, how to add a network model, and troubleshooting.
- **Tests** — 99 cases across six suites, driven by an in-process mock radio that speaks the protocol, including a mock that can go silent or announce a disconnect on demand.
- **`rigstreamtest -t reconnect`** — backend-agnostic close/reopen cycling, and `rigstreamtest-hw.sh` gains an opt-in `--tests reconnect`.
- **`rigstreamtest --format`** — request a sample format instead of the float default, so a backend's own path can be exercised on hardware rather than the frontend's conversion of it. `--require-native` already existed but could not succeed against any backend that does not serve float; it works now. `--list-streams` also reports `formats=`, and `rigstreamtest-hw.sh` gains an opt-in `--tests native` that opens every declared format and demands it arrive unconverted. Both are backend-agnostic.

Roughly 72 files. About a third of the added lines are tests and documentation. The `ic7300.c` file is split so `IC-705`, `IC-905` and `IC-9700` get their own files, which is most of the deletions; no behaviour changes for the serial models.

## Main features

- **CI-V unchanged over a new transport** — the network session is a transport, not a second command layer. Spectrum scope, transceive frames and every existing CI-V command work over LAN because nothing above the seam knows the difference.
- **Capabilities describe the session, not just the model** — the radio negotiates one codec and one sample rate per connection, so `dump_caps` reports what the *model* can do while `\stream_caps` reports what *this connection* can carry. An application can discover the real geometry before opening anything, which matters because rate conversion needs libsamplerate and many builds do not have it.
- **Stereo receive** — a dual-receiver radio carries both receivers in one stream, one per channel, with the second receiver's audio on the right channel.
- **I/Q** — the stereo wire codec carries `{I,Q}` pairs as a single complex channel.
- **Loss handling** — sequence tracking with retransmit requests, and gaps reported to the streaming subsystem's loss accounting rather than silently papered over.
- **Liveness and reconnect** — a configurable timeout notices a radio that has gone away, and opt-in `net_auto_reconnect` re-establishes the session in the background. Control recovers by itself; open streams are ended and must be reopened, because the new session has fresh sequence numbers.

## Trying it out

Needs a networked Icom and its **network user** credentials (set in the radio's own menu, not a Hamlib account). Pick the model number for your radio:

| Model | Radio | Hardware-tested |
|---|---|---|
| 3095 | IC-7610 (Network) | yes, including transmit |
| 3096 | IC-9700 (Network) | yes, including transmit |
| 3097 | IC-705 (Network) | yes, including transmit |
| 3098 | IC-905 (Network) | no |
| 3099 | IC-7760 (Network) | no |
| 3100 | IC-7300MK2 (Network) | no |

All six offer `AUDIO_RX`, `AUDIO_TX` and `IQ_RX`. The examples below use `$MODEL` — substitute a number from the table.

```sh
MODEL=3095            # IC-7610; see the table above

# control
rigctl -m $MODEL -r 192.168.1.50 -C net_username=USER,net_password=PASS

# receive audio to a WAV file
rigstreamtest -m $MODEL -r 192.168.1.50 -C net_username=USER,net_password=PASS \
    -t audio_rx -d 10 -w capture.wav

# every advertised mode in one go, with recordings
tests/rigstreamtest-hw.sh -m $MODEL -r 192.168.1.50 \
    -C net_username=USER,net_password=PASS
```

`rigs/icom/ICOM.md` §2–§3 covers connecting and every token; §5 is the wire format; §8 is troubleshooting.

No hardware is needed to review the protocol handling: the six test suites run the whole handshake, streaming and session-loss paths against the in-process mock.

## Known limitations

- **Three of the six models are untested on hardware.** `IC-7610`, `IC-9700` and `IC-705` are hardware-validated including transmit; `IC-905`, `IC-7760` and `IC-7300MK2` are written from the shared implementation and marked `Untested` in their model status. They are expected to work; nobody has shown that they do.
- **Auto-reconnect is verified against the mock, not on hardware.** Provoking it on a real radio needs an unplanned loss, which means physically removing it. The hardware reconnect test covers deliberate close/reopen, which is a different path.
- **ADPCM is defined by the protocol but not implemented**, and is rejected at `set_conf` rather than half-working.
- **The codec and rate tokens take a choice index, not a value.** `net_sample_rate=48000` is an index far outside the list and is refused; this trips people up, and is called out in the documentation.
- **Some codec/rate pairings carry no audio, and the radio gives no hint.** On an IC-7610 and an IC-9700, one rate per codec produces silence: the connection is accepted, an audio port is assigned, no error is reported and no packet ever arrives. All 12 cases — of 60 combinations measured across the two radios — are exactly those where the wire data rate is 48000 bytes/second. The radio advertises every rate involved in its own capability bitmap, and advertises nothing at all about codecs, so this is not something the backend can negotiate around. The cause is not established; it is documented in `ICOM.md` with the full matrix.
- **I/Q needs the two-channel codec** (`net_rx_codec=3`), so it is absent from a session negotiated with a mono codec rather than being offered and then failing.
- **One radio, one session.** The protocol allows a single client, so a second connection is refused while the first holds the slot — including briefly after a disconnect.

## Testing

`make check` passes: 17 suites, of which six are new here (99 cases) covering packet parse/build, the sequence buffer, the session state machine and handshake, configuration tokens, stream open/close, and session loss and recovery — all against the mock radio, so they run with no hardware in CI.

Hardware validation on an **IC-7610** and an **IC-9700**: the full `rigstreamtest-hw.sh` sweep including transmit at 1% into a dummy load, plus stereo-codec I/Q runs measured against the expected sample rate, and reconnect cycling that closes the rig both tidily and with a stream still running. No gaps, overruns or underruns. `rigctl --dump-caps` reports zero declaration warnings for all six models.

The 8-bit codecs' native path is verified on an IC-7610 with `--tests native`: over the 8-bit stereo codec, `PCM_U8` opens with `conversions=0x0` and carries 297600 bytes where `PCM_S16` carries 618240 for the same audio — the halving, measured rather than asserted. Over the µ-law codec the same case correctly reports that U8 would need a format conversion.

Every codec was measured against every sample rate on both radios — 60 combinations — to establish which pairings actually carry audio rather than trusting what the protocol defines. The decoded output was checked against the 16-bit codecs on the same band, so "bytes arrived" is not mistaken for "bytes are right". That sweep is what turned up both the 48000 bytes/second dead cells and a decode bug in the 8-bit codecs, which had never been exercised: their unsigned wire bytes were being read as 16-bit samples.

### Test findings

There are two _potential_ pre-existing flakes in `test_rigctld_commands`, both timing-sensitive, neither from this work and neither in code this PR touches:

- `rx_mute_zeros` fails when `subscribe_and_await_ack()` times out and the stream is auto-closed under it, taking five downstream assertions with it. Reproduced once in 24 runs at eight-way contention, which is what a parallel `make check` looks like.
- `rx_codec_passthrough_audio` asserts that the `codec_frames` counter is at least the number of datagrams observed. Under a Wine cross-build it failed 2 runs in 5 on an unchanged binary. It passes 10 of 10 natively.

Both were found by running the suites under contention and under the Windows cross build; neither reproduces in an ordinary serial run. These may need further investigation.

## Use of LLMs

LLMs were used as an assistant in producing this work.

The Hamlib backend architecture/implementation and the hardware debugging are author's work. LLM assistance was used more heavily for the tests and the documentation, and for routine or mechanical work throughout.

All of it has been reviewed, built and tested by the author, and validated against real radios.


mikaelnousiainen:
FYI @KJ5HST @jfrancis42 - PR for the first backends to use the streaming subsystem.

mikaelnousiainen:
The protocol implementation is based on the open-source kappanhang project and local testing/debugging with packet capture and AI-assisted analysis of the wire protocol.

KJ5HST:
Oh, man, I've been considering a MK2, and thought this might force me to get one ... but you did a good job saving me money with your design choices. :-)

I was able to build this and ran it against our bridge again. Everything built clean and ran like before with no errors. Here are my observations...

Your `test_rigctld_commands` reproduced — as a third case. 24 runs at 8-way contention: 1 failure, and it is neither of your two — `rx_counter_payload_audio`, first packet never arrives (`test_rigctld_commands.c:2040: n >= RIG_STREAM_HEADER_SIZE`, `packet 0: timeout`, then `only got 0 data packets, expected >= 3` at :2108). This is similar to your `rx_mute_zeros` (a subscribe/first-delivery deadline missed under scheduler load), same 1-in-24 rate, different test. Three distinct receive tests failing the same way at the same rate under the same condition looks like one root cause — the feeder/subscribe startup path racing a fixed deadline under contention — not as independent bugs.

One thing I think you should consider before merging: Unrecoverable losses never reach the subsystem's loss accounting. You noted that gaps are "reported to the streaming subsystem's loss accounting rather than silently papered over", but I can't find it. `stream_skip_samples()` is the API `src/stream.h` documents as backend-facing for exactly this (`gap_count`: "Radio-side packet gaps detected by backend") — and nothing under `rigs/` calls it; this backend would be its first caller. What actually happens when a missing sequence exhausts `ICOM_NETWORK_RETRANSMIT_MAX` or the tracker resyncs past `ICOM_NETWORK_MISSING_FLUSH`: the entry is dropped (`rxtrack_remove_at` / `rxtrack_reset`), `audio_resyncs` increments — a session-private counter I don't see anywhere — and the samples simply never arrive at `stream_backend_write`. On my end that means 3 things:

1. `rig_stream_get_stats()` reads `gaps=0, link_loss=0` on a lossy link. My bridge prints exactly those fields as its health line, so against an Icom network rig it would report a clean stream while the audio thins.
2. No index hole is inserted, so the loss is invisible to the sample-index timeline too — indistinguishable from a producer that paces slower.
3. Delivered-rate-vs-expected becomes the only detector left, and it can't tell loss from pacing.

The session exposes its resync counters through a getter (`network_session.c:1085`), and the RX thread that owns the stream is the natural place to translate. At give-up the missing count is known and the per-packet sample geometry is fixed for the session, so `stream_skip_samples(stream, n, RIG_STREAM_DROP_GAP)` can be sized; at a resync where the span is unknowable, passing 0 lands in the `gaps_unknown` path that exists for exactly that. I can easily re-run my tests if you do something like this.

A couple less significant observations:

- `stream_ringbuf_wait_data_locked()` wakes on `failed` only in the infinite-wait branch. The timed branch (`stream_ringbuf.c:285`) still loops on `closing` alone, so a reader blocked in a timed wait when the session dies ignores your broadcast, sleeps to its full deadline, and books an `underrun_count++` for a dead session before `rig_stream_read()` maps the result to `-RIG_EIO`. Correctness survives (the EIO still comes back; the next call hits the early check), but the wake is wired to half the wait paths and the spurious underrun lands in the same stats a session death makes interesting. The fix mirrors the `closing` handling: bail out of the timed loop on `failed` without counting an underrun.
- With both `closing` and `failed` set — reachable, since loss ends streams and the app then closes them — the two paths in `rig_stream_read()` disagree: the pre-wait check answers `-RIG_EIO` (`failed` wins), the post-wait check answers `-RIG_ENAVAIL` (`closing` wins). Either is fine, but the same answer from both would be better. I'd let `closing` win in both places — a deliberate local close is the app's own action and shouldn't morph into an I/O error report.

Now for the stuff you fixed for me:

- **`-RIG_EIO` for a dead session.** My bridge runs unattended at the radio site. Its RX worker treats `-RIG_ETIMEOUT` as "no data this tick", so until now a rig that died mid-session read as an idle stream forever — a live process, a healthy console, no audio moving. Against a backend that sets the failed flag, the same unchanged worker exits non-zero on the first read after the loss and the supervisor restarts it. I changed no code: the error branch I already had is the EIO path. That's what a well-shaped error contract addition looks like from the consumer side — existing consumers get the improvement for free.
- **The `rig_close()` teardown reorder.** My bridge closes its streams explicitly before `rig_close` so it never hit the window, but every consumer that relied on `rig_close` alone did. One doc line worth adding to the backend guide: since streams are now torn down first and `stream_state` is NULLed, a backend's `rig_close` must not touch streaming state.
- **The slot-hold honesty** ("the early attempts are expected to fail") — that detail feeds straight into supervisor restart policy for consumers like mine; likewise the documented liveness floor.

...and now a question:

Regarding `comm_reason` in chained topologies... The multicast `statusReason` and the `HAMLIB_STATE()` guidance make the reason readable locally, and you already note the rigctld text protocol can't carry it. The case I care about is the chain: a netrigctl consumer of a rigctld that fronts an Icom network model. When the server-side backend read fails with `-RIG_EIO`, does that propagate to the remote `\stream_read` consumer, or does the outer hop decay it back to an ordinary timeout? If it decays, the loss report dies two layers from the operator. I can measure this against the mock myself when I get time, but thought I'd pick your brain on it.

Finally, this branch predates the per-stage `require_native` mask (my build probe confirmed: `RIG_STREAM_CONV_ALL` absent). Whichever PR merges second inherits a semantic merge, not a textual one: under the mask, the "no conversion at all" your `--tests native` demands is `RIG_STREAM_CONV_ALL`, while a literal `require_native=1` becomes "refuse format conversion only". The intent survives either way; the flag plumbing and help text need one deliberate pass. Happy to adjust either way, since the mask semantics were my request.

The three-layer split (proto/session/backend) made this reviewable from the client side in a way a it otherwise would not have, and the mock-driven suites meant everything above except the hardware claims was checkable on a laptop. Nice work — again. It's cool to be able to confirm all this without having to go buy that MK2 ... then again, maybe...

mikaelnousiainen:
@KJ5HST Thanks for the detailed review, I will get back to your feedback in a couple of days! I'll likely add some commits to this PR to fix any pending issues or if some larger architectural changes still are needed. Minor changes can be addressed in follow-ups.

rigs/icom/network_session.c
```diff
@@ -0,0 +1,2148 @@
...
+    }
+
+    icom_network_session_teardown(s);
+}
+
+void icom_network_session_free(struct icom_network_session *s)
+{
+    if (s == NULL) { return; }
+
+    if (s->connected) { icom_network_session_disconnect(s); }
```
dchristle, line 2135: Could we stop and join any running workers here regardless of `connected`? That flag is false during reconnect and can also be false after partial thread startup. I'm concerned that closing in either case can free the session while a worker still has a reference to it.

rigs/icom/network_session.c
```diff
@@ -0,0 +1,2148 @@
...
+}
+
+/* Create a UDP socket connected to host:port. */
+static int icom_network_session_socket_open(struct icom_network_session_socket
+        *s, const char *host, uint16_t port)
+{
+    struct sockaddr_in sa;
+    int fd;
+
+    fd = socket(AF_INET, SOCK_DGRAM, 0);
```
dchristle, line 239: Where does Winsock get initialized for an application using Hamlib directly? The custom transport seems to bypass the usual initialization. Both the CLI and the mock do their own `WSAStartup()`, so those can pass while a fresh library consumer still fails at the first `socket()` call.

src/stream.c
```diff
@@ -1827,11 +1827,12 @@ int HAMLIB_API rig_stream_read(RIG *rig,
 
     pthread_mutex_lock(&rb->lock);
 
-    if (rb->closing)
+    if (rb->closing || rb->failed)
```
dchristle, line 1830: I think the new `-RIG_EIO` result needs a corresponding change in the rigctld feeder. It currently ignores the error and immediately reads again, so the 100 ms timeout no longer slows the loop down once the source has failed. Could we stop the feeder on a terminal error?

rigs/icom/network_session.c
```diff
@@ -0,0 +1,2148 @@
...
+
+    if (icom_network_packet_parse_audio(buf, n, &au) != RIG_OK)
+    {
+        return;
+    }
+
+    sequence = icom_network_get_le16(buf + ICOM_NETWORK_OFF_SEQUENCE);
+
+    int resync = icom_network_rxtrack_observe(&s->audio.rxtrack, sequence,
+                 icom_network_now_ms());
```
dchristle, line 933: How does a recovered audio packet get put back into sample order? As far as I can see, the tracker updates its missing set, but the payload still goes straight into the FIFO. That seems to let a retransmission arrive after the audio that followed it, and duplicates would get played again too.

rigs/icom/network_session.c
```diff
@@ -0,0 +1,2148 @@
...
+    {
+        if (icom_network_txbuf_add(&s->civ.txbuf, sequence, packet,
+                                   packet_length, icom_network_now_ms()) != 0)
+        {
+            rig_debug(RIG_DEBUG_WARN,
+                      "%s: CI-V sequence %u not retained for retransmit\n",
+                      __func__, (unsigned)sequence);
+        }
+
+        (void)send(s->civ.fd, packet, packet_length, 0);
```
dchristle, line 1714: Could we check the result of `send()` here? At the moment, successful packet construction is enough to return success even if the socket write fails. The receive loops also skip negative results, and traffic on the control socket can keep the session looking healthy while the audio or CI-V socket is failing.

rigs/icom/network_seqbuf.c
```diff
@@ -0,0 +1,238 @@
...
+            i++;
+            continue;
+        }
+
+        if (m->retries >= ICOM_NETWORK_RETRANSMIT_MAX)
+        {
+            /* Give up on this one. */
+            rxtrack_remove_at(rt, i);
+            continue; /* index i now holds a different entry */
+        }
```
dchristle, line 229: I agree that reporting the gaps would be useful. I'm less sure the resync counter is enough to drive it: exhausting retries doesn't increment that counter, and the packet payloads aren't all the same size. Could we carry a loss event through to delivery and report an unknown size when we can't establish the missing sample count?

It seems like after a discard for a full queue, the backend passes received bytes to `stream_backend_write()` without reporting the missing interval & no Icom path calls `stream_skip_samples()`. (I'm looking at icom_network.c:465-477 here too)

src/stream_ringbuf.c
```diff
@@ -261,12 +261,12 @@ int stream_ringbuf_wait_data_locked(struct rig_stream_ringbuf *rb,
     /* Block indefinitely: wait until data arrives or the ring is closed. */
     if (timeout_ms < 0)
     {
-        while (rb->count == 0 && !rb->closing)
+        while (rb->count == 0 && !rb->closing && !rb->failed)
```
dchristle, line 264: The indefinite wait checks `rb->failed`, but the timed wait checks only an empty ring and `closing`. When the Icom worker marks failure and broadcasts, an already-blocked timed reader wakes and waits again until its original deadline. After previously receiving data, it also increments the underrun counter. Only then does `rig_stream_read()` translate the result to `-RIG_EIO`.

Could the timed wait check `failed` as well?

rigs/icom/network_proto.c
```diff
@@ -0,0 +1,843 @@
...
+    }
+
+    if (bufsize < ICOM_NETWORK_CIV_LEN + civ_length)
+    {
+        return -RIG_EINVAL;
+    }
+
+    icom_network_packet_build_header(buf, bufsize, (uint32_t)(0x15 + civ_length),
+                                     ICOM_NETWORK_PACKET_TYPE_DATA, sequence,
+                                     local_id, remote_id);
```
dchristle, line 382: Could we use the socket role or message structure to distinguish these packets? A variable-length CI-V or audio packet can have the same total length as one of the management packets listed here. The length check wins before the data checks, so a valid response or audio packet appears to get dropped.

dchristle (changes requested):

mikaelnousiainen:
Thanks for all the review comments to both of you @KJ5HST and @dchristle - I'm working on commits to address these. There were some good findings from both reviews - and they do cause some larger changes, but take the (Icom) stream reliability and stability to a significantly better level!

I'm almost done with the changes and hope to push new commits today or tomorrow. If you can - please have a look at the changes -> if we could wrap up this PR and merge it afterwards. I'd like to get the features out for testing on a larger audience!

mikaelnousiainen:
One more thing, I really wish to get #2172 merged before this PR to get the `require_native` changes fully addressed - there are already some small conflicts between this PR and 2172.

rigs/icom/network_session.c
mikaelnousiainen: 👍 will fix this, disconnection + join/stop should not be conditional!

rigs/icom/network_session.c
mikaelnousiainen: 👍 Adding (refcounted) startup/cleanup pairs to session allocation

mikaelnousiainen:
@KJ5HST @dchristle Thanks for a careful pass and for confirming the EIO contract, the `rig_close()` teardown order and the slot-hold note from the consumer side. Everything you raised is considered or fully fixed on the branch. The per-thread replies to your inline comments have the details. This is the summary, including what the inline round turned up afterwards.

Note: I've noticed there are already some build failures showing up, will be fixing those!

LLM summary follows:

### Losses now reach the loss accounting

You were right that nothing reported them, but accounting at give-up alone
would not have been enough: the audio path wrote packets in arrival order, so a
retransmit landed after newer audio, a duplicate played twice, and a packet
arriving after we gave up was both lost and delivered.

- **Order first.** A new shared component, `src/stream_reorder.[ch]`, releases
  datagrams in sequence order with a bounded hold window. Pure logic, its own
  unit tests, meant to be reused by other sequence-numbered transports.
- **`net_rx_latency` is that window** (0–1000 ms, **default 0**); it used to be
  documented as a jitter buffer that no code implemented. At 0 a missing packet
  is given up as soon as the next arrives, late packets are dropped rather than
  played out of order, and no retransmits are requested, since a resend could
  never be used. Above 0 the missing packets are requested, and the request is
  repeated every half window while it lasts.
- **Every loss is reported before the payload that follows it.** Audio is
  concealed with format-correct silence through a new frontend helper,
  `stream_fill_gap()` (`RIG_STREAM_DROP_GAP | RIG_STREAM_DROP_CONCEALED`, new
  `concealed_samples_gap`); I/Q gets a sized `rig_stream_mark_gap()` hole, as
  you suggested; a jump too long to size is an unsized gap; and the session's
  own receive queue overflowing, which was silent too, is reported as a local
  overrun.
- **Sizing:** the radio splits each 20 ms frame into packets of at most 1364
  bytes, so a lost packet is sized from the last packet seen at the same
  position in the frame. I confirmed that cycle on an IC-7610 for every
  codec/rate combination that carries audio.
- **Also fixed:** ADPCM is reset at a loss, and time anchors use the backend's
  own sample position (they were rescaled twice under rate conversion).

### Failure signalling: the timed wait, closing vs failed, and the send side

All as you suggested. The timed wait wakes on `failed` and books no underrun
for a dead or closing ring; one helper maps ring state to the read result on
both paths, with `closing` winning everywhere; a failed stream hands out what it
already holds and then returns `-RIG_EIO`, so nothing received before the
failure is thrown away. The send side gets the same contract, where writes used
to keep "succeeding" into the ring of a dead session. Backends signal through
`stream_mark_failed(stream, reason)` and the reason is in the new
`rig_stream_stats.fail_reason` (`\stream_status` prints it).

### EIO through rigctld to netrigctl

It decayed, and worse than a timeout: the rigctld RX feeder only handled
`RIG_OK`, so on `-RIG_EIO` it busy-spun at 100% CPU while still answering
PINGs, and the remote client read `-RIG_ETIMEOUT` for ever. The reserved
`RIG_STREAM_CTRL_ERROR` frame now carries a 12-byte block (version, Hamlib
error code, `comm_reason`) with `seq = 0`; older clients already drop
header-only ERROR frames, so it is backward compatible. rigctld sends it when a
backend read or write fails, stops touching the backend, and re-sends it on
later PINGs/SUBSCRIBEs until the stream is closed or the keepalive expires. The
netrigctl client marks its own stream failed with that reason, so an unchanged
worker sees the same drain-then-`-RIG_EIO` behind rigctld as locally. The test
checks CPU time and fails with the fix disabled ("993 ms CPU in 1 s idle").

### Backend guide

Added the `rig_close` note: streams are closed and `stream_state` NULLed before
the backend's `rig_close` runs, so per-stream resources belong in
`stream_close`. New sections cover loss reporting (fill vs mark, local overrun,
codec reset), the reorder window and source failure.

### Flaky receive tests under 8-way contention

Agreed it looks like one cause, but your failure is worse than a missed
deadline: no SUBSCRIBE_ACK for three attempts and then nothing for another 2 s
means the server was silent for 4 s or more, which is either a starved feeder
thread or datagrams reaching the wrong socket (the server binds dual-stack
`[::]:0`, the test client `127.0.0.1:0`). It predates this branch, so the fix
gets its own PR; for now a failing receive prints the server-side stream state
(`diag:` line), so your next failure says which of the two it is.

### `require_native` mask

Agreed, and I'd rather not wait on either branch: they conflict only in three
option hunks in `tests/rigstreamtest.c`. The one real semantic issue was the
hardware script parsing the library's refusal log text, which your branch
rewrites; rigstreamtest now prints its own stable line
(`refused: native stream required, needs=FORMAT,CHANNELS`) and the script
matches that. The Icom docs and mock tests for `=RATE` follow once the mask
lands.

### From your inline comments

- **Freeing a session with workers running** was a use-after-free during a
  reconnect: the reconnect thread clears `connected` while it works, so `free()`
  skipped the join. It now always disconnects and joins, a close cuts a running
  handshake short, and a lock keeps application calls off sockets that the
  reconnect is replacing.
- **Winsock** was never started for a plain library consumer, since these models
  use `RIG_PORT_CUSTOM` and never go through `network_open()`. The session now
  holds its own reference.
- **`send()` results** are returned (`-RIG_EIO`), receive errors are recorded
  instead of skipped, and each socket has its own liveness clock, so the control
  socket's keepalives can no longer make a dead CI-V or audio socket look
  healthy. A socket that reported an error before going silent is reported as
  `SOCKET_ERROR` rather than `LINK_TIMEOUT`; it never ends a session sooner than
  the user's `net_liveness_timeout` allows.
- **Classification by length** dropped valid data: a CI-V frame of 43, 59, 75,
  107 or 123 bytes has the same total length as a management packet. Packets are
  now classified by the socket they arrived on, each checked against its own
  structure.
- **CI-V tracking.** Checking the CI-V path after your audio comment turned up
  three faults: duplicate frames were delivered twice (a repeated ACK can be
  taken as the answer to the next command); the radio's idles share the CI-V
  sequence, so tracking only the frames saw a false gap for each of the ~30
  idles a second and kept asking the radio to resend them; and a stream that
  starts counting again was mistaken for old traffic.
- **Two bugs found while testing this:** a `recv()` could block for ever after
  another thread consumed the socket's pending error, with teardown waiting on
  that thread (every `recv()` is bounded now); and a lost handshake reply was
  still unrecoverable when it was the radio's *first* tracked packet, because
  the radio numbers a stream's packets from 0 and tracking ignored 0.

### Winsock reference counting in the frontend (Windows)

Chasing the Winsock question above turned up a related fault in `src/network.c`
and `src/multicast.c`, which I've fixed in the same branch. It is unrelated to
streaming, so it is a separate commit and easy to split out.

Winsock counts references: every successful `WSAStartup()` needs its own
`WSACleanup()`, and releasing the last one stops **every** socket in the
process. Today `network_init()` starts Winsock once behind a `static int`
guard, for all three of its users (a port, the multicast publisher, the
multicast receiver), and `network_close()` then releases that single reference
whenever any one port closes. So closing one network rig could cancel the
sockets of another rig, of the multicast threads, or of the application itself
— and since the guard is cleared at the same time, a later open starts the
cycle again. `src/multicast.c` had its own imbalance: `multicast_init()` takes
a reference, but it was released at the end of the publisher thread, which is
`#if 0`-ed out and no longer runs, so that one was never given back.

Now each socket takes its own reference and gives back exactly that one:
`network_init()` per socket, `network_deinit()` when it closes, including the
failure paths of `network_open()` and of both multicast start functions, and
the multicast stop functions release what their start took.
`multicast_init()`'s reference is released in `multicast_stop()` instead of in
the thread. Behaviour outside Windows is unchanged (both helpers compile to
nothing), and a Hamlib consumer's own Winsock references are now untouchable by
the library.

Two notes: `multicast_stop()` also dereferenced `rig_state.multicast` after
testing it for NULL, which would crash when stopping something that was never
started, so that check is now before the dereference. And I can only
cross-compile this here, so it needs a run on Windows.

### Hardware

IC-7610 and IC-9700 over the LAN, transmitting at 1% into dummy loads:

- **Sweeps:** audio, stereo audio and I/Q (IC-9700 on 2 m/70 cm), loopback,
  transmit and full duplex pass with `gaps=0` and `concealed=0`.
- **Soaks:** 3-minute full-duplex runs on both radios (3 × 15 s transmit with
  45 s receive) and a 5-minute run on the IC-7610: no gaps, overruns, late
  transmit frames or false liveness alarms. The only blemish is one transmit
  underrun that the test tool books at shutdown, after it stops its own writer
  while the backend is still draining.
- **rigctld → netrigctl chain:** clean on both. The IC-9700 first needed a fix
  outside streaming: the range parser in `rigs/dummy/dummy_common.c` read the
  `vfo`/`ant` masks as signed and rejected `RIG_ANT_CURR` (`0x80000000`), which
  breaks any such rig behind rigctld on master too.
- **Sanitizers:** the Icom and streaming suites are clean under
  AddressSanitizer, and ThreadSanitizer reports no races in this code.

For a deliberately lossy link there is a new manual script,
`tests/rigstreamtest-hw-loss.sh` (dummynet on macOS, nftables or iptables on
Linux). It sweeps loss rates against reorder windows and checks the silence,
rigctld-chain and repeated-open cases against a real radio. Measured on an
IC-7610 with 5% of the radio's packets dropped, per 20 s run of about 2000
packets:

| Window | 0 ms | 20 ms | 50 ms | 100 ms | 200 ms |
|---|---|---|---|---|---|
| Packets still lost | 79 | 1 | 0 | 0 | 0 |

At 0 ms every loss is reported and no retransmits are requested; with a window
what remains is a packet whose resends were lost as well. The silence case ends
the stream with `-RIG_EIO` / `LINK_TIMEOUT` about 5 s after the radio goes
quiet, directly and through rigctld, and 20 of 20 opens succeed at 5% loss
(average 960 ms against about 800 ms on a clean link).

### Still to come

The netrigctl client has the same arrival-order problem as the Icom audio path
did — a reordered datagram rewinds its expected sequence and is double-counted
as link loss — so it adopts the reorder window next, possibly as its own PR.
The flaky-test fix and the Icom `require_native` docs and tests follow
separately, and the `RIG_ANT_CURR` parser fix is worth its own small upstream PR
since it is a master bug.

src/stream.c
mikaelnousiainen: Agreed, and it was worse than a slow loop: the RX feeder only handled `RIG_OK`, so on `-RIG_EIO` it spun at 100% CPU while still answering PINGs, and the client read `-RIG_ETIMEOUT` forever.

rigs/icom/network_session.c
mikaelnousiainen: You're right that this was not handled correctly. LLM summary of the changes follows:

It didn't: a retransmission was written after the audio that followed it, and a
duplicate played twice. Audio now goes through a shared reorder window
(`src/stream_reorder.[ch]`): released in sequence order, a missing packet waited
for at most `net_rx_latency` ms (default 0), late packets and duplicates
dropped rather than played out of order, and retransmits requested only when the
window is above zero. On an IC-7610 with 5% loss injected, a 50 ms window loses
nothing over 20 s.

Checking the CI-V path the same way turned up three more faults:

- **Duplicates** were delivered twice. The icom layer doesn't match an ACK or
  NAK to its command, so a repeated ACK could be taken as the answer to the next
  one. A frame whose exact sequence already arrived is now dropped
  (`civ_duplicate_dropped`). Reordering isn't needed there: transactions are
  serial.
- **Idles.** The radio numbers its idles in the same sequence as its CI-V
  frames — about 30 a second on an IC-7610 — and we tracked only the frames, so
  each idle looked like a lost packet: we kept asking the radio to resend idles,
  and the tracker resynced whenever idles outnumbered its window (the "CI-V
  sequence resync" warning at every close). Every numbered packet is tracked
  now, on the control socket too, which also gains retransmit requests
  (`civ_idles_not_requested`). The warning is gone on both radios.
- **Restarts.** A reopened stream counts from zero again. A packet from behind
  the newest one that never arrived is delivered rather than taken for a
  duplicate, and a run of them restarts the tracking. The audio window resyncs
  the same way, instead of first dropping up to 256 packets — seen on an
  IC-9700, where a new session received packets left over from the previous
  audio stream.

The same tracking change fixed a handshake failure under loss: the radio numbers
a stream's tracked packets from 0, and tracking ignored 0, so a lost login
response or capabilities reply left no gap to notice and the connect gave up
after 10 s. Tracking now expects 0 (`handshake_recovers_lost_first_reply`), and
20 of 20 opens succeed at 5% loss.

rigs/icom/network_session.c
mikaelnousiainen: Yes on both counts.

Summary:

- **Send results:** every send goes through one helper that records failures
  against its socket. `icom_network_civ_send()` and `icom_network_audio_send()`
  return `-RIG_EIO` when the send fails, and also when the socket has failed
  since its last packet — on a connected UDP socket the kernel reports the radio
  refusing a packet on whichever call comes next, usually the receive thread's.
- **Receive errors** are recorded the same way, with a short pause so a socket
  that fails immediately can't spin its thread.
- **Per-socket health:** each socket has its own liveness clock (audio only
  while a stream runs), so the control socket's keepalives can no longer make a
  dead CI-V or audio socket look healthy. Silence for `net_liveness_timeout`
  ends the session, reported as `SOCKET_ERROR` when that socket had an error
  since its last packet and `LINK_TIMEOUT` otherwise. Errors never end it sooner
  than the user's timeout, and `0` still means never give up. The radio sends
  every socket steady traffic — measured on both radios at about 23 packets a
  second on control, 46 on CI-V and 115 on audio — so silence is real; 3- and
  5-minute transmit soaks raise no false alarm.

`civ_silence_is_link_timeout` silences only the CI-V socket (before, the session
stayed valid); `civ_port_refusing_is_socket_error` closes the mock's CI-V port
and expects `-RIG_EIO` from a send and `SOCKET_ERROR` as the reason;
`socket_errors_respect_liveness_disabled` checks that a timeout of 0 keeps the
session.

Writing those found a hang: `select()` can report a socket readable because of
a pending ICMP error that a `send()` on another thread consumes first, and the
following `recv()` then blocked for good with teardown waiting on that thread.
Every `recv()` is bounded by `SO_RCVTIMEO` now.

rigs/icom/network_seqbuf.c
mikaelnousiainen: Agreed on both points. The new reordering implementation handles this.

Summary:

Audio no longer uses this tracker: the reorder window reports each gap when it gives up on it, and the gap travels with the payload that follows, so it reaches the stream before those bytes are written. Audio is
concealed through `stream_fill_gap()` (`RIG_STREAM_DROP_GAP | RIG_STREAM_DROP_CONCEALED`), I/Q gets a sized `rig_stream_mark_gap()` hole, and a gap whose size can't be established is reported as unsized, as you suggested. Sizing comes from the last packet seen at the same position in the radio's 20 ms frame (split at 1364 bytes, e.g. 1364 + 556 for 16-bit mono at 48 kHz).

The queue discard at `icom_network.c:465-477` is reported too: the dropped bytes are carried to the next payload and reported as a local overrun (`overruns`, `concealed_samples_overrun`) before it is written. Tests:
`audio_window_gives_up`, `audio_queue_overflow_reported`, `rx_audio_gap_concealed`, `rx_iq_gap_marked`, `rx_audio_restart_unsized`.

The tracker still serves CI-V, where a reply given up on already surfaces as a
command timeout.

src/stream_ringbuf.c
mikaelnousiainen: Fixed as you suggested.

Summary:

The timed wait wakes on `failed`, returns at once for a failed or closing ring and books no underrun for either. `rig_stream_read()` hands out whatever is still buffered and then returns `-RIG_EIO`, and `closing` wins over `failed`. Tests: `stream_ringbuf_failed_wakes_timed_wait`, `failed_wakes_timed_reader`, `failed_read_drains_then_eio`, `closing_wins_over_failed`.

rigs/icom/network_proto.c
mikaelnousiainen: Another good catch, classification is now done by socket role.

mikaelnousiainen:
Waiting for your final comments @KJ5HST @dchristle - I feel the PR is now in a good shape and the retransmit/loss infrastructure made reliability a lot better (there's the loss test that can be executed against a real rig).

The final build failure is odd - no tests fail, but the process exits with an error status, still fixing that.

KJ5HST:
> Waiting for your final comments

I have nothing more to add. I think it needs broader review and testing by people with these radios. I'm not one of them. Our focus is on how this integrates with WSJT-X and the broader streaming architectural landscape which you and I have previously discussed.

mikaelnousiainen:
👍 Thanks. This definitely needs wider testing so we can start stabilizing the codebase. So let's just try to get this merged to master (Hamlib 5.0 development), so that it's then available for integration and testing.

I will be monitoring and participating any discussion / issue reports / PRs regarding the streaming-related features.

mikaelnousiainen:
I think I got through (most if not) all the flaky test weirdness. However, there is yet another issue in rigctld (and the other daemons) with some rapid client connect/disconnect scenarios that seems to happen from time to time in tests. I will file a separate PR for it, as it's not the responsibility of this PR to start fixing bugs outside of the PR scope.

I went to re-run the last failed ubuntu GHA test job, hoping it will clear on its own.

@N0NB @dforsi If we don't get any further feedback here in a couple of days, I think this PR is finally ready for merging.

As always, I'll be monitoring for bug reports and PR and doing "real-world testing" on my own as now the network backend + streaming seems to be in a usable state!

mikaelnousiainen:
Filed #2222 to fix what I found during testing this Icom PR.

rigs/icom/frame.c
```diff
@@ -351,6 +386,28 @@ static int icom_one_transaction(RIG *rig, unsigned char cmd, int subcmd,
...
+    if (buf[4] != ACK && buf[4] != NAK && buf[4] != COL)
+    {
+        if (cmd != buf[4])
+        {
+            rig_debug(RIG_DEBUG_VERBOSE, "%s: cmd x%02x != buf x%02x so retry read\n",
+                      __func__, cmd, buf[4]);
+            goto again2;
+        }
+
+        if (subcmd != -1 && frm_len > 5 && subcmd != buf[5])
```
dchristle, line 402: The multi-byte subcommands can break existing serial radios, c.f. the IC-756PROII and IC-746 getters and commands such as 0x501 and 0x502. Right here, it's comparing the integer with a single reply byte.

rigs/icom/network_session.c
```diff
@@ -0,0 +1,2968 @@
...
+{
+    uint8_t rbuf[ICOM_NETWORK_SESSION_PACKET_MAX];
+    int response_length, ret;
+
+    if (!s->connected || s->audio.fd < 0 || s->audio_server_port == 0)
+    {
+        return -RIG_EINVAL;
+    }
+
+    if (s->audio_active) { return RIG_OK; }
```
dchristle, line 2523: `transport_lock` is held as a shared read lock here, so concurrent AUDIO_RX and AUDIO_TX opens can actually both read (observe) `audio_active == 0` and enter the startup path. They will race -- they reset and handshake on the same audio socket concurrently, and can each create an audio thread while overwriting `s->audio.thread`.

Could this transition be protected by a small session-level mutex around the check, handshake, and thread creation?

dchristle, line 2523: `transport_lock` is held as a shared read lock here, so concurrent AUDIO_RX and AUDIO_TX opens can actually both read (observe) `audio_active == 0` and enter the startup path. They will race -- they reset and handshake on the same audio socket concurrently, and can each create an audio thread while overwriting `s->audio.thread`.

Could this transition be protected by a small session-level mutex around the check, handshake, and thread creation?

rigs/icom/icom_network.c
```diff
@@ -0,0 +1,1203 @@
...
+        stream_caps[n++] = rx;
+
+        if (!m->rx_only)
+        {
+            tx = rx;
+            tx.type = RIG_STREAM_TYPE_AUDIO_TX;
+            stream_caps[n++] = tx;
+        }
+
+        if (m->iq_capable)
```
dchristle, line 1181: nit: Is opening `AUDIO_RX` and `IQ_RX` together meant to be supported? This publishes them as separate stream types, so the core permits one of each, but both RX workers consume the same session queue. This is a super edge case that probably can't happen (I can't think of how) but technically a client that overlaps them during a handoff, or keeps both open, splits packets between the two outputs. If they are alternate views of one radio-side flow, could the backend reject the second RX open instead?

rigs/icom/icom_network.c
```diff
@@ -0,0 +1,1203 @@
...
+        rig_debug(RIG_DEBUG_WARN, "%s: %u TX audio send failure(s) on this stream\n",
+                  __func__, st->tx_send_errors);
+    }
+
+    stream->backend_priv = NULL;
+    icom_network_stream_state_free(st);
+
+    /* The session's audio flow is stopped at disconnect, so concurrent RX/TX
+     * streams keep working; closing one stream does not tear down the other. */
+    return RIG_OK;
```
dchristle, line 967: This preserves the shared flow while another RX or TX stream is open, but it also leaves it running when this was the last stream. The audio thread keeps filling `audio_rx_q`; a later RX open takes the `audio_active` fast path and reads that stale queue, including any overrun accumulated while no stream existed. Could we keep a small session-level stream-user count, stop and clear the audio flow on the last close, and retain it while a concurrent RX/TX stream still needs it?

src/stream_net.c
dchristle, line 706: WDYT about making this go through `stream_backend_write()` as well? The new ERROR handling marks the stream failed, and codec frames already use the failed-stream-gated producer path. If there's a raw PCM datagram that's delayed, it'll still write directly to the ring after `-RIG_EIO` & that might mean a later read can then succeed again instead of remaining terminal.

rigs/icom/network_session.c
```diff
@@ -0,0 +1,2968 @@
...
+        }
+
+        return ret;
+    }
+
+    s->connected = 1;
+    pthread_mutex_lock(&s->loss_lock);
+    s->lost = 0;
+    s->loss_reason = RIG_COMM_REASON_NONE;
+    pthread_mutex_unlock(&s->loss_lock);
```
dchristle, line 2365: Could we fail the active backend streams before clearing lost here? In the contract for reconnecting, it says that open streams must be reopened, so this needs to be terminal for the old streams. The per-stream RX/TX workers are still running even though re-establishment tears down the session audio flow

dchristle (changes requested):

PR #2179, opened 2026-08-28
Execute the CI actions for pushes and pull requests on all branches

dforsi:
Drop the filtering of branch names so that actions are always executed. Now we have 2 active branches (`master` and `Hamlib-4.7`) and without filtering this will still work if we have more branches or with different names (e.g. `Hamlib-5.0`) in the future.

Currently, actions aren't executed for PRs on the `Hamlib-4.7` branch because the original workflows were executed only for the `master` branch (possibly the only active branch at that time), then I added `feature/*` for a `git-flow` style that isn't used in the official, I use it in my repository from which I create my PRs.

### DIFF
--- a/.github/codeql-analysis.yml
+++ b/.github/codeql-analysis.yml
@@ -2,10 +2,8 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [master, ]
   pull_request:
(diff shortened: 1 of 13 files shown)


N0NB:
Thanks, Daniele!

PR #2184, opened 2026-08-29
Bring IC7760 to full feature compatibility (and also few generic fixes for issues that were noticed and impacted 7760)

arekm:
Generic fixes:

- 3bc3e5838 fix(rig): tie asynchronous port reads to the reader thread's
  lifetime — the asyncio flag routed reads through the sync pipes before
  the reader thread existed and after it stopped, so nothing fed them.
- cac81a0b9 fix(rig): leave nothing started when the reader thread fails
  to start — a failed pthread_create() left pipe-routing on, the run flag
  set, and the private data allocated.
- 33de1f563 fix(rig): let the tuner control program handle only the
  tuner — an installed tuner-control program hijacked every
  rig_set_func(), not just RIG_FUNC_TUNER.
- d4482576f fix(icom): send the antenna selection once, not twice — rigs
  with a 3-byte antenna command got the identical frame twice, a
  leftover of a two-step removed in 4.3.
- e173656c0 fix(icom): start the select memory scan with its own
  subcommand — RIG_SCAN_SLCT sent 0E 01, the plain memory scan; the
  select scan is 0E 23 and had never been sent.
- 9554cbc98 fix(rig): stop a refused get_vfo from erasing it for the
  whole model — one refused get_vfo wrote NULL into the shared per-model
  rig_caps, disabling get_vfo for every handle in the process.
- 98e48335e refactor(icom): name the scope byte and drop the VFO-number
  macros — SCOPE_MAIN/SCOPE_SUB for the scope byte of the 27 commands;
  no behaviour change.
- 83eb62437 fix(icom): send scope settings for CURR to the selected
  receiver — RIG_VFO_CURR always went to the Main scope, and the mask
  mixed up MAIN_B, SUB_A and a bare VFO B on rigs with A/B on both
  receivers.
- 4201ad83a fix(icom): decode a negative lower edge in scope waveform
  data — a lower edge below 0 Hz (F marker in the 1 GHz digit of the
  five-byte field) was read as plain BCD, giving a 15 GHz edge and a
  negative span.

Each of these has a regression test (testtunercontrol, testicomant,
testicomscan, testicomgetvfo, testic9700civ, testicomscope), plus
build/test-infra commits: c8075cf79 (clean up the rigctld logs left by
the stream tests), 1af5080a0 (Winsock init in the socket tests),
4d2c573e8 (.gitignore gaps).

IC-7760 improvements and fixes:

75519e885 feat(ic7760): expose the OVF indicator and three more switches
8187a4e98 fix(ic7760): read back which of Main and Sub is selected
2347b7cc3 fix(icom): address the IC-7760 Main and Sub bands by name
2573342b6 fix(ic7760): let the voice memories be transmitted
bf85c4dfa feat(ic7760): advertise more of the attenuator range
0f3097ca5 feat(ic7760): advertise the select memory scan
42fd203f5 feat(ic7760): support the spectrum scope
6376e42fd fix(ic7760): the lowest CI-V rate is 4800, not 300
8139453f6 fix(ic7760): the antenna command carries three bytes, not four
ceb20752e fix(ic7760): advertise dual watch and the CI-V transceive switch
54e8a22a4 fix(ic7760): drop the AGC OFF setting the rig rejects
b9f2eda03 fix(ic7760): drop the priority scan the rig does not have
a3e8cd11d fix(ic7760): advertise the tuning step the rig actually has
024d08dd1 fix(ic7760): let the CW pitch keep its shared granularity
b9d74d5a7 fix(ic7760): correct the AM minimum power on 6 m
5c9328c2d fix(ic7760): add the missing 5 A point to the Id meter scale
103127b74 fix(ic7760): P.AMP 1 has 12 dB of gain, not 10
2e48f8c5e fix(ic7760): make the DRIVE gain and DIGI-SEL controls visible
ef69ab28a fix(ic7760): connect the backlight setting to a command
3e4d82738 fix(ic7760): give the SSB data modes their own filter widths
6ec7b1adb fix(ic7760): correct the default IF filter widths
64b3edda3 fix(ic7760): list all eight keyer and voice memories
e38d962a5 fix(ic7760): stop asking the rig whether it is switched on
271c3420f fix(ic7760): stop advertising two controls the rig does not have
7f6226857 fix(ic7760): send the VOX delay to the right setting

Test commits for the backend: 8fd33224e (caps pinned to the CI-V guide
revision A7788-8EX-2), fc7ced32f (peer buffer sizing in the socket
test), 44119cda3 and 5da88b22f (IC-7760 frame tests for the shared
get_vfo and scope fixes above). Values verified against the IC-7760
CI-V reference guide and, where marked in the commits, on real
hardware.

Does 4.7 branch require separate pull request?

Note: Claude was used to work on these. But I've tried to test what's
possible or what I can on my hardware (as mentioned in commit logs).


arekm:
Possible asyncio cleanup on top of that:

  https://github.com/arekm/Hamlib/compare/dev/ic7760...dev/asyncio-architecture

PR #2190, opened 2026-08-30
Add information on master and Hamlib-4.7 differences

dforsi:
(cherry picked from commit d7e94b5d0f07164901f503a102e4b790ec0f366a)

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,16 @@ applications.
 The master repository is https://github.com/Hamlib/Hamlib  
 The backup repository is https://sourceforge.net/projects/hamlib/  
 
-Daily snapshots are available at https://hamlib.sourceforge.net/snapshots/
+Daily snapshots of the *master branch* are available at
+https://hamlib.sourceforge.net/snapshots/
+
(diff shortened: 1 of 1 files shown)


N0NB:
The only change I would offer is that 5.0 has diverged enough that incompatibilities already exist.  However, since we're getting closer to a 4.7.3 release, it's probably a bit early to ask downstreams to begin porting.

@GeoBaltz @mikaelnousiainen What do you think?

GeoBaltz:
I would hope they started porting a year ago, but I'm not holding my breath.

I sent patches to both tlf and wsjt-x for the changes needed to compile/run on both 4.x and 5.0.  The git master of tlf works fine for me with 5.0, and should see no difference on 4.7.x; no release yet.  I have no idea what wsjt-x has done with the PR I submitted, but I can build & run on 5.0 with that patch.

The whole idea was to make it easy for the apps to have code that will function on both, adapting to whatever Hamlib is installed at compilation time.  At run time, both libraries can be installed (different .so versions) so apps can upgrade on their own timeline.

PR #2204, opened 2026-09-10
fix(tests): avoid duplicate parser state definitions

dchristle:
This PR removes duplicate definitions of `lock_mode` and `rig_powerstat` from `testctlparser`. These variables are already provided by core `rig.c`; defining them again causes GCC 16/Linux static links to fail with multiple-definition errors.

Tested with the focused static build and `testctlparser`.

### DIFF
--- a/tests/testctlparser.c
+++ b/tests/testctlparser.c
@@ -19,9 +19,6 @@
 
 #include "rigctl_parse.h"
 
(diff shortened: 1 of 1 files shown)


dforsi:
Thanks David.

Merged and cherry-picked to 4.7.

PR #2212, opened 2026-09-13
Fix macro argument parentheses for thread-safety and correctness

woodard:
This commit fixes macros where arguments are not properly enclosed in parentheses, which can cause issues when macro arguments are expressions containing operators.

The fixes follow CERT C coding standard rules for macro definitions.

Files modified:
- include/hamlib/amplifier.h: Fixed AMP_STATUS_N(n) macro
- include/hamlib/amplist.h: Fixed AMP_MAKE_MODEL(a,b) and AMP_BACKEND_NUM(a) macros
- include/hamlib/rig.h: Fixed Hz, kHz, MHz, GHz, s_Hz, s_kHz, s_MHz, s_GHz, RIG_VFO_N(n), and RIG_ANT_N(n) macros
- include/hamlib/riglist.h: Fixed RIG_MAKE_MODEL(a,b) macro
- include/hamlib/rotlist.h: Fixed ROT_MAKE_MODEL(a,b) macro

Example of the fix:
Before: #define RIG_VFO_N(n) (1u<<(n))
After:  #define RIG_VFO_N(n) ((1u)<<((n)))

This ensures that when the argument is an expression like (n+1), the expansion works correctly.

This particular policy is documented many places and also implemented in many static analyzers. So it is probably a good idea to update the source code to follow current best practices.

I don't know what legacy compilers Hamlib seeks to support but C has supported inline functions since C99, it may be worthwile to replace many of those macros with inline functions.


dforsi:
AFAICT this PR is doubling existing parentheses, even around the integer constant `1u`.

I suggest closing as "won't fix"

N0NB:
@woodard The baseline for 5.0+ is C11/C17.

woodard:
I would like to argue that it is not just doubling parenthesis there is a good reason for CERT C standards (PRE02-C: Macro replacement lists should be parenthesized and PRE01-C: Use parentheses within macro replacements), wrapping both the macro's parameters (n) and the overall expression (1u) in parentheses prevents critical flaws.

Here is an example of the problem. If you pass RIG_VFO_N(x & y), it expands to 1u << x & y. Because << has higher precedence than &, it evaluates as (1u << x) & y instead of 1u << (x & y). This is the kind of unobvious macro expansion that the extra parenthesis prevent. Adding the extra parenthesis does not add any additional run time complexity and an insignificant amount of extra compilation time. So I would argue that it is just a good idea.

Given the overall target is C11/C17, what I would do if this were my project is accept this patch but then suggest to the contributor, i.e. me, that they convert these macros to inline functions so that this kind of problem where you have to be a language lawyer to spot these kinds of problems doesn't come up.

Issue #1943, opened 2025-11-01
FlexRadio network CAT protocol support


mikaelnousiainen:
This is the main issue for ARDC Hamlib improvement project [Milestone 5 - FlexRadio network CAT protocol support](https://github.com/Hamlib/Hamlib/milestone/28) - plans and discussion. More granular issues will be created as we progress through the milestone.

# Project background

* See ARDC grant: https://www.ardc.net/apply/grants/2025-grants/hamlib-stability-enhancements-and-sdr-transceiver-support/
* See Open Collective page: https://opencollective.com/hamlib-stability-enhancements-and-sdr-transceiver-support

The project backlog is available in: https://github.com/orgs/Hamlib/projects/1 (still in progress)

#  Development Process

* GitHub issues in the main Hamlib repository will track the project's progress, milestones, and deliverables transparently.
* All code contributions will be made as GitHub Pull Requests against the main Hamlib repository for final community review and comments
* All code will be licensed under the existing GNU LGPL 2.1 (or later) license

# Description

To be written after [Milestone 3 - Network audio and I/Q data streaming support](https://github.com/Hamlib/Hamlib/milestone/26) is complete.

mikaelnousiainen:
We're testing FlexRadio backend, the audio + I/Q streaming is already working well (architecturally mirrors the Icom network backend we will publish first), but we're also adding most Hamlib backend features for rig control (that were missing) which takes some time to test and verify. If things go well, it's likely we will get this published for testing in September, latest.

mikaelnousiainen:
PR open for review in #2213

PR #2213, opened 2026-09-13
Extend FlexRadio backend. Add support for audio and I/Q data streaming.

mikaelnousiainen:
TL;DR: This is a large PR for extending the FlexRadio backend and adding audio + I/Q data streaming support. The implementation uses the Hamlib data streaming subsystem for delivering audio (and I/Q) data over network, merged in https://github.com/Hamlib/Hamlib/pull/2116. This PR implements https://github.com/Hamlib/Hamlib/issues/1943. Going for a large and complete PR was a (again) good way to produce something that's a cohesive and a working solution. 

We've tested this on a FLEX-8400M, but at some point its control part (the screen + knobs) failed during an update, so we were only able to control the rig via network. We have not yet done extensive, multi-hour streaming tests and it may be some failure scenarios may still need better handling. These can be fixed in follow-up PRs. Anyway, I think this should be ready to merge soon as it does not really change any existing behaviors.

Please do give this a try and send any feedback, so I can incorporate any critical fixes as soon as possible!

As a first step, please read the docs: `rigs/flexradio/FLEXRADIO.md` and the updates done to `HAMLIB_STREAMING.md` (design and normative wire protocol) and `HAMLIB_STREAMING_BACKEND_GUIDE.md` (how to add streaming to a backend).

LLM-generated summary of the PR details:

# FlexRadio SmartSDR: a working backend — control, metering, spectrum, and DAX streaming

This turns the existing **SmartSDR backend into one an application can actually
use**. The backend in tree connects and tunes, but declares one function
(`MUTE`) and one level (`PREAMP`, which the radio does not implement); there is
no metering, no spectrum, no slice handling worth the name, and no audio. This
PR adds those, and carries the radio's DAX audio and DAX-IQ through the data
streaming subsystem merged earlier.

Until now a FLEX could be tuned from Hamlib and nothing else: audio meant
running SmartSDR or DAX on Windows alongside whatever else you were doing. With
this, `rigctl -m 23013 -r <radio-ip>:4992` gives 25 levels, 15 functions, the
panadapter and the meters, and `rig_stream_open()` returns the receive audio —
the same API and the same `rigctld` connection as any other backend.

**Target:** 5.0 development. Built on the streaming subsystem already merged;
this PR adds no new public streaming API.

## What is in this PR

- **A slice model, `23013`.** SmartSDR presents up to eight receivers
  ("slices"); the eight existing per-slice models `23005`–`23012` each hard-wire
  one. The new model takes a `slice` setting instead, and can create the slice
  it needs and remove it again on close. The old models keep working unchanged —
  no configuration needs editing — and are now thin wrappers over the same code.
- **Control that reflects the radio.** 25 levels and 15 functions, each checked
  against the hardware rather than declared hopefully: AF/RF/SQL/AGC, NR/NB/APF,
  MICGAIN/COMP/VOX/monitor, CW pitch and speed, break-in delay, and the meter
  levels below. Antenna selection, tuning step, RIT/XIT, split, VFO ops.
  `PREAMP` is *removed* — the radio has no preamp, and declaring one was worse
  than declaring nothing.
- **Metering from the radio's own telemetry.** `STRENGTH`, `SWR`, `ALC`,
  `RFPOWER_METER` and `..._WATTS`, `TEMP_METER`, `VD_METER`, `ID_METER`, decoded
  from the VITA-49 meter stream and scaled against known-good readings — an idle
  PA at 27.6 °C, a 14.03 V supply, an SWR of exactly 1.00 with no carrier.
- **Spectrum** from the panadapter FFT, opt-in through the `spectrum` token, via
  the existing `rig_spectrum_line` callback.
- **Streaming** — `AUDIO_RX`, `AUDIO_TX` and `IQ_RX` over DAX. Audio is
  a 24 kHz stereo float pair; DAX-IQ runs at 24/48/96/192 kHz. The wire formats
  are fixed by the protocol, so the backend advertises exactly those and leaves
  every other format, rate and channel count to the frontend's conversion.
- **Session handling** — a reader thread on the control connection, keepalive,
  a configurable liveness timeout that notices a radio that has gone away, and
  opt-in `auto_reconnect` that rebuilds the session in the background.
- **`nat_traversal`** for reaching a radio through manual port forwarding, and
  the fix that makes audio work at all on a radio with SmartLink configured.
- **`rigs/flexradio/FLEXRADIO.md`** — twelve sections covering models,
  connecting, every token, what is implemented, the slice/VFO/split model,
  streaming, metering, spectrum, how it works, and troubleshooting. Replaces the
  directory's `README`.
- **Tests** — 183 cases across nine suites against an in-process mock radio that
  speaks the TCP protocol and sends real VITA-49 audio, I/Q, meter and FFT
  packets, plus `simflex`, a simulator the integration suite drives end to end.

64 files, about 21,300 added lines. Roughly half is tests and documentation.

## Main features

- **Slices are the VFO model, honestly mapped.** SmartSDR has no A/B: it has
  slices, and split means a second slice marked `tx=1`. `Main` is this model's
  slice and `Sub` is the transmit slice, `set_split_vfo` creates and removes the
  transmit slice, and `split_slice` pins it if you would rather choose. §5 of
  the documentation explains the mapping rather than pretending it is a Yaesu.
- **Status is scoped to the slice it belongs to.** The backend previously
  applied any slice's status to whichever slice it was bound to, so two live
  slices corrupted each other's frequency and mode. Found with two slices open
  on hardware.
- **DAX audio, and DAX-IQ, decoded correctly.** These are not the same format
  and the difference is not documented anywhere: audio is big-endian float at
  full scale 1.0, while DAX-IQ is *little-endian* float carrying left-justified
  fixed point at full scale 32768. See "A note on DAX-IQ" below.
- **Hardware timestamps when the radio has a clock worth carrying.**
  `RIG_STREAM_CAP_HW_TIME` is advertised, and each read reports which source it
  actually got — GPSDO, the radio's TCXO, or the host clock — rather than
  claiming traceability the radio cannot provide.
- **Transmit routing that survives contact with the radio.** SmartSDR keeps the
  input selection and the DAX flag separately and they can contradict each
  other, so `tx_audio_source` writes both together, and opening a transmit
  stream selects `dax` for as long as it is open — streamed audio is discarded
  otherwise.

## Trying it out

Needs a FLEX-6000 or 8000 series radio on the network. No credentials: SmartSDR
is an open TCP protocol on port 4992.

| Model | What it is | Hardware-tested |
|---|---|---|
| 23013 | SmartSDR, slice chosen by setting | yes, including transmit |
| 23005–23012 | SmartSDR slices A–H, one model each | yes (A), the rest share the code |

The radio is addressed by IPv4 literal — the control connection would accept a
name, but the VITA stream sockets parse the same string as a dotted quad.

```sh
MODEL=23013

# control
rigctl -m $MODEL -r 192.168.0.252:4992

# slice B instead of A
rigctl -m $MODEL -r 192.168.0.252:4992 -C slice=B

# receive audio to a WAV file
rigstreamtest -m $MODEL -r 192.168.0.252:4992 -t audio_rx -s 24000 -c 2 \
    -d 10 -w capture.wav

# I/Q at 192 kHz
rigstreamtest -m $MODEL -r 192.168.0.252:4992 -t iq_rx --iq -s 192000 -c 1 -d 10

# everything the model advertises, with recordings
tests/rigstreamtest-hw.sh -m $MODEL -r 192.168.0.252:4992
```

`FLEXRADIO.md` §2–§3 covers connecting and every token, §5 the slice and split
model, §9 how it works, §10 troubleshooting.

No hardware is needed to review this: the nine suites run control, metering,
spectrum, session loss and the streaming paths against the in-process mock.

## A note on DAX-IQ

Worth calling out because it is a trap for anyone else implementing this.

Every VITA-49 payload the radio sends is big-endian network order — except
DAX-IQ, which is little-endian. Byte-swapping it, as the rest of the payloads
require, turns valid samples into denormals: peaks of 8e-41, which print as
`0.000000` through a `%f`. Nothing else looks wrong. The packets are the right
size, arrive at the right rate and lose nothing, so every counter reads clean.

The samples are also counts, not the ±1.0 the streaming API documents for float
formats: the radio fills a float with left-justified fixed point, so full scale
is 32768.

Both halves are handled, and both simulators now send I/Q the way the radio
does — previously they sent it under the audio class code in big-endian, so
they agreed with a byte-swapping backend and the pair were wrong together. The
I/Q assertions were strengthened from "some sample is non-zero", which garbage
satisfies, to the contract itself: no denormals, everything inside ±1.0, and a
peak matching the simulator's half-scale tone. Both are mutation-verified —
restoring the byte-swap or dropping the normalisation each fails a test that
names which mistake was made.

**There is no I/Q transmit, and this PR does not claim one.** DAX I/Q binds to a
panadapter, a receive object, and the radio has no transmit stream type for it:
both `stream create type=dax_iq ... tx=1` and `type=dax_iq_tx` come back as a
plain `dax_iq` stream with `endpoint_type=Not Assigned`. Measured
content-differentially on a FLEX-8400M — same path, same power, varying only
the payload — a half-scale tone streamed to a `dax_iq` stream with PTT keyed
produces the same 0.001 W as silence. Transmit from a client is `type=dax_tx`,
the audio path, which `AUDIO_TX` uses.

## A note on VITA-49 timestamps

Smaller than the DAX-IQ trap but the same shape, and worth the paragraph
because it is easy to get wrong and hard to notice.

VITA-49 puts the timestamp selectors TSI at word0 bits 23-22 and TSF at 21-20.
This backend read them two bits high, so every packet came back `TSI_NONE` and
the code concluded the radio sends no time of day. It does: a FLEX-8400M audio
packet with `word0 = 0x38500107` carries `TSI=UTC` alongside an integer
timestamp of `0x6A726120`, which is a Unix epoch and only means anything if the
selector says UTC.

Nothing user-visible changed — the hardware-time branch also requires
`gpsdo_present` and `TSF=REAL_TIME_PS`, neither of which this radio offers —
but the wrong reading had been written down as a fact about the radio.

The unit test that should have caught it could not: `test_vita49`'s packet
builder set the selectors at the same wrong offsets the parser read them from,
so the two agreed on a shared mistake and the round trip passed. `simflex` had
them right all along, but nothing asserted the selector through it, and the
mock sets none. The builder now follows the spec, and the regression test is
built from those captured hardware bytes rather than from any builder, so it
cannot drift the same way again.

## Known limitations

- **DAX audio is 24 kHz stereo, and only that.** Not a backend limit — it is
  what the wire carries. Anything else is the frontend's conversion, which needs
  libsamplerate for rate changes and many builds do not have it.
- **A panadapter must be bound to a slice.** A standalone one created with
  `display pan create` sends FFT packets at the normal rate but emits one real
  frame and then a flat line forever, which looks exactly like a working
  spectrum of a dead band. The backend binds one; the failure mode is documented
  because it is worth knowing about.
- **Spectrum amplitude is unverified against a known signal.** The bins are
  heights, not calibrated levels, and no reference source was available.
- **The GPS time branch is untested** — it needs a GPSDO-equipped radio, and the
  one here reports `gpsdo_present=0`. The branch also wants
  `TSF=REAL_TIME_PS`, which this radio does not send; it sends a sample count.
  (An earlier draft said both streams declare `TSI=NONE`. That was this
  backend's own parser reading the selector two bits high — see the note on
  VITA-49 timestamps below. The radio sends `TSI=UTC`.)
- **Opus is specified but not implemented.** The radio offers it; carrying a
  compressed payload needs core support that does not exist yet, so it is
  written up rather than half-built.
- **Only slice A of the per-slice models is hardware-tested.** B–H share the
  same code path with a different index.

## Testing

`make check` passes: nine suites new here, 183 cases, covering configuration,
levels and metering, radio properties, the session state machine, slices, VFO
and split, stream open/close, the UDP dispatch path, and the VITA-49 codec —
all against the in-process mock, so they run with no hardware in CI.

Some of those cases exist because the happy path was already covered and was
not the interesting part. The status parser is fed truncated lines, fields with
no value, values that are not numbers and lines with no fields at all, and is
then required to have moved nothing and to still absorb the next good line —
which found a real defect on its first run: `RF_frequency=`, all a dropped TCP
segment looks like, left the rig reporting 0 Hz, indistinguishable to a reader
from a radio genuinely tuned there. The command-ordering test asserts that
`set_split_mode` sends the width *after* the mode, because the radio resets a
slice's filters when the mode changes and both commands are sent either way, so
asserting they were seen cannot tell a working sequence from a broken one.

Hardware validation on a **FLEX-8400M**, transmit at 1% into a dummy load on
ANT1: the full `rigstreamtest-hw.sh` sweep including full-duplex RX/TX, I/Q at
all four declared rates measured against expected byte rates (99–100%, zero
gaps), and every advertised level and function read back individually — all 25
levels and all 15 functions return values. `rigctl -m N -u` reports
`Overall backend warnings: 0` for all twelve FlexRadio models.

The full-duplex recording is the clearest single piece of evidence: PTT held for
20 s then released, and the capture shows the receiver live, then twelve seconds
of exact digital silence while the radio muted its own receiver, then live
again — with the gap counter at zero throughout. Silence arrived as samples on a
continuous timeline rather than as missing data.

The Windows cross build and its Wine test run were reproduced locally in a
container following `windows-cross-features.yml` step for step, and the Linux
`make distcheck` job likewise. Both found real failures, fixed here: the
*shipping backend* included POSIX socket headers directly and did not compile
for Windows at all; `inet_aton()` does not exist in winsock; `simflex` needed
the same treatment plus `$(NET_LIBS)`; and the new comm-reason symbols were
missing from `bindings/python/test_Hamlib_class.py`, which fails `distcheck`
under `--enable-pytest`.

Memory: all nine suites clean under AddressSanitizer on macOS, and clean under
LeakSanitizer on Linux — LSan is not supported on macOS, so leak coverage has to
come from a container.

## Relationship to the Icom network PR

Both branches carry the same small core additions, and whichever lands first
brings them: `rig_comm_reason_t` and `rig_strcommreason()` for saying *why* a
link dropped, and the `rig_close()` teardown reorder that tears streams down
before the backend releases what their threads are using. If the Icom PR is
merged first, those hunks disappear from this one.

## Use of LLMs

LLMs were used as an assistant in producing this work.

The Hamlib backend architecture/implementation and the hardware debugging are
author's work. LLM assistance was used more heavily for the tests and the
documentation, and for routine or mechanical work throughout.

All of it has been reviewed, built and tested by the author, and validated
against a real radio.


mikaelnousiainen:
I'll again aim to keep this PR open for a few weeks to get feedback and any critical fixes in. As we had trouble with our FlexRadio control part failing, I encourage folks to test this on their FlexRadios and give feedback! Especially RX and TX stability + the new Hamlib controls added should be verified with as many FlexRadio models as possible.

The `rigstreamtest-hw.sh` utility is a good place to start (documented in the PR description).

mikaelnousiainen:
Will be fixing the build failures also :)

rigs/flexradio/smartsdr_conf.c
```diff
@@ -0,0 +1,321 @@
...
+    },
+    {
+        TOK_TX_AUDIO_SOURCE, "tx_audio_source", "Transmit audio source",
+        "What the radio modulates from: mic, acc, pc or dax. The radio holds "
+        "the input selection and the DAX flag separately, and they can "
+        "contradict each other, so both are written together. Opening a "
+        "transmit stream selects dax for as long as it is open, since "
+        "streamed audio is discarded otherwise.",
+        "mic", RIG_CONF_COMBO, { .c = { .combostr = { "mic", "acc", "pc", "dax", NULL } } }
+    },
```
dforsi, line 63: Just a thought: it seems that those are the first `combostr` in lower case.

```
$ grep -rn RIG_CONF_COMBO --include=*.{c,h} | grep .combostr
rigs/dummy/dummy.c:108:        "VALUE1", RIG_CONF_COMBO, { .c = { .combostr = { "VALUE1", "VALUE2", "NONE", NULL } } }
rigs/dummy/rot_dummy.c:84:        "VALUE1", RIG_CONF_COMBO, { .c = { .combostr = { "VALUE1", "VALUE2", "NONE", NULL } } }
rigs/kenwood/tmd710.c:248:        NULL, RIG_CONF_COMBO, { .c = { .combostr = { "A", "B", "TXA-RXB", "TXB-RXA", NULL } } }
rigs/kenwood/ts480.c:146:        NULL, RIG_CONF_COMBO, { .c = { .combostr = { "OFF", "DNL Level 1", "DNL Level 2", "DNL Level 3", NULL } } }
rigs/kenwood/ts480.c:150:        NULL, RIG_CONF_COMBO, { .c = { .combostr = { "OFF", "Hb1", "Hb2", "FP", "bb1", "bb2", "c", "U", NULL } } }
rigs/kenwood/ts480.c:154:        NULL, RIG_CONF_COMBO, { .c = { .combostr = { "OFF", "Hb1", "Hb2", "FP", "bb1", "bb2", "c", "U", NULL } } }
rigs/kenwood/ts480.c:158:        NULL, RIG_CONF_COMBO, { .c = { .combostr = { "2.0 kHz", "2.4 kHz", NULL } } }
rigs/kenwood/ts2000.c:141:        NULL, RIG_CONF_COMBO, { .c = { .combostr = { "OFF", "H BOOST", "F PASS", "B BOOST", "CONV-EN", "USER", NULL } } }
rigs/kenwood/ts2000.c:145:        NULL, RIG_CONF_COMBO, { .c = { .combostr = { "OFF", "H BOOST", "F PASS", "B BOOST", "CONV-EN", "USER", NULL } } }
rigs/kenwood/ts2000.c:149:        NULL, RIG_CONF_COMBO, { .c = { .combostr = { "2.0 kHz", "2.2 kHz", "2.4 kHz", "2.6 kHz", "2.8 kHz", "3.0 kHz", NULL } } }
rigs/kenwood/k3.c:105:        NULL, RIG_CONF_COMBO, { .c = { .combostr = { "SWR", "ALC", NULL } } }
rigs/kenwood/ts590.c:134:        NULL, RIG_CONF_COMBO, { .c = { .combostr = { "OFF", "Hb1", "Hb2", "FP", "bb1", "bb2", "c", "U", NULL } } }
rigs/kenwood/ts590.c:138:        NULL, RIG_CONF_COMBO, { .c = { .combostr = { "OFF", "Hb1", "Hb2", "FP", "bb1", "bb2", "flat", "U", NULL } } }
rigs/kenwood/ts590.c:166:        NULL, RIG_CONF_COMBO, { .c = { .combostr = { "10 Hz", "100 Hz", "200 Hz", "300 Hz", "400 Hz", "500 Hz", NULL } } }
rigs/kenwood/ts590.c:170:        NULL, RIG_CONF_COMBO, { .c = { .combostr = { "2500 Hz", "2600 Hz", "2700 Hz", "2800 Hz", "2900 Hz", "3000 Hz", NULL } } }
rigs/kenwood/ts590.c:174:        NULL, RIG_CONF_COMBO, { .c = { .combostr = { "10 Hz", "100 Hz", "200 Hz", "300 Hz", "400 Hz", "500 Hz", NULL } } }
rigs/kenwood/ts590.c:178:        NULL, RIG_CONF_COMBO, { .c = { .combostr = { "2500 Hz", "2600 Hz", "2700 Hz", "2800 Hz", "2900 Hz", "3000 Hz", NULL } } }
rigs/icom/icom.c:553:    { TOK_SCOPE_MSS, "SPECTRUM_SELECT", "Spectrum Scope Main/Sub", "", "", RIG_CONF_COMBO, { .c = { .combostr = { "Main", "Sub", NULL } } } },
rigs/icom/icom.c:554:    { TOK_SCOPE_SDS, "SPECTRUM_DUAL", "Spectrum Scope Single/Dual", "", "", RIG_CONF_COMBO, { .c = { .combostr = { "Single", "Dual", NULL } } } },
rigs/icom/icom.c:555:    { TOK_SCOPE_EDG, "SPECTRUM_EDGE", "Spectrum Scope Edge", "Edge selection for fixed scope mode", "", RIG_CONF_COMBO, { .c = { .combostr = { "1", "2", "3", "4", NULL } } } },
rigs/icom/icom.c:557:    { TOK_SCOPE_CFQ, "SPECTRUM_CENTER", "Spectrum Scope Center Frequency Type", "", "", RIG_CONF_COMBO, { .c = { .combostr = { "Filter center", "Carrier point center", "Carrier point center (Abs. Freq.)", NULL } } } },
rigs/icom/icom.c:558:    { TOK_SCOPE_VBW, "SPECTRUM_VBW", "Spectrum Scope VBW", "Video Band Width", "", RIG_CONF_COMBO, { .c = { .combostr = { "Narrow", "Wide", NULL } } } },
rigs/icom/icom.c:559:    { TOK_SCOPE_RBW, "SPECTRUM_RBW", "Spectrum Scope RBW", "Resolution Band Width", "", RIG_CONF_COMBO, { .c = { .combostr = { "Wide", "Mid", "Narrow", NULL } } } },
```

rigs/flexradio/smartsdr_level.c
```diff
@@ -0,0 +1,546 @@
...
+        snprintf(value, sizeof(value), "%d", val.i);
+    }
+
+    RETURNFUNC(smartsdr_prop_set(rig, idx, value));
+}
+
+
+
+/* Meters the radio streams, as opposed to the settings it reports in status.
+ * Returns 1 when it handled the level. */
```
dforsi: This function returns 1 also if `if (*retval != RIG_OK)` at line 266

mikaelnousiainen:
@KJ5HST @dchristle @jfrancis42 Notifying you all here in case you wish to take a look - as you've been involved in the data streaming development. The Icom PR (#2178) refactoring will likely lead to some changes here as well, so I'll first try to wrap up the Icom PR.

rigs/flexradio/FLEXRADIO.md
```diff
@@ -0,0 +1,513 @@
...
+
+This directory holds three unrelated FlexRadio backends. Only the first is
+current hardware.
+
+| Backend | Files | Radios |
+|---|---|---|
+| **SmartSDR** | `smartsdr*.c/h` | FLEX-6000 and FLEX-8000 series |
+| DttSP | `dttsp.c` | DttSP-based SDR software |
+| SDR-1000 | `sdr1k.c` | FlexRadio SDR-1000 (parallel port) |
+
```
dforsi, line 15: I found this first part confusing because it is saying what this file does *not* describe and it is called FLEXRADIO.md but i doesn't describe the files flexradio.{c,h} so I suggest moving this part in the README that this PR would delete and renaming this file as SmartSDR.md or SMARTSDR.md

rigs/flexradio/FLEXRADIO.md
```diff
@@ -0,0 +1,513 @@
...
+(`in_use=1`) and that the model matches the slice you mean. A model bound to a
+non-existent slice still opens.
+
+**No audio from a stream that reports success.** Confirm `nat_traversal` is 0
+on a LAN. With it wrongly enabled the radio is never told the client's UDP
+port and sends nothing, while the stream still opens successfully.
+
+**I/Q stream fails to open.** Pass `-c 1`; the caps declare one channel. If it
+opens but no data arrives, add `-p` so a panadapter exists to bind to.
+
```
dforsi, line 462: Pass `-c` and `-p` to what program?

rigs/flexradio/FLEXRADIO.md
```diff
@@ -0,0 +1,513 @@
...
+open — which is exactly why they are failed rather than silently stalled.
+Control commands work again as soon as the session is back.
+
+**`client program` reports `0x10000002`.** The radio's program whitelist does
+not list Hamlib. This is not fatal and is treated as success.
+
+## 11. Testing
+
+```sh
+make -C test check          # includes test_smartsdr_stream and test_vita49
```
dforsi, line 502: This `test` directory is confusing, it is missing a README that tells the difference versus the `tests` directory and why we need both (eg. `test/test_rigstreamtest*` vs `tests/rigstreamtest*`)

rigs/flexradio/smartsdr_rig.c
```diff
@@ -0,0 +1,675 @@
...
+    { RIG_MODE_LSB,    "LSB",  1 },
+    { RIG_MODE_PKTUSB, "DIGU", 1 },
+    { RIG_MODE_PKTLSB, "DIGL", 1 },
+    { RIG_MODE_AM,     "AM",   1 },
+    { RIG_MODE_FM,     "FM",   1 },
+    { RIG_MODE_FMN,    "FMN",  1 },
+    { RIG_MODE_SAM,    "SAM",  1 },
+    { RIG_MODE_RTTY,   "RTTY", 1 },
+    { RIG_MODE_FMN,    "NFM",  0 },
+    /* Hamlib has no RIG_MODE_DFM, so the radio's DFM reads as FM. */
```
dforsi: There are some bits free for new RIG_MODE_* defines

rigs/flexradio/smartsdr_vita.h
```diff
@@ -0,0 +1,135 @@
...
+};
+
+
+/* Parse a VITA-49 header from a raw UDP packet.
+ * Returns 0 on success, -1 on error (packet too short or invalid). */
+extern int vita49_parse_header(const uint8_t *buf, int len,
+                               struct vita49_header *hdr);
+
+/* Byte-swap big-endian float32 samples to host order.
+ * src and dst may not overlap. */
```
dforsi: This comment should be moved before line 105

dforsi (approved):
I have looked at the files, but they are too complex for me to give an opinion.
They have many comments and most functions are short and theese are very good points.

mikaelnousiainen:
Thanks @dforsi ! Will go through these soon - once the Icom PR is ready for merging!

rigs/flexradio/FLEXRADIO.md
mikaelnousiainen: Indeed true, the `tests` directory should be renamed to something else - such as `apps` or `tools`. The new `test` directory is solely for testing. But it's outside of the scope of this PR. Adding README to `test`.

N0NB: Maybe `utilities` is a good descriptive but poor short name.

rigs/flexradio/smartsdr_rig.c
mikaelnousiainen: Uses `RIG_MODE_PKTFM` now

rigs/flexradio/FLEXRADIO.md
mikaelnousiainen: I added some documentation about `test` directory purpose. We should go through the `tests` directory rename (and what belongs there!) maybe in the mailing list.

rigs/flexradio/FLEXRADIO.md
mikaelnousiainen: Fixed with a better description.

rigs/flexradio/FLEXRADIO.md
mikaelnousiainen: Now the FLEXRADIO.md describes what it's about. SmartSDR is the protocol FlexRadios use.

rigs/flexradio/smartsdr_level.c
mikaelnousiainen: Fixed with a pointer variable for `handled`

rigs/flexradio/smartsdr_conf.c
mikaelnousiainen: 👍 Modified to be consistent with uppercase combo choices.

mikaelnousiainen:
First pass of fixes done. Now working to get the Icom PR #2178 first merged, as its fixes to tests will slightly affect this PR too.

rigs/flexradio/smartsdr_stream.c
```diff
@@ -0,0 +1,2252 @@
...
+
+    if (!ss)
+    {
+        return RIG_OK;
+    }
+
+    rig_debug(RIG_DEBUG_VERBOSE, "%s: closing stream_id=0x%x\n",
+              __func__, ss->vita_stream_id);
+
+    ss->running = 0;
```
dchristle, line 898: If the stream is closed while waiting for a future timed SOB, this line clears `running` & that'll cause the wait at line 115 to exit. 

The control flow will fall through to the "PTT-on" call & this means that cancelling a scheduled burst can key the transmitter immediately. I don't think the close path appears to explicitly inky it, either. So, we should abort target processing when stopping & ensure teardown releases any PTT owned by the stream. 

A regression test could schedule the SOB, close before its deadline, and assert that PTT never turns on.

rigs/flexradio/smartsdr_rig.c
```diff
@@ -0,0 +1,677 @@
...
+
+    if (!has_tx)
+    {
+        /* tx= comes from this slice's status and says whether it holds the
+         * transmitter; without it the radio would key a different slice. */
+        rig_debug(RIG_DEBUG_ERR,
+                  "%s: slice %c does not have PTT control\n", __func__,
+                  slicechar[priv->slicenum]);
+        RETURNFUNC(-RIG_ENTARGET);
+    }
```
dchristle, line 323: The `!has_tx` guard conflicts with the split support. Enabling split assigns TX to another slice, so once the status update arrives, the primary slice has `tx=0` and this rejects both PTT ON and OFF. If the status update has not arrived yet, line 341 (`SNPRINTF(cmd, sizeof(cmd), "slice set %d tx=1", priv->slicenum);`) instead reclaims TX for the receive slice. So, this will effectively defeat split.

rigs/flexradio/smartsdr_stream.c
```diff
@@ -0,0 +1,2252 @@
...
+        }
+
+        if (ss->burst_ptt && (tgt.flags & RIG_STREAM_TIME_FLAG_SOB))
+        {
+            rig_set_ptt(ss->rig, RIG_VFO_CURR, RIG_PTT_ON);
+        }
+
+        if (ss->burst_ptt && (tgt.flags & RIG_STREAM_TIME_FLAG_EOB))
+        {
+            rig_set_ptt(ss->rig, RIG_VFO_CURR, RIG_PTT_OFF);
```
dchristle, line 131: This will apply the EOB before sending the samples that precede it (the target lookup includes `tx_consumed + 128`) & this helper runs before the next frame is read/sent. For example, with `tx_consumed=0` and EOB at sample 128, this unkeys before samples 0-127 are sent. We should process the burst boundaries in sample order/ensure the EOB follows transmission of the preceding audio.

A good test to add alongside the fix would be a frame boundary test that asserts the ordering of audio and PTT commands.

dchristle (changes requested):

PR #2223, opened 2026-09-20
Add TH-D75 dual band monitoring support with DL command handlers

woodard:
This commit adds support for the Kenwood TH-D75's True Dual Receive capability through the RIG_FUNC_DUAL_BAND flag and dedicated DL command handlers in thd74.c.

The TH-D75 (and similar dual-band handhelds) feature two independent receiver circuits operating simultaneously, allowing true concurrent monitoring of two frequencies. This is distinct from the traditional "Dual Watch" implementation found on many radios.

Dual Watch vs True Dual Receive Architecture
---------------------------------------------

Dual Watch (HF / Basic Radios):
- Uses a single receiver with rapid time-sharing
- The processor toggles between VFO A and VFO B multiple times per second
- Only one frequency is actively received at any moment
- Audio locks onto whichever frequency first breaks squelch
- Cannot monitor two simultaneous conversations
- May produce audio dropouts or "clicking" during switching

True Dual Receive (Kenwood TH-D75):
- Two completely independent receiver paths operating in parallel
- Each VFO (A and B) has dedicated tuning and filtering circuits
- Can listen to two different frequencies simultaneously
- Independent audio routing and volume control per band
- One band can run background data (APRS, packet) while the other handles voice communications
- Supports V+V, U+U, V+U configurations, with optional HF/microwave monitoring on one side while maintaining ham band activity on the other

This Kenwood radio actually has features that are more similar to the traditional Dual Watch. The Kenwood Call Scan and Priority Scan are similar to Dual Receive.

Kenwood's Call Scan and Priority Scan are software-driven scanning modes that rely on time-division sampling through a single receiver circuit— making their underlying operation functionally identical to traditional Dual Watch, and fundamentally different from the True Dual Receive found in radios like the Kenwood TH-D75.

Call Scan: The radio scans through programmed memory channels and periodically checks a designated Call Channel before returning to the original frequency.

Priority Scan: While tuned to a frequency or scanning memory channels, the radio interrupts reception every few seconds to check a pre- designated Priority Channel for carrier signal. If activity is detected, it switches to the Priority Channel; otherwise, it jumps back.

Why They Are Functionally "Dual Watch"

All three modes (Call Scan, Priority Scan, Traditional Dual Watch) share:
1. Single Receiver Architecture - One physical receiver cannot "see" two frequencies simultaneously
2. Periodic Sampling Dropouts - Audio gaps, ticks, or dropouts occur when the radio retunes to check the priority/call channel
3. Mutual Exclusivity - Cannot hear two transmissions at once; locks onto whichever frequency first breaks squelch

In contrast True Dual Receive

Radios like the Kenwood TH-D75, TH-D74, and TM-V71A feature True Dual Receive with completely different hardware:
- Two Physical Receivers: Two separate independent receiver circuits
- Zero Audio Interruption: No sampling timers, switching glitches, or dropped audio frames
- Simultaneous Audio & Data: Can listen to two bands at once or run background data without interruption

Feature Comparison

| Operational Aspect | Call Scan / Priority Scan | Traditional Dual Watch | True Dual Receive (Kenwood TH-D75) | |--------------------|---------------------------|------------------------|-------------------------------------|
| Receiver Hardware  | 1 Single Receiver         | 1 Single Receiver      | 2 Independent Receivers            |
| Operating Method   | Time-division sampling    | Rapid toggling         | Continuous parallel reception      |
| Audio Impact       | Periodic dropouts/ticks   | Periodic dropouts/ticks| Smooth, uninterrupted             |
| Simultaneous Audio | No                        | No                     | Yes                                |
| Background Data    | Interrupted               | Interrupted            | Uninterrupted                      |

Because this feature is becomming increasingly common in portable and mobile radios and because there are other features in the radio more similar to Dual Watch lieve that it warranted a new new rig feature.

Though this feature is very likely to be applicable to other Kenwood radios like the TH-D74, the TH-D72, and the TH-D7 as well as the mobile radios like the TM-D710 and the new TM-D750 and likely other brand's Radios like Yaesu's which have similar capabilities, I did not implement this feature for any other radios because I do not have hardware to test against. Thefore, I only implemented it against the TH-D75. The hope is that others with access to the hardware will wire up so it works with their hardware.

The DL command handlers in thd74.c control the TH-D75's ability to set and query the dual band monitoring state, enabling the application to distinguish between these operational modes. One thing to note from my perspective the logic of the actual CAT string is reversed. DL 0 means dual band while DL 1 is single band. 

Files modified:
- include/hamlib/rig.h: Add RIG_FUNC_DUAL_BAND flag definition
- rigs/kenwood/thd74.c: Add DL command handlers for thd75_set_func and thd75_get_func
- tests/test_thd75_dual_band.c: Automated test program
- tests/test_thd75_dual_band_interactive.c: interactive test program that requires a user to actually verify what is on the display. This can act as a template for how to test the feature on other hardware.


woodard:
It looks to me like the test that is failing is the one that makes sure nothing was added. In this particular case, that was the point. I added RIG_FUNC_DUAL_BAND. I think that is one that you are going to have to wave that one through. 

If it ends up being something else please let me know.

include/hamlib/rig.h
```diff
@@ -1375,7 +1375,7 @@ typedef uint64_t setting_t;
 #define RIG_FUNC_SEND_VOICE_MEM CONSTANT_64BIT_FLAG (46)   /*!< \c SEND_VOICE_MEM -- Transmit in SSB message stored in memory */
 #define RIG_FUNC_OVF_STATUS CONSTANT_64BIT_FLAG (47)   /*!< \c OVF_STATUS -- Read overflow status 0=Off, 1=On */
 #define RIG_FUNC_SYNC       CONSTANT_64BIT_FLAG (48)   /*!< \c Synchronize VFOs -- Yaesu FTDX101 SY cmd, IC-7610 "Tracking" */
-#define RIG_FUNC_BIT49      CONSTANT_64BIT_FLAG (49)   /*!< \c available for future RIG_FUNC items */
+#define RIG_FUNC_DUAL_BAND  CONSTANT_64BIT_FLAG (49)   /*!< \c Dual Band or Multi-Band monitoring */
```
dchristle, line 1378: Here `RIG_FUNC_DUAL_BAND` is added, but there's no corresponding entry in `src/misc.c` for `rig_func_str`. So, `rig_parse_func("DUAL_BAND")` returns `NONE` & the named set/get commands will fail before reaching the backend.

But from an API design standpoint, I think it would be better to remove this new flag and instead expose the handlers through the existing `RIG_FUNC_DUAL_WATCH` flag. Despite its name, Hamlib defines that flag as "Dual Watch / Sub Receiver," and other backends like the K3 already use it for independent subreceiver or dual/single reception control.

tests/test_thd75_dual_band.c
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    int local_failed = 0;
+    
+    /* Test disabling Dual Band (DL 0) */
+    retval = rig_set_func(rig, RIG_VFO_CURR, RIG_FUNC_DUAL_BAND, 0);
+    if (retval != RIG_OK && retval != -RIG_ETIMEOUT && retval != -RIG_EIO) {
+        printf("  FAILED: rig_set_func returned %d\n", retval);
+        local_failed = 1;
+    }
```
dchristle, line 163: It looks like here & on 217, timeout and I/O errors are accepted, readback is skipped, and it'll count as a "pass" for the test. The toggle test skips on an initial write failure, too.

We should revise the tests so that they require successful commands and readbacks after opening the radio. The query errors should propagate separately from state and report any failed confirmations accurately.

tests/test_thd75_dual_band.c
```diff
@@ -0,0 +1,363 @@
+/*
```
dchristle, line 1: I recommend adding the DL command and error-path coverage to the existing `testthd75emu.c` rather than introducing a separate automated test executable. By default, this file checks only the constant and advertised capability; all protocol tests require hardware opt-in. Extending the existing emulator would let CI verify DL polarity, malformed replies, acknowledgement mismatches, and timeout handling without hardware.

dchristle (changes requested):

include/hamlib/rig.h
woodard: I kind of disagree with reusing `RIG_FUNC_DUAL_WATCH` for this because if I also want to enable the other dual watch like features like priority scan or call scan which are actually much more like dual watch, then you have already used that functionality for something else. That is why I put the big discussion about the differences between dual watch and dual band radios in the commit message. I think a better approach would be to introduce DUAL_BAND and then let various radios continue to use DUAL_WATCH if they like and if they have legacy programs that expect that, but over time move radios which have two independent transceivers move from DUAL_BAND to dual watch.

dchristle: I read the distinction in the description, but my concern is about Hamlib’s API contract: `RIG_FUNC_DUAL_WATCH` explicitly includes subreceiver control, and existing backends already use it for simultaneous dual reception. The hardware distinction for TH-D75 specifically doesn’t establish why this feature needs a different public flag.

So, is the scope of this PR to add TH-D75 dual-receive control using the existing API, or to introduce a new distinction in Hamlib’s API? For the former, this should use `RIG_FUNC_DUAL_WATCH`. For the latter, the proposal needs to define the new semantics, explain why the existing receiver and scan interfaces are insufficient, and address how existing backends and applications fit that distinction.

But right now, as written, the PR adds an overlapping public flag without resolving those questions. Taking a look at existing backends:

- [rig.h](https://github.com/Hamlib/Hamlib/blob/79b5e29d8d903ba6882fb9febdbb3e724c434227/include/hamlib/rig.h#L1366) explicitly says “Dual Watch / Sub Receiver”
- [K3 implementation](https://github.com/Hamlib/Hamlib/blob/79b5e29d8d903ba6882fb9febdbb3e724c434227/rigs/kenwood/k3.c#L2518) maps it to SB
- [Icom implementation](https://github.com/Hamlib/Hamlib/blob/79b5e29d8d903ba6882fb9febdbb3e724c434227/rigs/icom/icom.c#L7265) uses it for single/dual mode on IC-9100, IC-9700, and ID-5100
- [FTX-1 implementation](https://github.com/Hamlib/Hamlib/blob/79b5e29d8d903ba6882fb9febdbb3e724c434227/rigs/yaesu/ftx1/ftx1_func.c#L244) uses it for dual/single reception

So, there's precedent & the TH-D75 features here seem mostly in line. I do think you have a point that the existing API has a wrinkle -- two independently controllable behaviors need distinguishable API operations. If a radio supports both simultaneous reception and periodic priority checking, exposing both through one Boolean would be inadequate. But this PR doesn't establish the TH-D75 **needs** the new flag.

I'd keep this change within the existing API, and if the intent of the PR is to discuss a broader API redesign, that's something separate & independent of TH-D75 dual band monitoring.

PR #2224, opened 2026-09-21
test(rigctld): synchronize IQ TX packet acceptance

dchristle:
The IQ TX test waited for 720 producer frames, then immediately checked that three packets had been counted. Since the feeder updates the packet counter after writing the samples, thread scheduling could expose 720 frames with only two packets counted.

This PR waits for both conditions together before asserting them. The test still verifies that all three packets were accepted and that the expected 720 IQ frames reached the backend, without depending on scheduling order.

Validation:
- `make -C build/test -j4 test_rigctld_commands`
- `git diff --check`


dchristle:
@N0NB This should fix the recent CI failings on various PRs - the test is flaky and the failure is unrelated to the PR diffs.

